PR #46, opened 2017-08-01
Refining new features

PWhiddy:


jbkalmbach:
Just a couple of ideas. 1) We should add a git ignore file to the repo so that you're not adding all these .ipynb checkpoint files. And 2) are you planning on adding more explanation to the main belt example notebook?

PWhiddy:
There is a global git ignore file, I can add the checkpoints to the list.
Regarding the main belt example, I will add explanation once I've finished adding to it

PR #77, opened 2018-05-16
Hits rerun

smotherh:
run_search class has been refactored to allow eventual region_search class to share much of the functionality.

region_search class currently DOES NOT work.

More cleanup of run_search still to come, but at this stage, it properly reproduces prior results.

### DIFF
--- a/analysis/analysis_utils.py
+++ b/analysis/analysis_utils.py
@@ -180,7 +180,6 @@ class analysis_utils(object):
         hdulist = fits.open('%s/%s' % (im_filepath, self.return_filename(use_images[0])))
         w = WCS(hdulist[1].header)
         image_params['ec_angle'] = self.calc_ecliptic_angle(w)
-        image_params['ec_angle'] = np.pi + 1.25
         del(hdulist)
 
         images = [kb.layered_image('%s/%s' % (im_filepath, self.return_filename(f))) for f in np.sort(use_images)]
@@ -210,6 +209,9 @@ class analysis_utils(object):
         return(search,image_params)
 
     def process_results(self,search,image_params,likelihood_level):
+        """
+        Processes results that are output by the gpu search.
+        """
 
         keep = {'stamps': [], 'new_lh': [], 'results': [], 'times': [],
                 'lc': [], 'final_results': []}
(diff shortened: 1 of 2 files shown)


analysis/analysis_utils.py
```diff
@@ -108,14 +111,277 @@ def __init__(self):
...
+            chunk_values = (res_num, len(keep['results']), results[0].lh, results[-1].lh)
+            for header, val, in zip(chunk_headers, chunk_values):
+                if type(val) == np.int:
+                    print('%s = %i' % (header, val))
+                else:
+                    print('%s = %.2f' % (header, val))
+            print('---------------------------------------')
+            psi_curves = []
+            phi_curves = []
+            print(search.get_results(0,10))
```
jbkalmbach: Do we want to keep this or maybe set some verbose setting?

analysis/analysis_utils.py
```diff
@@ -108,14 +111,277 @@ def __init__(self):
...
+
+            if res_num+chunk_size >= 4000000:
+                likelihood_level = 20.
+                with open('%s/overload_error.txt' %
+                          (res_filepath), 'w') as f:
+                    f.write('In %i total results, %i were kept. Likelihood level down to %f.' %
+                            (res_num + chunk_size, len(keep['results']), line.lh))
+
+            res_num += chunk_size
+
```
jbkalmbach: Do you think we should filter here instead of passing the entire result of the while loop back out and then in to the filtering? The only thing I would be worried about then I guess would be the clustering because we would cluster on separate chunks and still get duplicate results in the same area of parameter space. Maybe we could redo this to stamp filter each chunk and then do a clustering after. What do you think?

analysis/run_search.py
```diff
@@ -6,16 +6,76 @@
...
+
+        # Save the results
+        self.save_results(res_filepath, out_suffix, keep)
+
+        end = time.time()
+
+        del(keep)
+        return
 
 class run_search(analysis_utils):
```
jbkalmbach: Should we rename this to grid search to be clearer?

analysis/analysis_utils.py
smotherh: I think we should keep this by default, but we could maybe cut down on the output in filter_results(), or at least put it behind some verbosity setting as you suggest.

analysis/analysis_utils.py
smotherh: I don't think we should do the clustering here, because both grid_search and region_search require processing, but region_search does not always require clustering, so I think it makes the most sense to just have filtering as a separate function.

analysis/run_search.py
smotherh: Yes, I think we definitely should. Only reason I haven't changed it yet is because I didn't want to break existing scripts that call run_search. But I think the class should definitely be grid_search.

PR #78, opened 2018-10-17
Hits rerun

smotherh:
Fixed a bug where PYTHONPATH was exported in the wrong order, causing the shell to be unable to find kbmod.


PWhiddy:
I still seem to be getting the failed tests when building on an aws instance. I'll look further into this

jbkalmbach:
LGTM. But maybe we should wait until Peter finishes looking into the AWS build?

PWhiddy:
I'm not 100% sure yet but this issue reminds me of the same one we had once installing on @smotherh's laptop where it would build just fine but a mismatching python version (I think?) caused all gpu dependent functions to fail. I'll see if I can write a test that will detect this

PR #83, opened 2019-02-28
Hits rerun

smotherh:
Several improvements to the code, with a focus on compatibility to the Lori Allen dataset and Region Search functionality.

### DIFF
--- a/analysis/analysis_utils.py
+++ b/analysis/analysis_utils.py
@@ -129,7 +129,7 @@ class analysis_utils(object):
 
         return patch_visit_ids
 
-    def load_images(self, im_filepath, time_file,psf_val=1.4,mjd_lims=None):
+    def load_images(self, im_filepath, time_file, psf_val=1.4, mjd_lims=None):
         """
         This function loads images and ingests them into a search object
 
@@ -181,7 +181,6 @@ class analysis_utils(object):
         flag_exceptions = [32,39] # unless it has one of these special combinations of flags
         master_flags = int('100111', 2) # mask any pixels which have any of
         # these flags in more than two images
-        
         hdulist = fits.open('%s/%s' % (im_filepath, self.return_filename(use_images[0])))
         w = WCS(hdulist[1].header)
         image_params['ec_angle'] = self.calc_ecliptic_angle(w)
@@ -231,8 +230,7 @@ class analysis_utils(object):
             print('Starting pooling...')
             pool = mp.Pool(processes=16)
             print('Getting results...')
-            if results is None:
-                results = search.get_results(res_num,chunk_size)
+            results = search.get_results(res_num,chunk_size)
             chunk_headers = ("Chunk Start", "Chunk Size", "Chunk Max Likelihood",
                              "Chunk Min. Likelihood")
             chunk_values = (res_num, len(keep['results']), results[0].lh, results[-1].lh)
--- a/analysis/analysis_utils.py
+++ b/analysis/analysis_utils.py
@@ -151,7 +151,6 @@ class analysis_utils(object):
         image_params : dictionary
             Contains image parameters such as ecliptic angle and mean Julian day
         """
-
         # Empty for now. Will contain x_size, y_size, ec_angle, and mjd before being returned.
         image_params = {}
 
@@ -162,7 +161,6 @@ class analysis_utils(object):
             image_time_dict[str(int(visit_num))] = visit_time
 
         chunk_size = 100000
-
         patch_visits = sorted(os.listdir(im_filepath))
         patch_visit_ids = self.get_folder_visits(patch_visits)
         patch_visit_times = np.array([image_time_dict[str(visit_id)] for visit_id in patch_visit_ids])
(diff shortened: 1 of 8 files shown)


analysis/analysis_utils.py
```diff
@@ -227,8 +228,11 @@ def process_results(self,search,image_params,res_filepath,likelihood_level):
         print("Processing Results")
         print('---------------------------------------')
         while likelihood_limit is False:
+            print('Starting pooling...')
             pool = mp.Pool(processes=16)
-            results = search.get_results(res_num,chunk_size)
+            print('Getting results...')
+            if results is None:
+                results = search.get_results(res_num,chunk_size)
```
jbkalmbach: This breaks the loop. You only ever get the results in the first chunk. This needs to be fixed before merging.

analysis/analysis_utils.py
```diff
@@ -227,7 +227,9 @@ def process_results(self,search,image_params,res_filepath,likelihood_level):
         print("Processing Results")
         print('---------------------------------------')
         while likelihood_limit is False:
+            print('Starting pooling...')
             pool = mp.Pool(processes=16)
+            print('Getting results...')
             results = search.get_results(res_num,chunk_size)
```
jbkalmbach: Now if you put something other than `None` in results, it never gets used. Did you want that option to still work?

smotherh: No. I think that was originally an attempt to get a single process_results class for grid search and region search. I eventually gave up and coded a separate process_region_results() function.

search/src/KBMOSearch.cpp
```diff
@@ -876,7 +903,32 @@ std::vector<RawImage> KBMOSearch::phiStamps(trajectory& t, int radius)
...
 	return createStamps(t, radius, imgs);
 }
-
+std::vector<float> KBMOSearch::psiCurves(trajectory& t)
+{
+    /*Generate a psi lightcurve for further analysis
+     *  INPUT-
+     *    trajectory& t - The trajectory along which to find the lightcurve
+     *  OUTPUT-
+     *    std::vector<double> - A vector of the lightcurve values
```
PWhiddy: Change 'double' to 'float'

search/src/KBMOSearch.cpp
```diff
@@ -876,7 +903,32 @@ std::vector<RawImage> KBMOSearch::phiStamps(trajectory& t, int radius)
...
+	for (auto& im : psiImages) imgs.push_back(&im);
+	return createCurves(t, imgs);
+}
+std::vector<float> KBMOSearch::phiCurves(trajectory& t)
+{
+    /*Generate a phi lightcurve for further analysis
+     *  INPUT-
+     *    trajectory& t - The trajectory along which to find the lightcurve
+     *  OUTPUT-
+     *    std::vector<double> - A vector of the lightcurve values
```
PWhiddy: double -> float

search/src/KBMOSearch.cpp
```diff
@@ -798,6 +798,33 @@ std::vector<RawImage> KBMOSearch::createStamps(trajectory t, int radius, std::ve
...
+std::vector<float> KBMOSearch::createCurves(trajectory t, std::vector<RawImage*> imgs)
+{
+    /*Create a lightcurve from an image along a trajectory
+     *
+     *  INPUT-
+     *    trajectory t - The trajectory along which to compute the lightcurve
+     *    std::vector<RawImage*> imgs - The image from which to compute the
+     *      trajectory. Most likely a psiImage or a phiImage.
+     *   Output-
+     *     std::vector<double> lightcurve - The computed trajectory
```
PWhiddy: double -> float

analysis/analysis_utils.py
```diff
@@ -194,14 +193,14 @@ def load_images(self, im_filepath, time_file, psf_val=1.4, mjd_lims=None):
...
+        #stack.apply_master_mask(master_flags, 2)
 
-        stack.grow_mask()
-        stack.grow_mask()
+        #stack.grow_mask()
+        #stack.grow_mask()
         
         # This applies a mask to pixels with more than 120 counts
-        #stack.apply_mask_threshold(120.)
+        stack.apply_mask_threshold(120.)
```
jbkalmbach: Should this be turned back on?

analysis/analysis_utils.py
```diff
@@ -7,6 +7,7 @@
 import csv
 import astropy.coordinates as astroCoords
 import astropy.units as u
+import pdb
```
jbkalmbach: Should we remove pdb imports before moving things to master?

analysis/analysis_utils.py
smotherh: I think so. Without using bitmasks, having a pixel threshold eliminates a lot of the ultra-high likelihood objects (e.g. 100+).

analysis/analysis_utils.py
smotherh: Probably. Done.

PR #98, opened 2019-10-07
Hits rerun

smotherh:


jbkalmbach:
Didn't look too closely but it looks fine to me if you want to merge it.

search/src/kernels.cu
```diff
@@ -184,180 +186,427 @@ extern "C" void devicePool(int sourceWidth, int sourceHeight, float *source,
...
+__global__ void searchFilterImages(int trajectoryCount, int width, int height,
+        int imageCount, int minObservations, float *psiPhiImages,
+        trajectory *trajectories, trajectory *results, float *imgTimes,
+        float sGL0, float sGL1, float sigmaGCoeff, float minLH)
+{
+
+    // Get trajectory origin
+    const unsigned short x = blockIdx.x*THREAD_DIM_X+threadIdx.x;
+    const unsigned short y = blockIdx.y*THREAD_DIM_Y+threadIdx.y;
+    float lcArray[MAX_NUM_IMAGES];
```
PWhids: These arrays consume a lot of thread local memory (looks like 100 * 4 * 4 bytes), which limits the number of threads that can be simultaneously active on each of the GPUs multiprocessors. Having more threads active on each multiprocessor allows more memory bandwidth to be used.
I'm not sure if you've noticed a significant performance impact when using these large local arrays, but it seems to me that you would have to reduce the number of threads per block, and would see quite a bit less performance.
If you've verified the results are correct and the performance is good, then no need to think too much about this! Just something to be aware of as a general GPU performance tip.
At the very least it might be worth it to try setting the MAX_NUM_IMAGES to the minimum you need and see if there is a noticeable performance difference.

jbkalmbach:
Actually looking at this branch again we might want to do a rebase of some sort since a lot of the things we cleaned out of master did not get cleaned out of here.

PR #103, opened 2021-02-20
Issue/102/update py orbfit extensions

jbkalmbach:


smotherh:
Looks good to me. I've updated my orbit fitting notebook to use the new KbmodInfo class.

PR #114, opened 2022-07-05
Added more tests

jeremykubica:
Added tests to more fully cover pooling and pixel interpolation.

Also added more accessor functions from RawImage to the python bindings.


jeremykubica:
I added a few more tests and cleaned up unneeded path variables in the LayeredImage class.

PR #117, opened 2022-07-14
Some quick cleanups for analysis

jeremykubica:
Added some comments to the functions, cleaned up unused imports, and fixed a few warnings (about np.float and np.int being deprecated)

### DIFF
--- a/analysis/run_search.py
+++ b/analysis/run_search.py
@@ -281,7 +281,7 @@ class run_search:
             self.config['mjd_lims'], self.config['visit_in_filename'],
             self.config['file_format'])
 
-        # Apply the mask to the images and set the PDF.
+        # Apply the mask to the images and set the PSF.
         if self.config['do_mask']:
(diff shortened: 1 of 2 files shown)


analysis/run_search.py
```diff
@@ -284,16 +280,19 @@ def run_search(self):
             self.config['im_filepath'], self.config['time_file'],
             self.config['mjd_lims'], self.config['visit_in_filename'],
             self.config['file_format'])
-        # Save values in image_params for later use
+
+        # Apply the mask to the images and set the PDF.
```
jbkalmbach, line 284: typo, should be PSF

jbkalmbach (approved):
Thanks for the updates. Just pointed out a couple typos I found.

analysis/run_search.py
jeremykubica: Fixed.

jeremykubica:
Fixed both typos. Thanks for catching.

PR #127, opened 2022-07-22
Part 2 of KBMOSearch refactoring

jeremykubica:
- Breaks the stamp creation and aggregation functions out into RawImage to reduce duplication and improve test coverage. 
- Removes some duplicate functions. 
- Changes CreateStamp to produce both interpolated and non-interpolated stamps.
- Rename some of the functions for consistency (which leads to a name change only in analysis_utils.py)
- 
Note: This change has passed the updated diff test (matching results and ps_ files.)

### DIFF
--- a/search/src/kernels.cu
+++ b/search/src/kernels.cu
@@ -44,7 +44,7 @@ __global__ void convolvePSF(int width, int height,
                     (y + j >= 0) && (y + j < height))
                 {
                     float currentPixel = sourceImage[(y+j)*width+(x+i)];
-                    if (currentPixel != NO_DATA) 
+                    if (currentPixel != NO_DATA)
                     {
                         float currentPSF = psf[(j+psfRad)*psfDim+(i+psfRad)];
                         psfPortion += currentPSF;
--- a/search/src/kernels.cu
+++ b/search/src/kernels.cu
@@ -11,7 +11,6 @@
 #define MAX_NUM_IMAGES 140
 
 #include "common.h"
-//#include "PointSpreadFunc.h"
 #include <helper_cuda.h>
 #include <stdio.h>
 #include <float.h>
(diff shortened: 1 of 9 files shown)


search/src/kernels.cu
```diff
@@ -30,26 +29,27 @@ __global__ void convolvePSF(int width, int height,
...
             // #pragma unroll
-            for (int i=minX; i<=maxX; i++)
+            for (int i = -psfRad; i <= psfRad; i++)
             {
-                float currentPixel = sourceImage[j*width+i];
-                if (currentPixel != NO_DATA) {
-                    float currentPSF = psf[(j-minY)*psfDim+i-minX];
-                    psfPortion += currentPSF;
-                    sum += currentPixel * currentPSF;
+                if ((x + i >= 0) && (x + i < width) &&
```
PWhiddy, line 43: In general good to avoid branching in GPU kernels when possible but in the common case all threads will be taking the same path anyway so it shouldn't really matter. And convolution won't be a performance bottleneck for the whole pipeline anyway as long as it's reasonably performant.

PWhiddy:
Good catch on the PSF index offsets!

tests/test_raw_image.py
```diff
@@ -28,12 +29,86 @@ def test_set_all(self):
...
+        p = psf(np.array(psf_data))
+        self.assertAlmostEqual(p.get_sum(), 1.0, delta=0.00001)
+
+        # Make a clean version of the image for the average function.
+        img2 = raw_image(self.width, self.height)
+        for x in range(self.width):
+            for y in range(self.height):
+                img2.set_pixel(x, y, self.img.get_pixel(x, y))
+
+        # Convolce the psf with the copy of the image.
```
PWhiddy, line 82: minor typo - "Convolce" -> "convolve"

search/src/kernels.cu
jeremykubica: Thanks for the heads up about branching in GPUs. I will try to think of alternate ways to do this type of thing in the future.

tests/test_raw_image.py
jeremykubica: Fixed.

PWhiddy:
@smotherh Are you still using a gaussian PSF or one derived from the images? If an asymmetric PSF is being used or is planned to be used in the future I think it'd be a good idea to add a test that verifies the PSF is being applied in an orientation that matches expectation. 
Maybe there's even existing software in the ecosystem which could handle PSF calculations end to end? Just a thought.

PWhiddy:
One way to test kernel orientation could be to compare results to scipy.signal.convolve2d, though I don't think any of its boundary modes exactly correspond to this implementations kernel normalization.

smotherh:
Yes, we are still using a gaussian psf, but we have plans to update this in the near future, so a test like what you propose would be a very nice thing to have.

jeremykubica:
Good call about the orientation dependent test. I added a simple manual case.

PR #129, opened 2022-07-29
Speed up the growMask function

jeremykubica:
Speed up the growMask function and a few small other cleanups.

Saves ~10 seconds on initializing the diff test.

### DIFF
--- a/search/src/RawImage.cpp
+++ b/search/src/RawImage.cpp
@@ -194,31 +194,55 @@ void RawImage::applyMask(int flags, const std::vector<int>& exceptions,
 	}
 }
 
-void RawImage::growMask()
+/* This implementation of growMask is optimized for steps > 1
+   (which is how the code is generally used. If you are only
+   growing the mask by 1, the extra copy will be a little slower.
+*/
+void RawImage::growMask(int steps)
 {
-	// Parallel?
-  // This function requires a temporary mask to store the FLAGGED values
-  // Without it, the function will fail to flag all pixels around the masked
-  // regions. Notably, it fails to flag pixels "down" and "right" of the
-  // masked regions.
-  std::vector<float> tmpMask;
-  tmpMask = pixels;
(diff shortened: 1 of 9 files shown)


search/src/RawImage.cpp
```diff
@@ -194,31 +194,55 @@ void RawImage::applyMask(int flags, const std::vector<int>& exceptions,
...
-  for(std::size_t i=0; i < pixels.size(); ++i)
-  {
-        if (tmpMask[i]==FLAGGED) { pixels[i] = NO_DATA; }
-  }
+    const int num_pixels = width * height;
+    std::vector<int> masked(num_pixels, -1);
+
+    // Set up the initial masked vector that stores the number of steps
+    // each pixel is from a masked pixel in the original image.
+    for (int x = 0; x < width; ++x)
```
PWhiddy, line 208: minor: this could be simplified into a single loop like the bottom one

tests/test_layered_image.py
```diff
@@ -169,6 +169,24 @@ def test_grow_mask(self):
...
+      self.image.grow_mask(3)
+
+      # Check that the mask has grown to all applicable pixels.
+      science = self.image.get_science()
+      for y in range(self.image.get_height()):
+         for x in range(self.image.get_width()):
+            # Check whether the point is a manhattan distance of <= 3 from
+            # one of the original masked pixels.
+            dx = abs(x - 10)
+            dy = min(abs(y - 11), abs(y - 12))
```
PWhiddy, line 187: minor/optional - this is a clever way to test the masking, but if the values 3, [10, 11], [10, 12] were variables it might be a little easier to read.

PWhiddy (approved):
Looks good, thanks!

search/src/RawImage.cpp
jeremykubica: Good call. I made that change.

PR #130, opened 2022-08-01
Combined the two GPU search functions

jeremykubica:
Merged the original GPU search with the GPU search + filter to reduce code duplication and improve extensibility.

As part of this, also updated the API for these functions. Instead of calling separate functions, helper functions enableGPUFilter and enableCorr are used to turn on gpu filtering and barycentric corrections respectively.


smotherh:
This is great. In particular, I'm making a note here that this change also allows the user inputs for sigmaG lims to be used in the GPU filter. This is the preferred functionality (and is really great to see added), but I think it might cause different results compared to earlier versions of the code in the case where sigamG lims are not [25,75] (i.e. the 25th and 75th percentiles).

PR #131, opened 2022-08-04
Clean up create_stamps.py

jeremykubica:
Fix behavior when load_filt_tools is set to False and add a bunch of comments.

### DIFF
--- a/analysis/create_stamps.py
+++ b/analysis/create_stamps.py
@@ -205,8 +205,6 @@ class CreateStamps(object):
                     axi += 1
         ax[0,0].set_title(
             'Total SNR={:.2f}'.format(coadd_SNR))
-        #ax[0,0].set_title(
-        #    'Total SNR={:.2f}\nSummed SNR={:.2f}'.format(coadd_SNR,summed_SNR))
         for axis in ['top','bottom','left','right']:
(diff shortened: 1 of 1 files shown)


analysis/create_stamps.py
```diff
@@ -51,34 +78,69 @@ def load_psi_phi(self, psi_filename, phi_filename, lc_index_filename):
...
         stamps = np.genfromtxt(stamp_filename)
         if len(np.shape(stamps)) < 2:
             stamps = np.array([stamps])
-        #stamp_normalized = stamps/np.sum(stamps, axis=1).reshape(len(stamps), 1)
 
         return stamps
 
     def stamp_filter(self, stamps, center_thresh, verbose=True):
-
+        """
```
jbkalmbach, line 115: Maybe we want to rename this function to something more specific since there are different stages of "stamp filtering"?

jbkalmbach (approved):
Looks good, just one comment about changing a function name to something more descriptive while we're at it.

analysis/create_stamps.py
jeremykubica: Good call. I updated this to max_value_stamp_filter

PR #135, opened 2022-08-22
Test whether we can compress the input images to uint8 and still be able to coadd low SNR images

connolly:
This notebook tests the ability to truncate images used in KBMOD to uint8 and still perform image coaddition with low SNR sources. Compression of between 4x and 8x seem possible for the size of the input images for the GPU


jeremykubica:
Closing this pull request because we implemented the compression in the main code.

PR #137, opened 2022-08-26
Correct combining of masked images

jeremykubica:
The median function would set masked pixels to zero, which can skew the distribution. The new code just computes the median over the non-masked pixels.

Also adds a computeMeanImage function.

Adds corresponding tests.

### DIFF
--- a/search/src/RawImage.cpp
+++ b/search/src/RawImage.cpp
@@ -442,12 +442,17 @@ RawImage createMedianImage(const std::vector<RawImage>& images)
                     num_unmasked += 1;
                 }
             }
-            
-            int median_ind = num_unmasked / 2;
-            std::nth_element(pixArray.begin(),
-                             pixArray.begin() + median_ind,
-                             pixArray.begin() + num_unmasked);
-            result.setPixel(x, y, pixArray[median_ind]);
+
+            if (num_unmasked > 0)
+            {
+                int median_ind = num_unmasked / 2;
+                std::nth_element(pixArray.begin(),
+                                 pixArray.begin() + median_ind,
+                                 pixArray.begin() + num_unmasked);
+                result.setPixel(x, y, pixArray[median_ind]);
+            } else {
+                result.setPixel(x, y, NO_DATA);
+            }
         }
     }
 
@@ -482,7 +487,7 @@ RawImage createSummedImage(const std::vector<RawImage>& images)
 
     return result;
 }
-    
+
 RawImage createMeanImage(const std::vector<RawImage>& images)
 {
     int num_images = images.size();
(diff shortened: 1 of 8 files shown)


search/src/RawImage.cpp
```diff
@@ -432,16 +431,28 @@ RawImage createMedianImage(const std::vector<RawImage>& images)
...
+
+            if (num_unmasked > 0)
+            {
+                int median_ind = num_unmasked / 2;
+                std::nth_element(pixArray.begin(),
+                                 pixArray.begin() + median_ind,
+                                 pixArray.begin() + num_unmasked);
+                result.setPixel(x, y, pixArray[median_ind]);
+            } else {
+                result.setPixel(x, y, NO_DATA);
```
jbkalmbach, line 454: How does leaving pixels with `NO_DATA` in the stamps affect filtering? That's the only change here I see that affects things further down the pipeline.

tests/test_raw_image.py
```diff
@@ -203,6 +246,50 @@ def test_create_summed_image(self):
...
+        self.assertEqual(mean_image.get_width(), 2)
+        self.assertEqual(mean_image.get_height(), 3)
+        self.assertAlmostEqual(mean_image.get_pixel(0, 0), 0.0, delta=1e-6)
+        self.assertAlmostEqual(mean_image.get_pixel(1, 0), -1.0, delta=1e-6)
+        self.assertAlmostEqual(mean_image.get_pixel(0, 1), 2.0, delta=1e-6)
+        self.assertAlmostEqual(mean_image.get_pixel(1, 1), 9.5/3.0, delta=1e-6)
+        self.assertAlmostEqual(mean_image.get_pixel(0, 2), 8.8/3.0, delta=1e-6)
+        self.assertAlmostEqual(mean_image.get_pixel(1, 2), 9.4/3.0, delta=1e-6)
+
+        # Apply masks to images 1 and 3.
```
jbkalmbach, line 278: Comment doesn't match code below.

jbkalmbach (approved):
Looks good. Thanks for the update, this change makes a lot of sense to add.

search/src/RawImage.cpp
jeremykubica: That's a good point. The maximum value based filtering will work fine, but the CNN filtering might not work correctly anymore. I changed this to return a result of 0.0 in the case of all the pixels being no data to stay consistent with the old code.

tests/test_raw_image.py
jeremykubica: Fixed.

jeremykubica:
I fixed your comments and, in the process, realized that the changes needed to be propagated a bit further into the python code. I also added a bunch of tests and realized that the way we have been computing median is not consistent between the python and c++ code (if there are an even number of elements np.median takes the average of the values), so I corrected this as well.

Can you take another look when you get a chance?

jbkalmbach (approved):
New changes look good. In the future it would be good to think about the best way to handle any masked pixels downstream in our filtering pipeline and then revisit what we do here. Should they be 0? Or should we come up with an alternate value to tell whatever we are using for filtering that there is no data? Because a real value around 0 is actual information compared to a pixel with no data.

Issue #48, opened 2017-08-02
Extend the PSF model to allow for a variable PSF (spatial and temporal)


connolly:
PSF's can vary spatially and temporally. The PSF class should be able to account for this and the stack.convolve() should be able to take in a set of PSFs that are appropriate for each image in the stack

jbkalmbach:
Based upon varying seeing in Lori Allen data we should have some way to at least change the Gaussian size over a series of images.

jeremykubica:
If we can realistically expect 1 psf per image, then it would be easy to save the psf within the LayeredImage class. If a PSF has been assigned, the class would use that. Otherwise it can use a provided default. Would this address the request in this bug?

jbkalmbach:
Yeah, I think that would work. We should have a PSF per image from the processed images.

jbkalmbach:
And if nothing else we could create a Gaussian for each image based upon the seeing in each image.

jeremykubica:
This was completed in an earlier PR

PR #139, opened 2022-08-30
Allow KBMOD to use a different PSF per image.

jeremykubica:
Creates an image specific PSF within layered images and uses those for convolution, adding objects, etc.

NOTE: This is a breaking change in that it changes the API for ImageStack and LayeredImage.


jeremykubica:
Closes #48

tests/test_image_stack.py
```diff
@@ -16,7 +16,8 @@ def setUp(self):
                                            60,    # dim_y = 60 pixels,
                                            2.0,   # noise_level
                                            4.0,   # variance
-                                           2.0 * i)  # time
+                                           2.0 * i,  # time
```
maxwest-uw, line 19: should we have a test added for the image_stack class to check that it can properly access individual image psfs?

maxwest-uw (commented):
altogether these are great changes, I had one suggestion for adding a small test to ensure that image_stack can access individual image psfs.
one thing to consider, is there anything we need to do/anybody we need to warn that this is 'breaking' change since it changes the image api? I know that kbmod isn't part of any pipelines yet but I want to make sure we don't break someones ongoing work!

tests/test_image_stack.py
jeremykubica: Good call. I added a new test (test_different_psfs) to check that it is handling the per-image PSFs correctly.

maxwest-uw (approved):
everything looks good to me! thanks for adding that test!

maxwest-uw (approved):
👍

PR #140, opened 2022-09-09
Structural cleanup of KBMOSearch

jeremykubica:
Removed some "global" data from the KBMOSearch class in order to simplify state tracking (more information is passed between functions where needed).


maxwest-uw (approved):
looks good to me!

PR #142, opened 2022-09-16
Improve on GPU Filtering (Part 1)

jeremykubica:
Improves the on GPU filtering to drop masked values (instead of filling with zero). Note that this impact the filtering behavior and increases the number of results returned. Any locally saved diff test goldens will need to be regenerated.

This change also refactors the filtering code into a separate function and provides C++ and python wrapped versions that can possibly be used by analysis_utils.py or other parts of the code in the future.

Adds tests for the filtering code to ensure the correct behavior.


jeremykubica:
I broke this pull request into two parts to unblock other filtering work. This is the first part that puts in the background functions but does **not** change functionality. The second PR will update the on GPU filtering.

jeremykubica:
I also added a tests with other coefficients at Hayden's suggestion.

search/src/filtering_kernels.cu
```diff
@@ -0,0 +1,75 @@
...
+    // Compute the index of each of the percent values in values
+    // from the given bounds sGL0, 0.5 (median), and sGL1.
+    const int pct_L = int((num_values + 1) * sGL0 + 0.5) - 1;
+    const int pct_H = int((num_values + 1) * sGL1 + 0.5) - 1;
+    const int median_ind = int((num_values + 1) * 0.5 + 0.5) - 1;
+
+    // Compute the values that are +/- 2*sigmaG from the median.
+    float sigmaG = sigmaGCoeff * (values[idxArray[pct_H]]
+                                  - values[idxArray[pct_L]]);
+    float minValue = values[idxArray[median_ind]] - 2 * sigmaG;
```
jbkalmbach, line 53: Do we want the sigmaG multiplier (in this case 2) to be something the user can set?

tests/test_filtering.py
```diff
@@ -0,0 +1,87 @@
...
+import tempfile
+import unittest
+
+class test_kernels_wrappers(unittest.TestCase):        
+    def test_sigmag_filtered_indices_same(self):
+        # With everything the same, nothing should be filtered.
+        values = [1.0 for _ in range(20)]
+        inds = sigmag_filtered_indices(values, 0.25, 0.75, 0.7413);
+        self.assertEqual(len(inds), 20)
+        for idx in inds:
```
jbkalmbach, line 12: Is this test supposed to be here? I thought everything in the loop should just be equal to 1.0?

tests/test_filtering.py
```diff
@@ -0,0 +1,87 @@
...
+        self.assertEqual(len(inds), len(values) - 1)
+
+        # The only missing value should be index=8.
+        for i in range(8):
+            self.assertTrue(i in inds)
+        self.assertFalse(8 in inds)
+
+    def test_sigmag_filtered_indices_other_bounds(self):
+        # Do the filtering of test_sigmag_filtered_indices_one_outlier
+        # with wider bounds [-5.232608, 7.232608].
```
jbkalmbach, line 37: This seems a little off to me. I would think the 85th percentile value should be 2 not 5.46 so that the bounds should be [-1.89, 3.89]. Is this a bug in the percentile value selection? Or is this a result of the percentile calculation working the way it should but in a different way than I expect?

search/src/filtering_kernels.cu
jeremykubica: That's a good question. It would be relatively easy to include it, but I am not sure how often the user will want to use a different value.

tests/test_filtering.py
jeremykubica: This is a really good catch. I've been using the percentile computation from the original code:
`int((num_values + 1) * sGL1 + 0.5) - 1`
but I think a better calculation would be to use:
`int(num_values * sGL1)`
Does that sound right to you?

tests/test_filtering.py
jeremykubica: I'm not sure why that test is there anymore. I will remove it.

search/src/filtering_kernels.cu
jbkalmbach: I think it will be a useful thing to set especially since it's something you might want to adjust when applying on different data sets.

tests/test_filtering.py
jeremykubica: Actually, after looking it up. It sounds like a common approximation for the index is:
`ceil(num_values * sGL1)`
To avoid problems with floating point precision and account for being zero indexed, we can use:
`int(ceil(num_values * 0.5) + 0.001) - 1;`

search/src/filtering_kernels.cu
jeremykubica: Sounds good. I will add it.

tests/test_filtering.py
jeremykubica: Good call. I added checks to the C++ function and then additional clipping in the on GPU function.

jeremykubica:
Thanks Bryce. I think I have addressed all your comments.

search/src/filtering_kernels.cu
jbkalmbach: Looks like width needs to replace the 2 in these functions still.

jbkalmbach (approved):
Just think width needs to get propagated into the code. But the percentile changes are great. Thanks!

search/src/filtering_kernels.cu
jeremykubica: Good catch again. I'm not sure how I missed that. I added it and also added to the tests to confirm that it is working.

maxwest-uw (approved):
👍

PR #145, opened 2022-09-22
Improve on GPU Filtering (Part 2)

jeremykubica:
Replace the GPU sigmaG filtering with the new version


search/src/kernels.cu
```diff
@@ -358,91 +362,58 @@ __global__ void searchFilterImages(int trajectoryCount, int width, int height,
...
+            phiArray[num_seen] = cPsiPhi.y;
             if (cPsiPhi.y == 0.0)
             {
-                lcArray[i] = 0;
+                lcArray[num_seen] = 0.0;
             } else {
-                lcArray[i] = cPsiPhi.x/cPsiPhi.y;
+                lcArray[num_seen] = cPsiPhi.x/cPsiPhi.y;
             }
+            num_seen += 1;
```
delucchi-cmu, line 381: Looks like `num_seen` is always the same as `i`, and will have a final value equal to `image_count`. Is there something I'm missing that makes this variable necessary?

search/src/kernels.cu
```diff
@@ -358,91 +362,58 @@ __global__ void searchFilterImages(int trajectoryCount, int width, int height,
...
             if (cPsiPhi.x == NO_DATA) continue;
 
             currentT.obsCount++;
             psiSum += cPsiPhi.x;
             phiSum += cPsiPhi.y;
-            psiArray[i] = cPsiPhi.x;
-            phiArray[i] = cPsiPhi.y;
+            psiArray[num_seen] = cPsiPhi.x;
+            phiArray[num_seen] = cPsiPhi.y;
             if (cPsiPhi.y == 0.0)
```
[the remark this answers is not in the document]
jeremykubica: Fixed.

search/src/kernels.cu
jeremykubica: Unfortunately `num_seen` isn't always the same as `i` because there is a conditional with a continue statement a few lines above. I will change this in the code to not use a continue and (hopefully) be clearer.

PR #146, opened 2022-09-22
Add some region search helper functions

jeremykubica:
- Adds a few helper functions to PooledImage that have been useful in debugging and understanding region search.
- Changes the region search LH evaluation to match both the linear search and other filters
- Cleans up an unused function and an unimplemented function definition.
- Adds tests to cover the new functions


maxwest-uw (commented):
lgtm

maxwest-uw (approved):

PR #148, opened 2022-09-28
Kalman filter port

maxwest-uw:
- port the kalman filtering code form analysis_utils.py into c++
- adds test for kalman filtering and related functions

### DIFF
--- a/analysis/analysis_utils.py
+++ b/analysis/analysis_utils.py
@@ -967,7 +967,7 @@ class PostProcess(SharedTools):
         results = old_results['results']
         keep = self.gen_results_dict()
         
-        keep_idx_results = kb.kalman_filtered_indices(psi_curves.tolist(), phi_curves.tolist(), len(psi_curves))
+        keep_idx_results = kb.kalman_filtered_indices(psi_curves.tolist(), phi_curves.tolist())
 
         end_time = time.time()
         time_elapsed = end_time-start_time
--- a/analysis/analysis_utils.py
+++ b/analysis/analysis_utils.py
@@ -963,20 +963,26 @@ class PostProcess(SharedTools):
         # Make copies of the values in 'old_results' and create a new dict
         psi_curves = np.copy(old_results['psi_curves'])
         phi_curves = np.copy(old_results['phi_curves'])
-        masked_phi_curves = np.copy(phi_curves)
-        masked_phi_curves[masked_phi_curves==0] = 1e9
+#         masked_phi_curves = np.copy(phi_curves)
+#         masked_phi_curves[masked_phi_curves==0] = 1e9
         results = old_results['results']
         keep = self.gen_results_dict()
-
-        print('Starting pooling...')
-        pool = mp.Pool(processes=self.num_cores)
-        zipped_curves = zip(
-            psi_curves, phi_curves, [j for j in range(len(psi_curves))])
-        keep_idx_results = pool.starmap_async(
-            self._return_indices, zipped_curves)
-        pool.close()
-        pool.join()
-        keep_idx_results = keep_idx_results.get()
+        
+#         print(len(psi_curves))
+        
+        keep_idx_results = kb.kalman_filtered_indices(psi_curves.tolist(), phi_curves.tolist(), len(psi_curves))
+        
+#         print('Starting pooling...')
+#         pool = mp.Pool(processes=self.num_cores)
+#         zipped_curves = zip(
+#             psi_curves, phi_curves, [j for j in range(len(psi_curves))])
+#         keep_idx_results = pool.starmap_async(
+#             self._return_indices, zipped_curves)
+#         pool.close()
+#         pool.join()
+#         keep_idx_results = keep_idx_results.get()
+#         print(keep_idx_results)
+        print(keep_idx_results[0:100])
         print('---------------------------------------')
         return(keep_idx_results)
 
--- a/analysis/analysis_utils.py
+++ b/analysis/analysis_utils.py
@@ -850,7 +850,7 @@ class PostProcess(SharedTools):
             new_lh = 0
             good_index=[-1]
         else:
-            new_lh = self._compute_lh(
+            new_lh = kb.calculate_likelihood_psi_phi(
                 psi_curve[good_index], phi_curve[good_index])
         return(index,good_index,new_lh)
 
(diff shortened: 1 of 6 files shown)


search/src/Filtering.cpp
```diff
@@ -46,5 +47,207 @@ std::vector<int> sigmaGFilteredIndices(const std::vector<float>& values,
...
+std::tuple<std::vector<double>, std::vector<double>> calculateKalmanFlux(std::vector<double> fluxValues, 
+                                                                         std::vector<double> vars,
+                                                                         std::vector<int> fluxIdx, bool reverse)
+{
+    int fluxSize = fluxValues.size();
+    double kalmanFluxes[fluxSize];
+    
+    std::vector<double> xhat(fluxSize, 0.0);
+    std::vector<double> p(fluxSize, 0.0);
+    double xhatMinus[fluxSize];
```
drewoldag, line 85: I'm behind on C++, but why are some variables defined as `std::vector<double>` vs. `double x[size]`?

search/src/Filtering.cpp
```diff
@@ -46,5 +47,207 @@ std::vector<int> sigmaGFilteredIndices(const std::vector<float>& values,
...
+        for(int i = 1; i < fluxIdx.size(); i++) {
+            int idx = fluxIdx[i];
+            int idxMinus = fluxIdx[i - 1];
+            
+            xhatMinus[idx] = xhat[idxMinus];
+            pMinus[idx] = p[idxMinus] + q;
+
+            k[idx] = pMinus[idx] / (pMinus[idx] + vars[idx]);
+            xhat[idx] = xhatMinus[idx] + k[idx]*(fluxValues[idx]-xhatMinus[idx]);
+            p[idx] = (1.0-k[idx])*pMinus[idx];
```
drewoldag, line 109: Are lines 107-109 the same as 121-123? And if so, would it be beneficial to turn them into a separate function?

search/src/Filtering.cpp
```diff
@@ -46,5 +47,207 @@ std::vector<int> sigmaGFilteredIndices(const std::vector<float>& values,
...
+    // Mask phi values, calculate the flux,
+    // the inverse phi values, and get a list
+    // of positive flux values.
+    for(int i = 0; i < numValues; i++) {
+        double masked_phi = phiCurve[i];
+        if(masked_phi == 0.0 ) {
+            masked_phi = 1e9;
+        }
+        if(masked_phi < -999.0) {
+            masked_phi = 9999999.0;
```
drewoldag, line 157: Are the values of `masked_phi` on line 154 and 157 magic numbers, or are they just suppose to be very large?

drewoldag:
Looks cool to me, I left a couple of questions, primarily for my own interest :)

jeremykubica (changes requested):

tests/test_filtering.py
```diff
@@ -87,6 +87,97 @@ def test_sigmag_filtered_indices_three_outliers(self):
...
+        phi_values.append([-999.1 for _ in range(20)])
+        
+        inds = kalman_filtered_indices(psi_values, phi_values)
+        
+        self.assertEqual(len(inds), 20)
+        for i in inds[:-1]:
+            self.assertTrue(i[1] == [-1])
+            self.assertEqual(i[2], 0.0)
+            
+    def test_kalman_filtered_indices_bright_first_obs(self):
```
jeremykubica, line 148: Can you also add a test where the kalman filter rejects one of the observations in the middle? Maybe something like:
```
psi_values = [[10., 10., 0.001, 10., 10.]]
phi_values = [[1., 1., 1., 1., 1]]
```

search/src/Filtering.cpp
maxwest-uw: in c++ traditional arrays are a fixed size while std::vectors are dynamically allocated!

search/src/Filtering.cpp
maxwest-uw: in the original python code, the fluxes are passed through the kalman filter once and then all the flux and variance values are reversed and passed through again. In the c implementation I decided against doing that to save on time/memory complexity. I originally tried doing this by writing the loop once and having the 'direction of calculation' handled by an iteration variable but it ended up being really hard to read, so for code clarity's sake I just split it into two similar loops, forwards and backwards. definitely open to changing up the structure if y'all have a better idea since I agree that it's kind of ugly 😭

search/src/Filtering.cpp
maxwest-uw: I think that they might just be large numbers? I kept the values the same as the ones in the python implementation but it might not actually make a difference 🤷‍♀️

analysis/analysis_utils.py
```diff
@@ -1084,112 +1079,6 @@ def apply_clustering(self, keep, image_params):
...
-            new_lh = self._compute_lh(new_psi,new_phi)
-            return (val_on, flux_idx[keep_idx], new_lh)
-        else:
-            reverse_idx = len(flux_idx)-1 - keep_idx_back
-            new_psi = psi_values[flux_idx[reverse_idx]]
-            new_phi = phi_values[flux_idx[reverse_idx]]
-            new_lh = self._compute_lh(new_psi,new_phi)
-            return (val_on, flux_idx[reverse_idx], new_lh)
-
     def _compute_lh(self, psi_values, phi_values):
```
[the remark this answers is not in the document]
maxwest-uw: yeah good idea! I just did this when I was testing to make sure it wouldn't break anything so I'll make that change

search/src/Filtering.cpp
```diff
@@ -46,5 +47,207 @@ std::vector<int> sigmaGFilteredIndices(const std::vector<float>& values,
...
+    }
+    
+    return psiSum / sqrt(phiSum);
+}
+    
+/* Given a set of fluxValues,
+   inverse Phi values, and a
+   vector of positive flux indicies,
+   apply the kalman filters to 
+   said positive values. */
```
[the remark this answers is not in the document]
maxwest-uw: yes, that's correct, I'll expand the comment

search/src/Filtering.cpp
```diff
@@ -46,5 +47,207 @@ std::vector<int> sigmaGFilteredIndices(const std::vector<float>& values,
...
+        p[fluxIdx[0]] = vars[fluxIdx[0]];
+        
+        for(int i = 1; i < fluxIdx.size(); i++) {
+            int idx = fluxIdx[i];
+            int idxMinus = fluxIdx[i - 1];
+            
+            xhatMinus[idx] = xhat[idxMinus];
+            pMinus[idx] = p[idxMinus] + q;
+
+            k[idx] = pMinus[idx] / (pMinus[idx] + vars[idx]);
```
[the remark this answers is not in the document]
maxwest-uw: I'll try this out

search/src/Filtering.cpp
```diff
@@ -46,5 +47,207 @@ std::vector<int> sigmaGFilteredIndices(const std::vector<float>& values,
...
+    
+    // Mask phi values, calculate the flux,
+    // the inverse phi values, and get a list
+    // of positive flux values.
+    for(int i = 0; i < numValues; i++) {
+        double masked_phi = phiCurve[i];
+        if(masked_phi == 0.0 ) {
+            masked_phi = 1e9;
+        }
+        if(masked_phi < -999.0) {
```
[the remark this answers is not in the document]
maxwest-uw: yeah I was unsure about what it did in the original code as well, but I left it in just in case it affected something 🤔

search/src/Filtering.cpp
cchris28: here it would be more about memory allocation I think. What do you gain by using a c style array?

maxwest-uw: (resolving since I just made these into single double variables anyway lol)

jeremykubica (approved):

PR #150, opened 2022-10-03
Small optimizations

jeremykubica:
Make several small fixes to the GPU code that speed it up and prevent passing back invalid values.

### DIFF
--- a/search/src/kernels.cu
+++ b/search/src/kernels.cu
@@ -289,7 +289,7 @@ __global__ void searchFilterImages(int trajectoryCount, int width, int height,
     // We also set (x, y) because they are used in the later python
     // functions.
     trajectory best[RESULTS_PER_PIXEL];
-    for (int r=0; r < RESULTS_PER_PIXEL; ++r)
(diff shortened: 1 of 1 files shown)


search/src/kernels.cu
```diff
@@ -445,7 +454,7 @@ __global__ void searchFilterImages(int trajectoryCount, int width, int height,
         for (int r = 0; r < RESULTS_PER_PIXEL; ++r)
         {
             if (currentT.lh > best[r].lh &&
-                currentT.obsCount >= minObservations)
+                currentT.lh >= minLH)
```
drewoldag: Maybe my logic is flawed, but is it possible for `currentT.lh < minLH` given the second argument in the conditional on line 377?

maxwest-uw (approved):
everything here looks good to me! one thing is that since we're making changes to the number of results that we're returning during the search step, it might be smart to let users know to expect these changes. (I know when I was testing kalman filtering changes and I pulled in some changes that caused the total number of search results to drop a bit due to filtering I thought I broke something 😭 ). idk what the best way to communicate with end users is tho other than the slack channel, just something to think about!

PR #151, opened 2022-10-03
A bunch of small cleanups

jeremykubica:
- Remove the base class for images since we do not use any of the functions of inheritance.
- Remove an unneeded main file (since tests are run through python now).
- Remove an unused constant.
- Change the image classes to store only one representation of their dimensions.


search/src/LayeredImage.cpp
```diff
@@ -70,6 +70,7 @@ void LayeredImage::readHeader(const std::string& filePath)
 		fits_report_error(stderr, status);
 
 	// Read image Dimensions
+	long dimensions[2];
 	if (fits_read_keys_lng(fptr, "NAXIS", 1, 2, dimensions, &fileNotFound, &status))
 		fits_report_error(stderr, status);
```
delucchi-cmu, line 76: I can't explain why, but it weirds me out that just below, we pull out the `width` and `height` from the `dimensions` array, then keep using the `dimensions` array. There's something in all of this that still smells bad and I can't put a finger on it.

And the `fileNotFound` thing is never used.

jeremykubica: That's fair. I updated the line below to use `width` and `height`.

As for the `fileNotFound`... I'm not sure how this is meant to be used. Errors already get picked up and reported via `fits_report_error`.

maxwest-uw (approved):

PR #153, opened 2022-10-10
Refactor image code and offload mask growing to GPU

jeremykubica:
This change refactors the structure of the image code to:
- Move the image related kernels, such as convolution and pooling, into their own file (image_kernels.cu)
- Offload the `grow_mask` function to the GPU and reduce the diff test load time by ~25 seconds
- Combine two nearly identical "save" functions in `RawImage`.
- Move psi and phi generation from `KBMOSearch` to `LayeredImage` (and add tests).

In addition the code adds some bounds checking to the results loading in analysis_utils and combines filtering_kernels.cu into kernels.cu.


search/src/KBMOSearch.cpp
```diff
@@ -142,59 +142,26 @@ std::vector<trajRegion> KBMOSearch::regionSearch(float xVel, float yVel, float r
...
-}
-
 void KBMOSearch::savePsiPhi(const std::string& path) {
     preparePsiPhi();
     saveImages(path);
 }
 
 void KBMOSearch::preparePsiPhi() {
     if (!psiPhiGenerated) {
-        startTimer("Preparing psi and phi images");
```
delucchi-cmu, line 158: Is the timing information no longer helpful for this method? Are you all done speeding it up? =]

search/src/RawImage.cpp
```diff
@@ -169,39 +143,43 @@ void RawImage::applyMask(int flags, const std::vector<int>& exceptions, const Ra
...
-    // Set up the initial masked vector that stores the number of steps
-    // each pixel is from a masked pixel in the original image.
-    std::vector<int> masked(num_pixels, -1);
-    for (int i = 0; i < num_pixels; ++i) {
-        if (pixels[i] == NO_DATA) masked[i] = 0;
-    }
+void RawImage::growMask(int steps, bool on_gpu) {
+    if (on_gpu) {
+        deviceGrowMask(width, height, pixels.data(), pixels.data(), steps);
+    } else {
```
delucchi-cmu, line 149: To avoid adding another nesting layer, just break out early for gpu processing. i.e.

```
if(on_gpu){
  deviceGrowMask();
  return;
}
//keep everything else the same.
```

search/src/image_kernels.cu
```diff
@@ -0,0 +1,319 @@
+/*
```
delucchi-cmu, line 1: A lot of the content of this file is from splitting the kernels.cu file. Can this be branched off that file instead, to maintain that semantic link between them?

search/src/ImageStack.cpp
```diff
@@ -88,13 +88,7 @@ void ImageStack::convolvePSF() {
...
 }
 
-void ImageStack::saveMasterMask(const std::string& path) {
-    // std::cout << masterMask.getWidth() << "\n";
-    // std::cout << masterMask.getHeight() << "\n";
-    masterMask.saveToFile(path);
-    // RawImage test(100, 100);
-    // test.saveToFile(path);
-}
+void ImageStack::saveMasterMask(const std::string& path) { masterMask.saveToFile(path, false); }
```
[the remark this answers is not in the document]
jeremykubica: Good call. I put together a pull request (#157) to rename all of these.

search/src/KBMOSearch.cpp
jeremykubica: The way the code was written there is only one activate timer in the KBMOSearch object. Since we are using a timer outside this function, we want that to be the active one.

search/src/image_kernels.cu
jeremykubica: I'm not seeing a good way to do that in git. There appears to be not 'git cp' type command. There are some work arounds online, but they didn't actually work for me.

search/src/RawImage.cpp
jeremykubica: Done.

maxwest-uw (approved):

PR #154, opened 2022-10-12
Update formatting of C++ Code

jeremykubica:
Uses clang-format to update the C++ code formatting to match LSST DM's style. Using instructions here: https://developer.lsst.io/cpp/clang-format.html

### DIFF
--- a/search/src/.clang-format
+++ b/search/src/.clang-format
@@ -1,0 +1,12 @@
+---
+Language: Cpp
+BasedOnStyle: Google
+ColumnLimit: 110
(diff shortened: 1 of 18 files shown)


maxwest-uw (approved):
since this is a code formatting change with no intentional code functionality changes, I pulled in the branch, ran the unit tests, checked to make sure kbmod gave the expected result on real observatory data (DECam), and looked over one of the cpp files I'm familiar with (Filtering.cpp) to make sure there wasn't anything untoward. everything ran well and looks good to me!

PR #155, opened 2022-10-12
Change handling of single/multi thread computation of PostProccess filtering + more tests

maxwest-uw:
- changed default number of cores from 30 to 1
- made it so that if the number of cores is set to 1, PostProcess will calculate filters single threaded for a speed boost
- added unit tests for filters in both single and multi thread

### DIFF
--- a/analysis/analysis_utils.py
+++ b/analysis/analysis_utils.py
@@ -680,17 +680,24 @@ class PostProcess(SharedTools):
             else:
                 self.percentiles = [25,75]
             self.coeff = self._find_sigmaG_coeff(self.percentiles)
-        print('Starting pooling...')
-        pool = mp.Pool(processes=self.num_cores)
+            
         num_curves = len(psi_curves)
         index_list = [j for j in range(num_curves)]
         zipped_curves = zip(
             psi_curves, phi_curves, index_list)
-        keep_idx_results = pool.starmap_async(
-            self._clipped_sigmaG, zipped_curves) 
-        pool.close()
-        pool.join()
-        keep_idx_results = keep_idx_results.get()
+        keep_idx_results = []
+        if self.num_cores > 1:  
+            print('Starting pooling...')
+            pool = mp.Pool(processes=self.num_cores)
+            keep_idx_results = pool.starmap_async(
+                self._clipped_sigmaG, zipped_curves) 
+            pool.close()
+            pool.join()
+            keep_idx_results = keep_idx_results.get()
+        else:
+            for z in zipped_curves:
+                keep_idx_results.append(self._clipped_sigmaG(z[0], z[1], z[2]))
+            
         end_time = time.time()
         time_elapsed = end_time-start_time
         print('{:.2f}s elapsed'.format(time_elapsed))
@@ -730,15 +737,22 @@ class PostProcess(SharedTools):
         results = old_results['results']
         keep = self.gen_results_dict()
 
-        print('Starting pooling...')
-        pool = mp.Pool(processes=self.num_cores)
         zipped_curves = zip(
             psi_curves, phi_curves, [j for j in range(len(psi_curves))])
-        keep_idx_results = pool.starmap_async(
-            self._clipped_average, zipped_curves)
-        pool.close()
-        pool.join()
-        keep_idx_results = keep_idx_results.get()
+        
+        keep_idx_results = []
+        if self.num_cores > 1:
+            print('Starting pooling...')
+            pool = mp.Pool(processes=self.num_cores)
+            keep_idx_results = pool.starmap_async(
+                self._clipped_average, zipped_curves)
+            pool.close()
+            pool.join()
+            keep_idx_results = keep_idx_results.get()
+        else:
+            for z in zipped_curves:
+                keep_idx_results.append(self._clipped_average(z[0], z[1], z[2]))
+        
         end_time = time.time()
         time_elapsed = end_time-start_time
         print('{:.2f}s elapsed'.format(time_elapsed))
@@ -986,6 +1000,7 @@ class PostProcess(SharedTools):
         print('---------------------------------------')
         print("Applying Stamp Filtering")
         print('---------------------------------------', flush=True)
+        start_time = time.time()
         i = 0
         passing_stamps_idx = []
         num_results = len(keep['results'])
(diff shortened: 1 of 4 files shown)


tests/test_filtering.py
```diff
@@ -1,9 +1,111 @@
...
+        # Select one pixel to mask in every other image.
+        self.masked_x = 5
+        self.masked_y = 6
+        
+        self.imlist = []
+        for i in range(self.imCount):
+            time = i/self.imCount
+            im = layered_image(str(i), self.dim_x, self.dim_y, 
+                                self.noise_level, self.variance, time,
+                                self.p)
```
jeremykubica, line 75: Small comment: There is an extra space at the start of lines 74 and 75. They should be aligned with str(i).

tests/test_filtering.py
```diff
@@ -1,9 +1,111 @@
...
+            im.add_object(self.start_x + time*self.x_vel+0.5,
+                           self.start_y + time*self.y_vel+0.5,
+                           self.object_flux)
```
jeremykubica, line 78: Small comment: There is an extra space at the start of lines 77 and 78. They should be aligned with self.start_x.

tests/test_filtering.py
```diff
@@ -1,9 +1,111 @@
...
+        self.config['filter_type'] = 'kalman'
+        self.config['max_lh'] = 1000.
+        
+        kb_p = PostProcess(self.config)
+        
+        self.stack = image_stack(self.imlist)
+        self.search = stack_search(self.stack)
+        self.search.search(self.angle_steps, self.velocity_steps,
+                         self.min_angle, self.max_angle, self.min_vel,
+                         self.max_vel, int(self.imCount/2))
```
jeremykubica, line 100: Align 99 and 100 with line 98 (add two spaces).

analysis/analysis_utils.py
```diff
@@ -999,23 +1014,35 @@ def apply_stamp_filter(
...
                 keep['stamps'].append(np.array(stamps_slice)[passing_stamps_chunk])
                 i+=chunk_size
             del(stamp_filt_results)
-            del(stamp_filt_pool)
         if len(keep['stamps']) > 0:
             keep['stamps'] = np.concatenate(keep['stamps'], axis=0)
             keep['final_results'] = np.unique(np.concatenate(passing_stamps_idx))
         print('Keeping %i results' % len(keep['final_results']), flush=True)
+        end_time = time.time()
+        time_elapsed = end_time-start_time
```
jeremykubica, line 1044: Add spaces around minus: "end_time-start_time" -> "end_time - start_time"

tests/test_filtering.py
```diff
@@ -1,9 +1,111 @@
...
+        # create a trajectory for the object
+        self.trj = trajectory()
+        self.trj.x = self.start_x
+        self.trj.y = self.start_y
+        self.trj.x_v = self.x_vel
+        self.trj.y_v = self.y_vel
+
+        # search parameters
+        self.angle_steps = 150
+        self.velocity_steps = 150
```
jeremykubica, line 60: Given that you know the object's real trajectory, you might be able to reduce the number of steps here and speed up the test. 150 is a reasonably large number of steps.

tests/test_filtering.py
```diff
@@ -1,9 +1,111 @@
...
+        self.min_vel = 5.0
+        self.max_vel = 40.0
+
+        # Select one pixel to mask in every other image.
+        self.masked_x = 5
+        self.masked_y = 6
+        
+        self.imlist = []
+        for i in range(self.imCount):
+            time = i/self.imCount
```
jeremykubica, line 72: Add a space around the division sign: "i/self.imCount" -> "i / self.imCount"

tests/test_filtering.py
```diff
@@ -1,9 +1,111 @@
...
+            im = layered_image(str(i), self.dim_x, self.dim_y, 
+                                self.noise_level, self.variance, time,
+                                self.p)
+            im.add_object(self.start_x + time*self.x_vel+0.5,
```
jeremykubica, line 76: Add spaces around the mathematical operators.

tests/test_filtering.py
```diff
@@ -1,9 +1,111 @@
...
+        self.config['cluster_type'] = None
+        self.config['cluster_function'] = None
+        self.config['clip_negative'] = None
+        self.config['mask_bits_dict'] = None
+        self.config['flag_keys'] = None
+        self.config['repeated_flag_keys'] = None
+        
+        # Set up old_results object for analysis_utils.PostProcess
+        psi_curves = [[1., 1., 1.], [1., 1., 1.], [1., 1., 1.]]
+        phi_curves = [[1., 1., 1.], [1., 1., 1.], [1., 1., 1.]]
```
jeremykubica, line 23: Would it be worth adding a 4th example that gets rejected by the filters?

tests/test_filtering.py
```diff
@@ -200,7 +302,62 @@ def test_calculate_likelihood_psiphi_zero_or_negative_phi(self):
...
-
+        
+    def test_apply_clipped_average_single_thread(self):      
+        kb_post_process = PostProcess(self.config)
+        
+        res = kb_post_process.apply_clipped_average(self.old_results, None, {}, 0.5)
+        
+        self.assertEqual(len(res), 3)
+        for r in res:
+            self.assertEqual(r[1], [-1])
```
jeremykubica, line 313: I'm not sure that I'm following this assert. I'm probably missing something. What is r[1]? From the code it looks like it is the indices that pass the filtering. But, if that's the case, how does it equal -1?

jeremykubica (commented):
This is a really nice change. I've added a bunch of stylistic comments.

The only blocking comment is the question in test_filtering.py on line 313. I want to make sure I understand what the tests are doing.

tests/test_filtering.py
maxwest-uw: [-1] is the list of passing indices for a given psi/phi curve when it's rejected, all of the results for these particular test curves are filtered out because that's easier to test but I think you're right that there should be a mix of passes and rejections for each test.

jeremykubica: If we could include a curve where not all results are filtered (it could have all results pass), that would give us better coverage on the tests.

maxwest-uw:
note: I looked back over the tests and thought it was weird that the clipped filtering filters out psi and phi curves that are all the same values.  i figured out why this is happening with clipped_average and changed the test data to reflect this but I'm still working on tweaking it for the other ones. it had to do with how the filtering removes the 'clipped' elements

maxwest-uw:
changed the filtering tests to get better coverage, added more comments, and moved the search for stamp filtering into the actual stamp filtering tests for a speed up :)

jeremykubica (approved):
This looks good. Thanks for expanding the test coverage.

PR #160, opened 2022-10-20
Clean up filtering config

jeremykubica:
This is a few small clean ups and doesn't change the functionality of the code:
- Clean up the naming in the filtering configuration to allow for future support of other filters (beyond just sigmaG).
- Remove some unused variables from analysis_utils.py
- Fix formatting in kernels.cu
- Change where load_results breaks so as not to load any filtered results.


search/src/kernels.cu
```diff
@@ -124,25 +124,25 @@ __global__ void searchFilterImages(int imageCount, int width, int height,
             phiSum += cPhi;
             psiArray[i] = cPsi;
             phiArray[i] = cPhi;
-            if (cPhi == 0.0) {
+            if (cPhi != 0.0) {
```
delucchi-cmu, line 127: This change looks scary!

If `cPhi` *is* zero, then we divide by it later...

analysis/analysis_utils.py
```diff
@@ -481,19 +481,21 @@ def load_results(
             print("---------------------------------------")
 
             for i, line in enumerate(results):
+                # Stop as soon as we hit a result below our limit, because anything after
+                # that is not guarrenteed to be valid due to potential on-GPU filtering.
+                if line.lh < lh_level:
```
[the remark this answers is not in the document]
jeremykubica: I guess I should say that this PR doesn't change the global behavior. Since the results with a likelihood under lh_limit get filtered later on. This just moves up the filtering.

search/src/kernels.cu
jeremykubica: Oh. This is a legitimate error that got in somehow. I'm not sure. But I've reverted it.

analysis/run_search.py
```diff
@@ -279,9 +279,9 @@ def do_gpu_search(self, search, img_info, ec_angle, post_process):
 
         # If we are using gpu_filtering, enable it and set the parameters.
         if self.config["gpu_filter"]:
-            print("Using in-line GPU filtering methods", flush=True)
+            print("Using in-line GPU sigmaG filtering methods", flush=True)
             self.config["sigmaG_coeff"] = post_process._find_sigmaG_coeff(self.config["sigmaG_lims"])
-            search.enable_gpu_filter(
+            search.enable_gpu_sigmag_filter(
```
maxwest-uw, line 284: i like the change here to specify what type of filtering is done on gpu!

analysis/analysis_utils.py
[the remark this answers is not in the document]
maxwest-uw: yeah i agree with jeremy, same behavior as before but makes the code a bit cleaner!

maxwest-uw (approved):
everything here looks good to me! (also this makes me think that I might go through analysis_utils and clean out all the unused variables since there are a lot of them iirc 😭 )

PR #161, opened 2022-10-20
Enable the analysis code to load per-image PSFs

jeremykubica:
Enable a file of per-image PSFs to be loaded. Images use the default PSF if no file is provided or the image name is not in the PSF file. Adds per-image PSFs to the regression test and allows them to be optionally specified for the diff_test.py.

Also cleans up the unused `fetch_wcs` parameter.


analysis/analysis_utils.py
```diff
@@ -171,7 +198,7 @@ def load_images(
         img_info.load_image_info_from_files(filenames)
 
         # Create a list of visit times and visit times shifts to 0.0.
-        img_info.set_times_mjd(np.array([image_time_dict[str(int(visit_id))] for visit_id in use_images]))
+        img_info.set_times_mjd(np.array(visit_times))
```
maxwest-uw, line 201: good change here

tests/regression_test.py
```diff
@@ -43,7 +43,7 @@ def make_trajectory(x, y, vx, vy, flux):
     return t
 
 
-def make_fake_image_stack(times, trjs):
+def make_fake_image_stack(times, trjs, psf_vals):
```
maxwest-uw, line 46: maybe I'm missing that one of the tests already covers this, but should we have one that tests run_search when a psf_file isn't provided?

jeremykubica: Good call. I updated the test so roughly 1/3 of the psfs will use the default and 2/3s will use the value from the file. Since the default is used when a file is not provided, I believe this should now cover that case. But let me know if you think it should be extended more.

maxwest-uw (approved):

Issue #163, opened 2022-10-21
Fix analysis_utils.py masking options


jeremykubica:
Analysis_utils.py's apply_mask function takes in, but does not use, the flux threshold. Either remove the option or apply it during masking.

jeremykubica:
Closed by #164

PR #164, opened 2022-10-21
Masking fix

### LINKED ISSUE #162: Fix ImageStack::simpleDifference to account for masking
Currently ImageStack::simpleDifference doesn't account for masked pixels and includes them in the average computation.

jeremykubica:
Fix a few errors that I discovered while writing documentation, including:
- The `mask_threshold `parameter is never applied.
- The global mask doesn't reset to all zeros when `createGlobalMask` is called and can lead to incorrect results if it is called multiple times.
- The functions `createAveTemplate` and `simpleDifference` do not currently account for masked pixels.
- Mask functions are applied even if they are no keys to mask (wasted cost).
- `avgTemplate` is not reset when `createAveTemplate` called leading to potential errors if it is called more than once.

Also adds a parameter `mask_grow` to indicate how much to extend the masked pixels.

Adds a lot of tests to cover these cases.

Closes #162 and #163


jeremykubica:
Need to fix something. Will resubmit.

jeremykubica:
Added NO_DATA support for the subtract template function in layered data.

maxwest-uw (approved):
nice changes! 2 things:
- will the change to apply_mask possibly change the number of results? might be worth letting dino test
- I like the new test_analysis_utils file, but we already have some analysis_tuils test in the test_filtering.py file. I think we should keep this new file tho and I can move all the tests from that over to this new one :)

jeremykubica:
The change to apply mask will only change the number of results if the user changes the parameters (specifically grow_mask). The additional checks only skip the masking step if there are no keys to mask (and thus the mask shouldn't be applied).

PR #165, opened 2022-10-25
Analysis cleanup

jeremykubica:
This change includes a variety of cleanups for kbmod:
- Allow the time file to be optional (using the data from the image file if a visit ID is not in the time file).
- Update the regression test to test for times not provided in the times file.
- Update the regression test to test filtering of images by data (`mjd_lims`).
- Clean up bug where `LayeredImage` time could be different than time in `ImageStack`. Always prefer the time in the time file if it exists.
- Remove unneeded loading helper functions and merge them into main load function (adds 1-2 lines to loader function). 
- Remove the unneeded `file_format` parameter (this information is already available in the function).
- Add usage and acknowledgements information to `known_objects.py`
- Remove unused `region_search` class. All core functionality remains. We are just removing the non-working python class.
- Add a setter function for `LayeredImage` time (and tests).
- Inline some common `LayeredImage` functions.
- Remove unneeded array creation from psi and phi curve generation.


maxwest-uw (approved):
love the changes to loading in image time stamps 😎

PR #169, opened 2022-10-28
Rework the build system

DinoBektesevic:
This is pip install-able on `baldur` with the mandatory requirement that `nvcc` compiler exists and is findable. For `baldur` that means prepending the path to `nvcc` to the `PATH`. Tests pass.

This Draft-PR is intended to spark discussion project layout and use and/or ease-of-use. It can still bear a lot of TLC and I still have to verify all the compile flags are the same.

Project structure was reorganized to match a more traditional Python project structure. Previously `analysis` and `search` were treated as 2 different projects in the same directory and were glued together by env var manipulations. They now build as a single project driven by setuptools with cmake. 

Cmake struggles detecting the correct Python - this will likely be the major reason for failed builds on different systems. There is no good tooling to achieve consistent Python env detection and `find_package( PythonInterp )` is, apparently, deprecated, but `find_package(Python COMPONENTS Interpretor/Module/Development)` will not find the correct Python most of the time. There is no easy fix for this because of Cmake's preference to build against the lowest consistent `glibc` used by dependencies. Since we have an old `anaconda` and a system-wide `nvcc`, it's a pretty old system-wide one on `baldur` and a luck of the draw on other systems.

The build will actually build 2 libraries: `kbmod` and `kbmodpy`, since `kbmod` is the namespace used in the previous `search` project and conflicts with the top-level project name. So, at the moment this will mean that `kbmodpy` and `kbmod` are importable separately, and that `kbmodpy` contains `kbmod` simultaneously. How to organize this is something we could discuss.

The python scripts built as part of `kbmodpy`, previously `analysis`, were used as an example only. It was my understanding that we're not sure how many scripts, and in what shape, we want to keep - so the use will be kludgy. Which and in what form we want to keep the existing `analysis` and `examples` scripts is something to revisit. 

The `analysis` scripts require an eclectic set of python libraries to work, I've listed them all as `required` for installation. We should separate those out into required and optional(s). I don't think we really need `ipywidgets` or `tqdm` under `required`.
The heavy hitting library is `tensorflow`, it would be nice if that can live under some other env name so that we don't have to install ~1GB of libraries when we test basic functionality. To get it to work, since it'll detect the system GPUs, make sure that alongside `nvcc` in `PATH` the correct `LD_LIBRARY_PATH` is set - if they're not already discoverable.

Some of the included Python scripts required `pyOrbfit` - so ensure correct paths for that are set as well or that it's discoverable. 

Couple of scripts that required the entire LSST stack to be installed have not been imported. I suggest you do the same for scripts that use PyOrbfit and/or Tensorflow if the installation doesn't go well in; see `kbmod/src/kbmodpy/__init__.py`

There is also an pretty old version of `cfitsio` that we link against that was vendored in `lib` from `search/pybinds/lib`.
It should be possible to build PyOrbfit and CFITSIO in a same way to pybinds11 and get rid of almost all of the env-var-kludge except requiring that a system-wide install of `nvcc` is present and accessible. In what way/shape/form is something that could be discussed.


maxwest-uw:
just glancing over this but these changes look great so far 😎 👍 I'll download and test it on Monday

PWhiddy:
Love this!
  
Couple of thoughts:

> Cmake struggles detecting the correct Python - this will likely be the major reason for failed builds on different systems. There is no good tooling to achieve consistent Python env detection and find_package( PythonInterp ) is, apparently, deprecated, but find_package(Python COMPONENTS Interpretor/Module/Development) will not find the correct Python most of the time. There is no easy fix for this because of Cmake's preference to build against the lowest consistent glibc used by dependencies. Since we have an old anaconda and a system-wide nvcc this is a pretty old system-wide one on baldur and the luck of the draw on other systems.

Are you hoping to eventually be able to build this with docker? That would eliminate the fiddliness of finding python versions and any other system dependencies, even cuda and nvcc. The only dependency that it can't capture is the gpu driver. [nvidia-docker2](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html#docker) can be used to run contianers which use GPUs. 

> The build will actually build 2 libraries: kbmod and kbmodpy, since kbmod is the namespace used in the previous search project and conflicts with the top-level project name. So, at the moment this will mean that kbmodpy and kbmod are importable separately, and that kbmodpy contains kbmod simultaneously. How to organize this is something we could discuss.

A better way to organize this could be using the convention discussed here:
https://github.com/pybind/pybind11/issues/1004
Basically, build cpp bindings as ```_kbmod```, do ```import from _kbmod import *``` in the core python library, and add ```_kbmod``` as an extension module to the package.

>There is also a pretty old version of cfitsio that we link against that was vendored in lib from search/pybinds/lib. It should be possible to build PyOrbfit and CFITSIO in a same way to pybinds11 and get rid of almost all of the env-var-kludge except requiring that a system-wide install of nvcc is present and accessible.

Definitely agree here, I originally just copied a cfitsio binary as a quick solution and it hasn't been updated since. If there's some repository for cfitsio which can be tracked that would be ideal.

DinoBektesevic:
@PWhiddy 

> Love this!

I love you :heart: 

> Are you hoping to eventually be able to build this with docker? That would eliminate the fiddliness of finding python versions and any other system dependencies, even cuda and nvcc. The only dependency that it can't capture is the gpu driver. [nvidia-docker2](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html#docker) can be used to run contianers which use GPUs.

One way of thinking about it is saying "this is how we envision this library to be used because...". Another might be to observe where, how and by whom we see the library being used (now or in the future). If majority of our users are whomever the users of LINCC-Hub are and LINCC-Hub adopts Docker I don't see a problem with this. If our users are running on clusters with a pre-configured queuing/scheduler/workflow management system, which isn't likely to adopt Docker, then perhaps it's better not to? It's a very good question and I'm not sure I have an, or am in a position to give, a definitive answer.        
Certainly I share your worries about fiddliness of the build system, but build systems,  once deployed, mature and it's not impossible, although cumbersome, to resolve these edge-cases. 

> A better way to organize this could be using the convention discussed here: [pybind/pybind11#1004](https://github.com/pybind/pybind11/issues/1004) Basically, build cpp bindings as `_kbmod`, do `import from _kbmod import *` in the core python library, and add `_kbmod` as an extension module to the package.

This is also a great comment. Since I'm not that sure what is "everything" KBMoD should do I don't have a good overview of how we should organize modules. Since the directory and much of that project seems to have adopted `search` as the name we could also have:
```
kbmod
kbmod.search
kbmod.analysis
kbmod.something
```
layout. A longer-time user or developer, i.e. you, is probably in a better position to describe what they want kbmod to be. I am flexible on the issue. It's a bit short-term, but if you see this, hopefully before our meeting today, and come up with a general module naming and layout scheme I'd be happy to adopt it.

maxwest-uw:
nice change with making our own version of the cuda errors 😎 👍 . also from my monkeying around with the code it seems like the build system changes have made importing kbmod much faster! it used to take a second or so before but now it's near instantaneous

PWhiddy:
> One way of thinking about it is saying "this is how we envision this library to be used because...". Another might be to observe where, how and by whom we see the library being used (now or in the future). If majority of our users are whomever the users of LINCC-Hub are and LINCC-Hub adopts Docker I don't see a problem with this. If our users are running on clusters with a pre-configured queuing/scheduler/workflow management system, which isn't likely to adopt Docker, then perhaps it's better not to? It's a very good question and I'm not sure I have an, or am in a position to give, a definitive answer.
> Certainly I share your worries about fiddliness of the build system, but build systems, once deployed, mature and it's not impossible, although cumbersome, to resolve these edge-cases.

Gotcha. If docker isn't available where the software is deployed, it might be less worthwhile. It still could be useful for purposes of development, using it to build artifacts for the core library, run CI/tests, or if at some point someone wants to deploy on cloud servers.

> Since I'm not that sure what is "everything" KBMoD should do I don't have a good overview of how we should organize modules
...
> A longer-time user or developer, i.e. you, is probably in a better position to describe what they want kbmod to be. I am flexible on the issue. It's a bit short-term, but if you see this, hopefully before our meeting today, and come up with a general module naming and layout scheme I'd be happy to adopt it.

It does depend on the planned future use, but the naming scheme you proposed seems reasonable. Feel free to go ahead with that, or we could discuss in the next meeting

DinoBektesevic:
Ok, I think this is far enough for 1 PR. We can add cfitsio and PyOrbfit as dependencies in individual PRs, pending next Monday's discussion, and we need to sort out whatever is happening in `kbmodpy.py`. 

The latest goodies are:
    - cleanup of warnings from build script and better cmake policy setting
    - splitting the installation requirements into `required` and `analysis` (pip install .[analysis] for the full list of packages)
    - Splitting the `analysis` scripts into top-level and hidden ones based on usage - this is to make life bearable otherwise imports and compiles are long. "Usage" of a script is based on what currently runs in the `tests` and what doesn't. 
    - Cleanup of imports and removal of not really used packages. TODO get rid of pandas. 
    - SCM versioning was added so now we have `>>>  kbmod.\_\_version\_\_` print out `0.5.1.dev339+g26880ce.d20221103` based off of github tags. Autoupdated.

It would be really cool if people tested this against different cmakes, pythons and envs and reported back.

.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "search/pybinds/pybind11"]
-	path = search/pybinds/pybind11
-	url = https://github.com/pybind/pybind11.git
+[submodule "lib/pybind11"]
+	path = lib/pybind11
+	url = git@github.com:pybind/pybind11.git
```
jeremykubica, line 3: I'm curious what the difference is between these two URLs? The new URL is giving me a "can't verify host" type warning message whereas the old URL works fine.

src/kbmod/analysis/__init__.py
```diff
@@ -0,0 +1,22 @@
+# we skip all imports because of the large
```
jeremykubica, line 1: Should all these lines be deleted instead of commented out?

jeremykubica (commented):
One more request: Can you update the README.md file to reflect these changes. It would be great if it could include information on:
1) How to install from complete scratch.
2) How to rebuild and test if you change the C++ files
3) How to just run the tests if you change the python files (i.e. `cd tests/` and `python -m unittest`)

setup.cfg
```diff
@@ -0,0 +1,57 @@
+[metadata]
+name = kbmod
+authors = [
+    { name = "KBMoD Developers", email="tmp@email.email"},
+    ]
```
DinoBektesevic, line 5: We need to fill in some authors, but I wasn't sure what the base way of doing that was. We can list everyone individually, but that list might grow substantial (a little bit already is) or we can just write "KBMoD Developers" without an email. This is how all the other projects (astropy numpy etc...) did it.

.gitmodules
DinoBektesevic: This was a mistake on my part, I had an very old default git that didn't support `git mv` of submodules so I had to edit by hand. Returned to `https`

setup.cfg
DinoBektesevic: Thanks for the votes! "KBMoD Developers" wins.

jeremykubica (approved):

PR #170, opened 2022-10-31
Move Clipped Average Filter to C++

jeremykubica:
- Results in a >10x speed up in execution.
- Include logic to handle NO_DATA tags (which aren't passed by the python code yet)
- Add more tests to confirm the logic

### DIFF
--- a/search/src/Filtering.cpp
+++ b/search/src/Filtering.cpp
@@ -46,6 +46,98 @@ std::vector<int> sigmaGFilteredIndices(const std::vector<float>& values, float s
     return result;
 }
 
+std::vector<int> clippedAverageFilteredIndices(const std::vector<float>& psi_curve,
+                                               const std::vector<float>& phi_curve,
+                                               int num_clipped, int n_sigma, float lower_lh_limit) {
+    /*
+     * This function applies a clipped median filter to a set of likelihood values. The largest
+     * likelihood values (N=num_clipped) are eliminated if they are more than n_sigma*st_dev
+     * away from the median, which is calculated excluding the largest values.
+     *
+     * Arguments:
+     *     psi_curve - The psi values.
+     *     phi_curve - The phi values.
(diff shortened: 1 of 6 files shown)


search/src/Filtering.cpp
```diff
@@ -46,6 +46,98 @@ std::vector<int> sigmaGFilteredIndices(const std::vector<float>& values, float s
...
+    sort(scores.begin(), scores.end());
+
+    // Find all the indices where:
+    // 1) LH > lower_lh_limit
+    // 2) They do not have one of the top 'num_clipped' likelihoods.
+    // 3) LH != NO_DATA
+    int first_valid = 0;
+    int last_valid = 0;
+    double lh_sum_good = 0.0;
+    for (int t = 0; t < num_times - num_clipped; ++t) {
```
cchris28, line 92: did you consider using an iterator here? I am curious if it makes it any faster or if the compiler optimizes it to be equivalent.

jeremykubica: I didn't benchmark that particular change.

search/src/Filtering.h
```diff
@@ -9,6 +9,7 @@
 #ifndef FILTERING_H_
 #define FILTERING_H_
 
+#include <queue>
```
maxwest-uw, line 12: maybe I'm missing something, but I don't think we're using a queue anywhere in these changes, is it necessary for anything else?

maxwest-uw (commented):
one super small question :) everything else looks good

search/src/Filtering.h
jeremykubica: Good catch. That is a mistake. I originally added it, thinking I would use a queue like the python approach. But it turns out that it is easier to use a sorted array multiple times and just track indices. So I never ended up using the queue. I've removed the include.

maxwest-uw (approved):

PR #171, opened 2022-11-01
Reduce test flakiness

jeremykubica:
Reduce test flakiness by reducing the amount of Gaussian noise in the images and increasing object brightness.


DinoBektesevic (approved):
I'm a little bit confused as to why we increase `i` counter in the test by 5. isn't it better to have the same seed everywhere so that tests don't change if the number of images changes? 

PR looks just fine to me otherwise.

jeremykubica:
The "+ 5" is completely arbitrary. I removed it to avoid confusion. 

We do want to have different seeds for each fake image to provide a more realistic test of varying background noise. But on any given run we want the same seed for image[k] as we used for the previous runs so that the test itself isn't random (and gives the same results every time). If we add more images we want them to use different backgrounds from each other, but the same backgrounds in all future runs.

PR #174, opened 2022-11-02
Split the region search into its own class

jeremykubica:
Make the division in logic clearer between the linear search and the region search by breaking it into its own class.


PWhiddy (approved):
Besides a couple method access qualifiers, code isn't changed just moved? If so LGTM.

jeremykubica:
You are correct the internals of the functions have not changed. Only the qualifiers (KBMOSearch to KBMORegionSearch) were changed.

PR #175, opened 2022-11-04
Clean ups and refactoring pre-work

jeremykubica:
- Rerun clang format on the files (lots of small style improvements).
- Add a new TrajectoryResults class that we can later use to pass around the combination of trajectories + valid indices.
- Modify the stamp production functions to use TrajectoryResults, but keep the functions as thin wrappers so that we do not need to modify the python code yet.

### DIFF
--- a/src/kbmod/search/KBMOSearch.cpp
+++ b/src/kbmod/search/KBMOSearch.cpp
@@ -233,68 +233,145 @@ void KBMOSearch::fillPsiAndPhiVects(const std::vector<RawImage>& psiImgs,
             phiVect->push_back(phiRef[p]);
         }
     }
+} 
+
+std::vector<RawImage> KBMOSearch::scienceStamps(const TrajectoryResult& trj, int radius, bool interpolate,
+                                                bool keep_no_data, bool all_stamps) {
+    std::vector<RawImage> stamps;
+    int num_times = stack.imgCount();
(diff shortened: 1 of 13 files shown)


delucchi-cmu (approved):
I love a good prefactor.

src/kbmod/search/KBMOSearch.cpp
```diff
@@ -235,60 +235,134 @@ void KBMOSearch::fillPsiAndPhiVects(const std::vector<RawImage>& psiImgs,
...
+
+RawImage KBMOSearch::meanScienceStamp(const TrajectoryResult& trj, int radius, bool use_all) {
+    return createMeanImage(scienceStamps(trj, radius, false, true, use_all));
+}
+
+std::vector<RawImage> KBMOSearch::meanScienceStamps(const std::vector<TrajectoryResult>& t_array,
+                                                    int radius) {
+    const int num_results = t_array.size();
+    std::vector<RawImage> results(num_results);
     omp_set_num_threads(16);
```
drewoldag, line 306: Just curious how you choose the number of threads for these calls `omp_set_num_threads(xx)`

jeremykubica: This has been carried over from the original code. It is unlikely to be optimal and we should do something more adaptive here.

maxwest-uw (approved):

PR #179, opened 2022-11-10
Provide a function to create coadded stamps (sum and mean) on GPU

jeremykubica:
The change also includes some cleanup of includes / definitions needed to include ImageStack in one of the .cu files.

### DIFF
--- a/src/kbmod/search/KBMOSearch.cpp
+++ b/src/kbmod/search/KBMOSearch.cpp
@@ -316,19 +316,22 @@ std::vector<RawImage> KBMOSearch::meanScienceStamps(const std::vector<Trajectory
 
 std::vector<RawImage> KBMOSearch::coaddedScienceStampsGPU(std::vector<trajectory>& t_array,
                                                           int radius, bool compute_mean) {
+    // Right now only limited stamp sizes are allowed.
+    if (2 * radius + 1 > MAX_STAMP_EDGE || radius <= 0) {
+        throw std::runtime_error("Invalid Radius.");
+    }
+
     const int num_images = stack.imgCount();
     const int width = stack.getWidth();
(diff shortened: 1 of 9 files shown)


src/kbmod/search/KBMOSearch.cpp
```diff
@@ -314,6 +314,55 @@ std::vector<RawImage> KBMOSearch::meanScienceStamps(const std::vector<Trajectory
...
 }
 
+std::vector<RawImage> KBMOSearch::coaddedScienceStampsGPU(std::vector<trajectory>& t_array,
+                                                          int radius, bool compute_mean) {
+    const int num_images = stack.imgCount();
+    const int width = stack.getWidth();
+    const int height = stack.getHeight();
+
+    // Copy the image pixels into a single array.
+    unsigned int ppi = width * height;
```
cchris28: how often does this get called compared to how often the result is used? Might be a couple of extra copies involved here (or not, hard to tell from just this code).

jeremykubica: It won't be called too often, but I would like to remove any unneeded copies.

The specific line you are commenting on seems to be from an old copy of the PR though. I removed the copy of image pixels into a single array part.

maxwest-uw (approved):
lgtm!

cchris28 (commented):
FYI, an alternative to the familiar idiom (provided we are not restricted to ancient versions of the compiler):
  for (int i = 0; i < size(d); ++i)
is
  for (const auto& i: d)
This is useful when d is not modified by the clause AND you do not care (much) about the current index. It can lead to some optimizations. It requires that d is iterable. But that is generally a good idea anyway if you are using containers.

maxwest-uw (approved):

PR #180, opened 2022-11-21
Update GPU Filtering

jeremykubica:
Update the on GPU filtering to not count NO_DATA in the statistics computation.


src/kbmod/search/kernels.cu
```diff
@@ -139,61 +140,20 @@ __global__ void searchFilterImages(int imageCount, int width, int height,
...
-                int idx = idxArray[i];
-                if (lcArray[idx] <= maxValue) {
-                    maxKeepIndex = i;
-                } else {
-                    break;
-                }
-            }
-            
+            int maxKeepIndex = num_seen - 1;
+            sigmaGFilteredIndicesCU(lcArray, num_seen, params.sGL_L, params.sGL_H,
```
maxwest-uw, line 148: nice, I was wondering why we had separate sigma filtering code when we already had a function for it on GPU 👍 this cleans stuff up a lot.

maxwest-uw (approved):

PR #181, opened 2022-11-21
Extend the GPU Stamp generation to do filtering

jeremykubica:
Do stamp generation on GPU to provide a further speedup.

### DIFF
--- a/src/kbmod/search/KBMOSearch.cpp
+++ b/src/kbmod/search/KBMOSearch.cpp
@@ -14,8 +14,8 @@ extern "C" void deviceSearchFilter(int imageCount, int width, int height, float*
                                    trajectory* trajectoriesToSearch, int resultsCount,
                                    trajectory* bestTrajects);
 
-void deviceGetCoadds(ImageStack& stack, perImageData image_data, int radius, bool do_mean,
-                     int num_trajectories, trajectory *trajectories,
+void deviceGetCoadds(ImageStack& stack, perImageData image_data, int num_trajectories,
+                     trajectory *trajectories, stampParameters params,
                      std::vector<std::vector<bool> >& use_index_vect, float* results);
 
 
@@ -326,9 +326,9 @@ std::vector<RawImage> KBMOSearch::meanScienceStamps(const std::vector<Trajectory
 
 std::vector<RawImage> KBMOSearch::coaddedScienceStampsGPU(std::vector<trajectory>& t_array,
                                                           std::vector<std::vector<bool> >& use_index_vect,
-                                                          int radius, bool compute_mean) {
+                                                          stampParameters& params) {
     // Right now only limited stamp sizes are allowed.
-    if (2 * radius + 1 > MAX_STAMP_EDGE || radius <= 0) {
+    if (2 * params.radius + 1 > MAX_STAMP_EDGE || params.radius <= 0) {
         throw std::runtime_error("Invalid Radius.");
     }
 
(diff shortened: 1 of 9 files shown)


src/kbmod/search/KBMOSearch.cpp
```diff
@@ -340,40 +340,47 @@ std::vector<RawImage> KBMOSearch::coaddedScienceStampsGPU(std::vector<trajectory
...
+            results[t] = RawImage(stamp_width, stamp_width, current_pixels);
         }
-        results[t] = RawImage(stamp_width, stamp_width, current_pixels);
     }
     return results;
 }
 
 std::vector<RawImage> KBMOSearch::coaddedScienceStampsGPU(std::vector<trajectory>& t_array,
-                                                          int radius, bool compute_mean) {
+                                                          stampParameters& params) {
```
cchris28, line 376: If params is read only then consider making it a const &

src/kbmod/search/image_kernels.cu
```diff
@@ -236,9 +236,88 @@ extern "C" void deviceGrowMask(int width, int height, float *source,
...
+    // Initialize the variables for tracking the peak's location.
+    pixelPos result = { 0, 0 };
+    float max_val = img[0];
+    float dist2 = c_x * c_x + c_y * c_y;
+
+    // Search each pixel for the peak.
+    for (int y = 0; y < height; ++y) {
+        for (int x = 0; x < width; ++x) {
+            if (img[y * width + x] > max_val) {
+                max_val = img[y * width + x];
```
PWhiddy, line 253: The compiler likely is going to keep `img[y * width + x]` in a register for reuse, but it might be good to capture it into a variable just to be sure this doesn't cause multiple global memory reads.

src/kbmod/search/image_kernels.cu
```diff
@@ -236,9 +236,88 @@ extern "C" void deviceGrowMask(int width, int height, float *source,
     checkCudaErrors(cudaFree(deviceResultImg));
 }
 
+extern "C" __device__ __host__ pixelPos findPeakImageVect(int width, int height, float *img,
```
PWhiddy, line 239: I like how these inner functions are cross-compiled to cpu for testing! That's a great idea.

src/kbmod/search/common.h
```diff
@@ -111,6 +113,34 @@ struct trajRegion {
     float flux;
 };
 
+struct stampParameters {
+    int radius = 10;
+    StampType stamp_type = STAMP_SUM;
+    bool do_filtering = false;
```
PWhiddy, line 119: For some reason, I'm only just now noticing lots of snake_case naming has been added to the c++ code. Should there be a preferred convention going forward?

jeremykubica: Good point. In general we have been moving the code toward LSST DM's style guide (https://developer.lsst.io/cpp/style.html#variable-names-must-be-camelcase-or-snake-case), but that supports both camelcase and snakecase. I am fine with either, but tend to default to snake case. I would be interested to hear if there is a general preference.

src/kbmod/search/image_kernels.cu
jeremykubica: Good call. I fixed it.

src/kbmod/search/KBMOSearch.cpp
jeremykubica: Done

PWhiddy (approved):

Issue #106, opened 2021-07-21
1-3 px systematic offset between KBMOD reported trajectory and scipy minimize fine-grid barycentric trajectory


smotherh:
When comparing the result of the KBMOD trajectory to the fine-grid barycentric scipy minimize best-fit result, there appears to be a 1-3 pixel systematic offset in the object center. It is unclear what is causing this, but the disagreement seems too large to be the result of the barycentric correction in DEEP data, given the short (~4hr) time baseline.

Potential causes include:

- Off-by-one errors in computing starting pixel
- Off-by-one error in generating stamps in post-processing (not KBMOD stamps)
- Small PSF errors
- Inherent KBMOD limitation due to selecting starting pixel location as a discrete value (instead of a fractional float)
- Error in the warp WCS

PWhiddy:
@smotherh Is the offset always in the same direction?

smotherh:
Just commenting that this is still present. @PWhiddy yes, the offset is always in the same direction. It is most commonly a 1px offset. I think 2 or 3 px offsets are due to randomness. i.e. Sometimes the grid search snaps to a neighboring trajectory.

It looks more like an off-by-one error or a zero versus one indexing somewhere. It's not really a problem, since the joint-fit fixes this problem very well, and we need to run the jointfit regardless.

PWhiddy:
Thinking about this again, I'm recalling a similar discussion with @jbkalmbach about small systematic offsets in calculated pixel coordinates. IIRC the issue was that we were using different conventions for fractional pixel coordinates. The reason was that one set of coordinates were shifted 0.5 in x and y because of different ideas of what the "center" of a pixel ought to be. I'm sure Bryce was right and I was wrong, but it's possible that discrepancy still exists between the search and analysis code.

jeremykubica:
Closing since we removed the barycentric code (and are doing reprojection instead).

PR #182, opened 2022-11-22
Fix a bunch of bugs in stamp generation.

jeremykubica:
- mean and median stamps were ignoring goodIdx (introduced in PR-175)
- on-GPU stamp creation was using 0.5 shifts while the previous stamp generation was not. Make the on-GPU stamp generation consistent with the current stamp generation.
- One of the TrajectoryResult constructors had an allocation bug for the vector.
- Increase test coverage to include goodIdx for mean and median stamps.
- Add parity tests between the CPU and GPU stamp creation functions.

### DIFF
--- a/src/kbmod/search/KBMOSearch.cpp
+++ b/src/kbmod/search/KBMOSearch.cpp
@@ -285,7 +285,7 @@ std::vector<RawImage> KBMOSearch::medianScienceStamps(const std::vector<Trajecto
 
 #pragma omp parallel for
     for (int s = 0; s < num_results; ++s) {
-        results[s] = medianScienceStamp(t_array[s], radius, true);
+        results[s] = medianScienceStamp(t_array[s], radius, false);
     }
     omp_set_num_threads(1);
 
@@ -317,7 +317,7 @@ std::vector<RawImage> KBMOSearch::meanScienceStamps(const std::vector<Trajectory
 
 #pragma omp parallel for
     for (int s = 0; s < num_results; ++s) {
-        results[s] = meanScienceStamp(t_array[s], radius, true);
+        results[s] = meanScienceStamp(t_array[s], radius, false);
     }
     omp_set_num_threads(1);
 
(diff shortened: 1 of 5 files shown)


src/kbmod/search/KBMOSearch.cpp
```diff
@@ -317,7 +317,7 @@ std::vector<RawImage> KBMOSearch::meanScienceStamps(const std::vector<Trajectory
 
 #pragma omp parallel for
     for (int s = 0; s < num_results; ++s) {
-        results[s] = meanScienceStamp(t_array[s], radius, true);
+        results[s] = meanScienceStamp(t_array[s], radius, false);
```
maxwest-uw, line 320: ok so just to repeat back the logic of this fix to make sure we're on the same page, the previous change made it so that median stamps no longer only use GoodIdx, so they need to not use all stamps (hence this and the above change to false).

maxwest-uw (commented):
I think this looks good, just want to confirm I got the logic right!

PWhiddy:
Possibly related to #106?

src/kbmod/search/KBMOSearch.cpp
jeremykubica: Correct. Before 175 meanStamps contained the following loop:
`for (int i = 0; i < goodIdx[s].size(); ++i) {`
`if (goodIdx[s][i] == 1) {`
`pixelPos pos = getTrajPos(t, i);`
`stamps.push_back(imgs[i].getScience().createStamp(pos.x, pos.y, radius, false, true));`
`}`
`}`
So each time step was checked against `goodIdx` before pushing it into the array of stamps. And the mean/median was built on only those images inserted into the array.

This pruning logic was moved to `scienceStamps ` (line 254), but only applies if `all_stamps` is false. Otherwise, as the name implies, it uses all the stamps for the mean / median. I believe the error was passing true to `meanScienceStamp` which forces it to use all of the indices.

jeremykubica:
> 

Responding to PWhiddy's comment: It is possible. The 0.5 is applied during the search, but not during stamp creation. This is only 0.5 of a pixel, so I would be surprised if it made a 3 pixel difference. But we could try adding the 0.5 everywhere and see if helps the problem. I'd like to do that as a separate change though and keep the behavior consistent with the previous versions during this change.

maxwest-uw (approved):

PR #183, opened 2022-11-28
Refactor GPU Stamp Creation

jeremykubica:
Refactor the GPU stamp creation to support median stamps. This breaks the stamp creation and stamp filtering into two different GPU steps (that can be done without copying the image data twice though).

This also makes a few small changes to on-GPU stamp creation and filtering to match the CPU code (> to >= and moving where we round the predicted position to an int).

Add a bunch of new tests to confirm the behavior.

### DIFF
--- a/src/kbmod/search/image_kernels.cu
+++ b/src/kbmod/search/image_kernels.cu
@@ -8,7 +8,7 @@
 #ifndef IMAGE_KERNELS_CU_
 #define IMAGE_KERNELS_CU_
 
-#define MAX_STAMP_IMAGES 140
+#define MAX_STAMP_IMAGES 200
 
 #include <assert.h>
 #include "common.h"
@@ -327,7 +327,6 @@ __global__ void device_get_coadd_stamp(int num_images, int width, int height, fl
     if (trj_index < 0 || trj_index >= num_trajectories)
         return;
     trajectory trj = trajectories[trj_index];
-    int use_index_offset = num_images * trj_index;
 
     // Get the pixel coordinates within the stamp to use.
     const int stamp_width = 2 * params.radius + 1;
@@ -339,7 +338,13 @@ __global__ void device_get_coadd_stamp(int num_images, int width, int height, fl
     if (stamp_y < 0 || stamp_y >= stamp_width)
         return;
 
+    // Compute the various offsets for the indices.
+    int use_index_offset = num_images * trj_index;
+    int trj_offset = trj_index * stamp_width * stamp_width;
+    int pixel_index = stamp_width * stamp_y + stamp_x;
+
     // Allocate space for the coadd information.
+    assert(num_images < MAX_STAMP_IMAGES);
     float values[MAX_STAMP_IMAGES];
 
     // Loop over each image and compute the stamp.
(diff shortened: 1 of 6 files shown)


src/kbmod/search/image_kernels.cu
```diff
@@ -8,6 +8,8 @@
 #ifndef IMAGE_KERNELS_CU_
 #define IMAGE_KERNELS_CU_
 
+#define MAX_STAMP_IMAGES 140
```
delucchi-cmu, line 11: How was this number determined? What might happen if someone changes it?

src/kbmod/search/image_kernels.cu
```diff
@@ -347,74 +352,109 @@ __global__ void device_get_coadd_stamp(int num_images, int width, int height, fl
...
-                if ((img_x >= 0) && (img_x < width) && (img_y >= 0) && (img_y < height)) {
-                    int pixel_index = pixels_per_image * t + img_y * width + img_x;
-                    if (image_vect[pixel_index] != NO_DATA) {
-                        int stamp_index = stamp_y * stamp_width + stamp_x;
-                        sum[stamp_index] += image_vect[pixel_index];
-                        count[stamp_index] += 1.0;
-                    }
-                }
+        // Get the stamp and add it to the list of values.
+        int img_y = currentY - params.radius + stamp_y;
```
delucchi-cmu, line 364: NIT - swap the lines for `int img_y = ...` and `int img_x = ...`. It's weird to see them in this order and I did a double take.

src/kbmod/search/image_kernels.cu
```diff
@@ -347,74 +352,109 @@ __global__ void device_get_coadd_stamp(int num_images, int width, int height, fl
...
+        }
     }
 
-    // Compute the mean if needed.
-    if (params.stamp_type == STAMP_MEAN) {
-        for (int i = 0; i < stamp_ppi; ++i) {
-            sum[i] = (count[i] > 0.0) ? sum[i]/count[i] : 0.0;
+    // Do the actual computation from the values.
+    float result = 0.0;
+    if ((params.stamp_type == STAMP_MEDIAN) && (num_values > 0)) {
```
delucchi-cmu, line 377: Since the following ~20-30 lines don't many any sense if `num_values == 0`, maybe just go ahead and return if that's the case (or set results[thing] == 0). I think it makes the flow read more naturally. Or you could turn the remainder into a proper switch/case clause.

delucchi-cmu:
Just readability nits. Otherwise LGTM.

src/kbmod/search/image_kernels.cu
jeremykubica: Did both (early return and switch/case)

src/kbmod/search/image_kernels.cu
jeremykubica: Done

src/kbmod/search/image_kernels.cu
jeremykubica: This was set to equal the number used in the main search algorithm. It balances the memory used on the GPU with the number of images supported. Increasing the number just uses more memory. Decreasing it could cause a problem, so I added an assert to check that it is valid in case someone reduces it too much.

maxwest-uw (approved):
lgtm!

PR #185, opened 2022-11-29
Clean up MakeStamp API

jeremykubica:
Cleans up the functions in KBMOSearch to remove duplication and improve clarity. The historical functions such as `medianStamps` and `meanStamps` are removed and the python code is switched over to use the newer functions (that use `TrajectoryResults`).

I further tested this by running the diff test with all three stamp types and confirming that **all** of the output files were identical to the old code (including the stamp files).

### DIFF
--- a/src/kbmod/search/KBMOSearch.cpp
+++ b/src/kbmod/search/KBMOSearch.cpp
@@ -260,17 +260,9 @@ std::vector<RawImage> KBMOSearch::scienceStamps(const TrajectoryResult& trj, int
     return stamps;
 }
 
-inline std::vector<RawImage> KBMOSearch::scienceStampsForFilter(const TrajectoryResult& trj, int radius) {
-    return scienceStamps(trj, radius, false, true, false);
-}
-
-inline std::vector<RawImage> KBMOSearch::scienceStampsForViz(const TrajectoryResult& trj, int radius) {
-    return scienceStamps(trj, radius, true, false, true);
-}
-
-std::vector<RawImage> KBMOSearch::scienceStamps(trajectory& t, int radius) {
+std::vector<RawImage> KBMOSearch::scienceStampsForViz(const trajectory& t, int radius) {
     TrajectoryResult trj(t, stack.imgCount());
-    return scienceStampsForViz(trj, radius);
+    return scienceStamps(trj, radius, true, false, true);
 }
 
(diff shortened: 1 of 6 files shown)


src/kbmod/search/KBMOSearch.cpp
```diff
@@ -398,27 +377,8 @@ std::vector<RawImage> KBMOSearch::coaddedScienceStampsGPU(std::vector<Trajectory
...
-
 RawImage KBMOSearch::summedScienceStamp(const TrajectoryResult& trj, int radius, bool use_all) {
-    return createSummedImage(scienceStamps(trj, radius, false, true, use_all));
-}
-
-// To be deprecated in later PR.
-RawImage KBMOSearch::stackedScience(trajectory& t, int radius) {
-    TrajectoryResult trj(t, stack.imgCount());
-    return createSummedImage(scienceStamps(trj, radius, false, false, true));
+    return createSummedImage(scienceStamps(trj, radius, false, false, use_all));
```
delucchi-cmu, line 381:
```suggestion
    return createSummedImage(scienceStamps(trj, radius, false /*=interpolate*/, false /*=keep_no_data*/, use_all));
```

Reducing to one implementation method looks great, but the calls can be confusing if they're just a few booleans. Adding some context to the method calls would improve readability a lot here.

delucchi-cmu:
Is there a test bed where the old and new files can be generated and compared? Or was it just some nice bash command to check files were identical? Is there some place these kinds of tips and tricks live for kbmod developers?

jeremykubica:
> Is there a test bed where the old and new files can be generated and compared? Or was it just some nice bash command to check files were identical? Is there some place these kinds of tips and tricks live for kbmod developers?

Yes. `tests/diff_test.py` does this for the main results and coadded stamp files. You need to run it once with `generate_goldens` on your data set of choice, but after that you can automatically compare. There are a bunch of other files that aren't tested because they're information is already included in the two above. But they can be manually confirmed with something like `diff tmp/phi_science_validation_short.txt goldens/phi_science_validation_short.txt`

src/kbmod/search/KBMOSearch.cpp
jeremykubica: Good point. I expanded on this suggestion by not only including the in-argument comments, but also providing some comments about "why things work they way they do." I spent a lot of time tracing through the functions to understand the historical behavior (which functions used which options), so I think more detail here would be very helpful in the future.

maxwest-uw (approved):

PR #189, opened 2022-12-02
Switch analysis_utils to use gpu stamp generation/filtering be default

jeremykubica:
This change does a few large-ish things. First it switches analysis_utils to use the on-GPU stamp generation and filtering by default. This will add a little overhead to runs with few results, but the stamp generation isn't large portion of compute time there anyway. It will significantly speed up the stamp filtering step for runs with lots of results.

This also fixes a bug in the original code with how stamp filtering was broken into chunks. Before the stamps were processed in chunks to save memory, but the generation did not use the correct matching lc_index for each result (it would pass a subset of N results and use always use the lc_index for 0:N results). 

Change also includes some formatting cleanups and removing one compiler warning.


maxwest-uw (approved):
excellent 👌

PR #193, opened 2022-12-12
Change ResultList to store the times

jeremykubica:
Currently every ResultRow stores a list of all times, but this is redundant since they are all the same. This change updates ResultRow to just take the number of times and to accept all_times only when outputting a list of valid times. A list of all times is stored in ResultList instead.


src/kbmod/analysis_utils.py
```diff
@@ -1028,28 +1026,23 @@ def _cluster_results(self, results, x_size, y_size, v_lim, ang_lim, mjd_times, c
...
-                individual cluster.
+
+        Arguments:
+            results (list): A list of kbmod trajectory results such as are
+                stored in keep['results'].
+            x_size (int):  The width of the images used in the kbmod stack,
+                such as are stored in image_params['x_size'].
+            y_size (int):  The height of the images used in the kbmod stack,
+                such as are stored in image_params['y_size'].
+            v_lim (list):  The velocity limits of the search, such as are
```
delucchi-cmu, line 1037: It's weird and not immediately obvious why some of the args are ints and some lists (and it's different from the original docstring).

jeremykubica: I expanded the comments to explain why v_lim and ang_lim are lists and what they contain. The x_size and y_size were incorrectly described as lists in the former comment.

PR #194, opened 2022-12-13
Finish refactoring of analysis_utils.py result storage

jeremykubica:
This change transitions all of analysis_utils.py to use ResultList for tracking and passing around results. The goal is to provide a separate interface for storing and filtering results so that the user does not need to manually update the relevant fields of the `keep` dictionary.  The tradeoff will be occasionally higher overhead due to storing the results in row-wise format.


src/kbmod/analysis_utils.py
```diff
@@ -372,325 +358,213 @@ def load_results(
...
         start = time.time()
-        num_images = search.get_num_images()
 
-        # Create a subset of the trajectories and good_idx for results in result_idx.
-        res_subset = [keep["results"][ind] for ind in result_idx]
-        good_subset = [keep["lc_index"][ind] for ind in result_idx]
+        # Create the list of trajectory results to use.
+        trj_list = keep.trj_result_list(indices_to_use=result_idx)
 
+        # Make the stamps.
```
delucchi-cmu, line 453: This is a super-satisfying cleanup!

src/kbmod/analysis_utils.py
```diff
@@ -372,325 +358,213 @@ def load_results(
...
-                tmp_results["phi_curves"] = tmp_phi_curves
-                tmp_results["results"] = results
-                keep_idx_results = filter_func(tmp_results, filter_params)
-                keep = self.read_filter_results(
-                    keep_idx_results, keep, tmp_psi_curves, tmp_phi_curves, results, mjds, lh_level
-                )
+            batch_size = result_batch.num_results()
+            print("Extracted batch of %i results for total of %i" % (batch_size, total_count))
+            if batch_size > 0:
+                # Do rounds of filtering each is linear and we might be able to get
```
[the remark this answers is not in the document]
jeremykubica: I'm not sure it is a TODO yet. We'd need to benchmark to see if it is worth the extra complexity.

src/kbmod/analysis_utils.py
```diff
@@ -372,325 +358,213 @@ def load_results(
...
-                keep_idx_results.append(self._clipped_sigmaG(z[0], z[1], z[2]))
+            for i, row in enumerate(result_list.results):
+                single_res = self._clipped_sigmaG(row.psi_curve, row.phi_curve, i)
+                row.filter_indices(single_res[1])
 
         end_time = time.time()
         time_elapsed = end_time - start_time
         print("{:.2f}s elapsed".format(time_elapsed))
         print("Completed filtering.", flush=True)
         print("---------------------------------------")
```
DinoBektesevic, line 529: I had to go look up what `zip_curves_list` does, which means it could probably have a better name. I guess it didn't occur to me in the other PR.

If it's not too big of a bother, it only appears, perhaps `zip_phi_psi_idx`? It's clearer and shorter. Bonus points if you return a generator instead of a list (just use `()` instead of `[]` in return) to stay true to name; I'm not sure `starmap_async` would notice the difference and there could be a small perf gain.

This entire phrase also repeats itself verbatim in `apply_clipped_average` so I'd suggest replacing it with a private generic method like:

```python
def __parallel_apply_statistic(self, result_list, statistic):
    start_time = time.time()

    if self.num_cores > 1:
        zipped_curves = result_list.zip_phi_psi_idx()

        keep_idx_results = []
        print("Starting pooling...")
        pool = mp.Pool(processes=self.num_cores)
        keep_idx_results = pool.starmap_async(statistic, zipped_curves)
        pool.close()
        pool.join()
        keep_idx_results = keep_idx_results.get()

        for i, res in enumerate(keep_idx_results):
            result_list.results[i].filter_indices(res[1])
    else:
        for i, row in enumerate(result_list.results):
            single_res = statistic(row.psi_curve, row.phi_curve, i)
            row.filter_indices(single_res[1])

    # TODO: get actual logging 
    end_time = time.time()
    time_elapsed = end_time - start_time
    print("{:.2f}s elapsed".format(time_elapsed))
    print("Completed filtering.", flush=True)
    print("---------------------------------------")

    return result_list
```
and then these parts of `apply_clipped_sigmaG` and `apply_clipped_average` are just:
```python
# TODO: replace with logging
print("Applying Clipped-sigmaG Filtering")
self.lc_filter_type = filter_params["sigmaG_filter_type"]
results_list = self.__parallel_apply_statistic(results_list, self._clipped_sigmaG)

```
and
```python
# TODO: replace with logging
print("Applying Clipped-Average Filtering")
self.lc_filter_type = filter_params["sigmaG_filter_type"]
results_list = self.__parallel_apply_statistic(results_list, self._clipped_average)
```

src/kbmod/analysis_utils.py
```diff
@@ -835,39 +685,30 @@ def _clipped_average(self, psi_curve, phi_curve, index, num_clipped=5, n_sigma=4
...
         """
         print("Applying Kalman Filtering")
         start_time = time.time()
-        # Make copies of the values in 'old_results' and create a new dict
-        psi_curves = np.copy(old_results["psi_curves"])
-        phi_curves = np.copy(old_results["phi_curves"])
 
-        keep_idx_results = kb.kalman_filtered_indices(psi_curves.tolist(), phi_curves.tolist())
+        for row in result_list.results:
+            keep_idx_results = kb.kalman_filtered_indices([row.psi_curve], [row.phi_curve])
```
DinoBektesevic, line 701: Is this correct because it's confusing? Old code had psi and phi as numpy arrays, so a `psi.tolist()` returns the same thing, just as a straight python `list`. `ResultRow.set_psi_phi` sets them as `lists`. So a `[psi/phi]` would be a list of lists in my mind? 
This seemingly works, because nothing breaks when I run it, but its confusing me.

src/kbmod/analysis_utils.py
```diff
@@ -931,105 +758,82 @@ def apply_stamp_filter(
...
                 stamps_slice = search.gpu_coadded_stamps(results_slice, params)
                 for ind, stamp in enumerate(stamps_slice):
                     if stamp.get_width() > 1:
-                        keep["stamps"].append(np.array(stamp))
+                        result_list.results[ind + start_idx].stamp = np.array(stamp)
                         all_valid_inds.append(ind + start_idx)
 
                 # Move to the next chunk.
                 start_idx += chunk_size
+                del stamps_slice
```
DinoBektesevic, line 793: Is this required?

src/kbmod/analysis_utils.py
```diff
@@ -931,105 +758,82 @@ def apply_stamp_filter(
...
+            np.array(result_list.trajectory_list()),
             cluster_params["x_size"],
             cluster_params["y_size"],
             cluster_params["vel_lims"],
             cluster_params["ang_lims"],
             cluster_params["mjd"],
         )
-        rs.filter_results(cluster_idx)
+        result_list.filter_results(cluster_idx)
         del cluster_idx
```
DinoBektesevic, line 828: I'm not sure we need this either, or is there something about Python's GC that makes this so memory intensive this is required?

DinoBektesevic (changes requested):
Is this something I'm doing wrong or can you reproduce? I don't think we should merge until this is cleared up.

```python
(kbmod) [filtered_results_2] $: python deep_analysis.py N5 B1f_20201020
Loading image file data
Loading map of EXPNUM to image index.
Loaded 84 results from /nvme/.../results/data/B1f_20201020/N5/results_B1f_N5_rescls.txt.
Finding object occurences in the data files.
Result 0 mached orbit 760.
Result 82 mached orbit 10536.

RESULTS ------------
Known Fakes: 10
Found Objects: 84
Recovered fakes: 2
Missed known fakes: 8
```

```python
(kbmod) [main] $: python deep_analysis.py N5 B1f_20201020
Loading image file data
Loading map of EXPNUM to image index.
Loaded 78 results from /nvme/.../results/data/B1f_20201020/N5/results_B1f_N5_main.txt.
Finding object occurences in the data files.
Result 0 mached orbit 760.

RESULTS ------------
Known Fakes: 10
Found Objects: 78
Recovered fakes: 1
Missed known fakes: 9
```

src/kbmod/analysis_utils.py
jeremykubica: I think it is correct, but less efficient. The previous code took a list of lists with one list 'row' for each result. In the new code we put the loop outside. Since the code is still expecting a list of lists, we feed them in one at a time. It's not a great solution. Let me know if you see a better way to do this.

src/kbmod/analysis_utils.py
jeremykubica: I don't think so. It was in the original code, so I just moved it into the loop. But I don't think this is actually needed, so let me remove it.

src/kbmod/analysis_utils.py
jeremykubica: Removed.

src/kbmod/analysis_utils.py
```diff
@@ -372,325 +358,213 @@ def load_results(
...
 
+        # Make the stamps.
         if stamp_type == "cpp_median" or stamp_type == "median":
-            trj_res = [kb.trj_result(t, num_images, inds) for t, inds in zip(res_subset, good_subset)]
-            coadd_stamps = [np.array(stamp) for stamp in search.median_sci_stamps(trj_res, radius)]
-        elif stamp_type == "cpp_mean" or stamp_type == "mean":
-            trj_res = [kb.trj_result(t, num_images, inds) for t, inds in zip(res_subset, good_subset)]
-            coadd_stamps = [np.array(stamp) for stamp in search.mean_sci_stamps(trj_res, radius)]
+            coadd_stamps = [np.array(stamp) for stamp in search.median_sci_stamps(trj_list, radius)]
+        elif stamp_type == "cpp_mean":
```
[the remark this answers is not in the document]
jeremykubica: No. That was an accident. Thanks for catching.

src/kbmod/analysis_utils.py
```diff
@@ -372,325 +358,213 @@ def load_results(
...
-                'clipped_sigmaG'
-                'clipped_average'
-                'kalman'
-            chunk_size : int
-                The number of results to load at a given time from search.
-            max_lh : float
-                The maximum likelihood threshold for an acceptable results.
+
+        Arguments:
+            search (`kbmod search`): The search function object.
```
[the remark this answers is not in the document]
jeremykubica: Fixed.

src/kbmod/analysis_utils.py
```diff
@@ -307,28 +295,27 @@ def __init__(self, config):
         self.mask_bits_dict = config["mask_bits_dict"]
         self.flag_keys = config["flag_keys"]
         self.repeated_flag_keys = config["repeated_flag_keys"]
+        self._mjds = mjds
         return
```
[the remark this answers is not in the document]
jeremykubica: You are correct. I will remove the return.

src/kbmod/analysis_utils.py
maxwest-uw: if you wanted to, you could just create a binding for [kalmanFilterIndex](https://github.com/dirac-institute/kbmod/blob/main/src/kbmod/search/Filtering.cpp#L224) and call that directly instead! I also know we're thinking about getting rid of kalman filtering entirely so it might not be worth the effort 😭

src/kbmod/analysis_utils.py
```diff
@@ -931,105 +754,83 @@ def apply_stamp_filter(
...
-
-        keep["final_results"] = np.array(all_valid_inds)
-        keep["stamps"] = np.array(keep["stamps"])
-        print("Keeping %i results" % len(keep["final_results"]), flush=True)
+        if result_list.num_results() <= 0:
+            print("Skipping. Nothing to filter.")
+            return
+        
+        print("Stamp filtering %i results" % result_list.num_results())
+        while start_idx < result_list.num_results():
```
maxwest-uw, line 770: this is so much more readable 👌

src/kbmod/run_search.py
```diff
@@ -329,8 +330,8 @@ def _count_known_matches(self, keep, img_info, search):
 
         # Extract a list of predicted positions for the final results.
         found_objects = []
-        for index in keep["final_results"]:
-            ppos = search.get_mult_traj_pos(keep["results"][index])
+        for row in result_list:
```
maxwest-uw, line 333: the interface pretty much works the same as it did before so my koffi version of `_count_known_matches` won't really be affected 🙂

maxwest-uw (approved):
one possible suggestion on improving memory efficiency of kalman filter changes, but also since we talked about getting rid of the other filtering functions in this week's kbmod meeting anyway I think that change is optional, so I'll go ahead and approve :)

src/kbmod/analysis_utils.py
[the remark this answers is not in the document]
jeremykubica: I changed this to a single function and removed the second linear pass.

src/kbmod/analysis_utils.py
jeremykubica: I changed the name. I'm holding off on combining the two functions because we were discussion removing everything except sigmaG filtering anyway.

src/kbmod/analysis_utils.py
jeremykubica: I think we should prune the filtering options, so let's leave it as-is until we decide which filtering options to keep.

jeremykubica:
I made most of the changes (with the exception of combing the two filtering functions). I also updated the new comments to numpydoc format.

jeremykubica:
I also updated most of the other pythondoc comments in analysis/ to match the style.

DinoBektesevic (approved):

PR #198, opened 2022-12-28
Koffi integration

### LINKED ISSUE #188: Replace known_objects.py with KOFFI
Max West designed KOFFI as a generalized approach to looking up and matching known objects. kbmod should use this code.

maxwest-uw:
- removes the KnownObjects class (and all associated tests) from kbmod and replaces it with the functionality in [koffi](https://github.com/lincc-frameworks/koffi).
- adds a new search parameter `known_obj_obs` (known object observations) that controls the minimum number of observations that koffi needs to associate a result to a known object (default: 3)
- _count_known_matches now outputs a list of all potential matches along with the result_id and the total number of observations matched.
- closes #188


src/kbmod/run_search.py
```diff
@@ -332,40 +334,50 @@ def _count_known_matches(self, result_list, img_info, search):
...
         print("-----------------")
-        known_objects = KnownObjects()
+        matches = {}
+        known_obj_thresh = self.config["known_obj_thresh"]
+        min_obs = self.config["known_obj_obs"]
         if self.config["known_obj_jpl"]:
             print("Quering known objects from JPL")
-            known_objects.jpl_query_known_objects_mult(img_info)
+            matches = koffi.jpl_query_known_objects_stack(
+                ps_list, metadata, min_observations=min_obs, tolerance=known_obj_thresh
```
DinoBektesevic, line 360: This is a personal preference but I always like when keyword name is included and each argument is on its own line.

```
koffi.jpl_query_known_object_stack(
    potential_sources=ps_list,
    image=metadata,
    min_observations=min_obs,
    tolerance=known_obj_tresh
)
```

DinoBektesevic (approved):
This looks great! No changes required, just a personal-preference remark.

src/kbmod/run_search.py
maxwest-uw: seems like a good idea to do actually! should be consistent across all params anyway

PR #201, opened 2023-01-07
CI tests

DinoBektesevic:
Runs CI tests at every pull request to `main` and at every push to `main` or any of the tagged versions. 
Builds and deploys the HTML version of the docs at https://epyc.astro.washington.edu/~kbmod/ every Monday at midnight UTC. 
Not sure if this is the best place for it, but it was the only one available atm. 

Tests are run on baldur. They're a bit slow, but [pass apparently](https://github.com/dirac-institute/kbmod/actions/runs/3859468994).
Test comparing results after processing a pointing group is not a part of this PR. 
The history is a a bit of a mess, I tried my best to clean it up, but I should probably give it another pass and edit some of the commit messages.


.github/workflows/test_build.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+        # Run linters before attempting to compile and install
+        # the package, faster fails.
+      - name: Linter
+        uses: psf/black@stable
+        with:
+          src: "./src"
```
jeremykubica, line 53: Do we want to extend this to the tests/ and src/analysis/ directories?

jeremykubica (approved):
Added on optional question.

maxwest-uw (approved):

.github/workflows/test_build.yaml
DinoBektesevic: Added the directories to the linter check. Ran black on them to get them fixed - I put that in a separate commit if you want to revert the changes, but note that because of the fail-fast strategy no tests will pass until we get them to be compliant.

DinoBektesevic:
I fixed the history, and the review comments, but something is odd with the self-hosted runner. The Actions seem to be stuck waiting to connect, but the Runners tab in the settings tab states the runner is "active" and points to the following test as ["starting up".](https://github.com/dirac-institute/kbmod/actions/runs/3879351446/jobs/6616596734). Running connection tests from baldur pass, so there's something a bit weird happening. 

I can't test this out today unfortunately.

PR #204, opened 2023-01-25
Restructure KBMOD_Demo notebook to use fake data

jeremykubica:
Creates a fake data generator class to use for examples and notebooks. Restructures one of the notebooks (KBMOD_Demo) to use fake data instead of hard coded data paths.

### DIFF
--- a/src/kbmod/fake_data_creator.py
+++ b/src/kbmod/fake_data_creator.py
@@ -11,7 +11,7 @@ from kbmod.search import *
 class FakeDataSet:
     """This class creates fake data sets for testing and demo notebooks."""
 
-    def __init__(self, width, height, num_times, noise_level=2.0, use_seed=False):
+    def __init__(self, width, height, num_times, noise_level=2.0, psf_val=0.5, obs_per_day=3, use_seed=False):
         """The constructor.
 
         Parameters
@@ -21,16 +21,22 @@ class FakeDataSet:
         height : int
             The height of the images in pixels.
         num_times : int
-            The number of times steps (number of images).
+            The number of time steps (number of images).
         noise_level : float
             The level of the background noise.
+        psf_val : float
+            The value of the default PSF.
+        obs_per_day : int
+            The number of observations on the same night.
         use_seed : bool
             Use a deterministic seed to avoid flaky tests.
         """
         self.width = width
         self.height = height
-        self.psf_val = 0.5
+        self.psf_val = psf_val
         self.noise_level = noise_level
+        self.num_times = num_times
+        self.use_seed = use_seed
         self.trajectories = []
 
         # Generate times with multiple observations per night
@@ -43,47 +49,34 @@ class FakeDataSet:
             self.times.append(t)
 
             seen_on_day += 1
-            if seen_on_day == 3:
+            if seen_on_day == obs_per_day:
                 seen_on_day = 0
                 day_num += 1
 
         # Make the image stack.
-        self.stack = self.make_fake_image_stack(use_seed)
+        self.stack = self.make_fake_image_stack()
 
-    def make_fake_image_stack(self, use_seed=False):
+    def make_fake_image_stack(self):
         """Make a stack of fake layered images.
 
         Returns
         -------
         stack : image_stack
         """
-        img_count = len(self.times)
-        t0 = self.times[0]
         p = psf(self.psf_val)
 
         image_list = []
-        for i in range(img_count):
-            if use_seed:
-                img = layered_image(
-                    ("%06i" % i),
-                    self.width,
-                    self.height,
-                    self.noise_level,
-                    self.noise_level**2,
-                    self.times[i],
-                    p,
-                    i,
-                )
-            else:
-                img = layered_image(
-                    ("%06i" % i),
-                    self.width,
-                    self.height,
-                    self.noise_level,
-                    self.noise_level**2,
-                    self.times[i],
-                    p,
-                )
+        for i in range(self.num_times):
+            img = layered_image(
+                ("%06i" % i),
+                self.width,
+                self.height,
+                self.noise_level,
+                self.noise_level**2,
+                self.times[i],
+                p,
+                i if self.use_seed else -1,
+            )
             image_list.append(img)
 
         stack = image_stack(image_list)
@@ -97,10 +90,9 @@ class FakeDataSet:
         trj : trajectory
             The trajectory of the fake object to insert.
         """
-        img_count = len(self.times)
         t0 = self.times[0]
 
-        for i in range(img_count):
+        for i in range(self.num_times):
             dt = self.times[i] - t0
             px = trj.x + dt * trj.x_v + 0.5
             py = trj.y + dt * trj.y_v + 0.5


src/kbmod/fake_data_creator.py
```diff
@@ -0,0 +1,206 @@
...
+        self.psf_val = 0.5
+        self.noise_level = noise_level
+        self.trajectories = []
+
+        # Generate times with multiple observations per night
+        # separated by ~15 minutes.
+        self.times = []
+        seen_on_day = 0
+        day_num = 0
+        for i in range(num_times):
```
drewoldag, line 41: This seems like pretty specific logic, particularly the conditional on line 46. Maybe one more comment about why you're resetting `seen_on_day` every three days would be good.

Also is `3` a magic number? Should it be a "global variable" defined outside of `FakeDataSet`? Similarly, is `57130.2` a magic number?

Finally, is there a reason that you use `num_times` and `use_seed` directly, instead of assigning it to a class variable? It looks like `use_seed` is passed to `make_fake_image_stack`, but you could just reference `self.use_seed` instead.

src/kbmod/fake_data_creator.py
```diff
@@ -0,0 +1,206 @@
...
+        self.stack = self.make_fake_image_stack(use_seed)
+
+    def make_fake_image_stack(self, use_seed=False):
+        """Make a stack of fake layered images.
+
+        Returns
+        -------
+        stack : image_stack
+        """
+        img_count = len(self.times)
```
drewoldag, line 60: It looks like you're indirectly referencing `num_times` here by taking `len(self.times)`?

src/kbmod/fake_data_creator.py
```diff
@@ -0,0 +1,206 @@
...
+        t0 = self.times[0]
+        p = psf(self.psf_val)
+
+        image_list = []
+        for i in range(img_count):
+            if use_seed:
+                img = layered_image(
```
drewoldag, line 67: You could use `functools.partial` here to avoid duplication here. 

```
from functools import partial

layered_image_partial = partial(
      layered_image,
      width = self.width,
      height = self.height,
      noise = self.noise_level, 
      noise_squared = self.noise_level**2,
      psf = p,
      seed = None)

for i in range(img_count):
    this_id = ("%06i", % i)
    if use_seed:
        layered_image_partial(id = this_id, seed = i)
    else:
        layered_image_partial(id = this_id)
...
```

src/kbmod/fake_data_creator.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def insert_object(self, trj):
+        """Insert a fake object given the trajectory.
+
+        Parameters
+        ----------
+        trj : trajectory
+            The trajectory of the fake object to insert.
+        """
+        img_count = len(self.times)
```
drewoldag, line 100: Same comment as above.

src/kbmod/fake_data_creator.py
```diff
@@ -0,0 +1,206 @@
...
+            filename = f"{dir_path}/{img.get_name()}.fits"
+
+            # If the file already exists, delete it.
+            if Path(filename).exists():
+                os.remove(filename)
+
+            # Save the file.
+            img.save_layers(data_dir + "/")
+
+            # Open the file and insert fake WCS data.
```
drewoldag, line 174: Would any of these numbers ever need to change? Alternatively if these number change, will it break unit tests? If so a comment about that would be helpful.

drewoldag (approved):
Left a few comments, but didn't see any blockers.

src/kbmod/fake_data_creator.py
delucchi-cmu: But `num_times` didn't get set as a member variable. You're saying it should?

src/kbmod/fake_data_creator.py
```diff
@@ -0,0 +1,206 @@
...
+    def make_fake_image_stack(self, use_seed=False):
+        """Make a stack of fake layered images.
+
+        Returns
+        -------
+        stack : image_stack
+        """
+        img_count = len(self.times)
+        t0 = self.times[0]
+        p = psf(self.psf_val)
```
delucchi-cmu, line 62: This is just always 0.5. Should it be configurable, too?

src/kbmod/fake_data_creator.py
delucchi-cmu: or

```
... p,
i if use_seed else None,
)
```

I take it functools.partial is the more pythonic way?

src/kbmod/fake_data_creator.py
drewoldag: Yeah, that's my though, since it seems to be implicitly used a couple of times in other methods.

src/kbmod/fake_data_creator.py
drewoldag: Yeah, that's a good point @delucchi-cmu, you're way is clearer. It makes more sense to go with that. If the function was going to be called in many different places with the same arguments, then `partial` is great. But in this case we can call it only once using your approach.

src/kbmod/fake_data_creator.py
```diff
@@ -0,0 +1,206 @@
...
+        """The constructor.
+
+        Parameters
+        ----------
+        width : int
+            The width of the images in pixels.
+        height : int
+            The height of the images in pixels.
+        num_times : int
+            The number of times steps (number of images).
```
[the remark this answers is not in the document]
jeremykubica: Made the change

src/kbmod/fake_data_creator.py
```diff
@@ -0,0 +1,206 @@
...
+        noise_level : float
```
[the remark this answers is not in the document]
jeremykubica: It is the standard deviation of the flux. I think the KBMOD'er should have enough context or not want to set it.

src/kbmod/fake_data_creator.py
jeremykubica: Good calls. I made `use_seed` and `num_times` class variables. I also made the PSF value and the number of times to see an object each day arguments.

57130.2 is an arbitrary date. But KBMOD doesn't use absolute dates (just relative dates from the start of the images), so this shouldn't need to be changed.

src/kbmod/fake_data_creator.py
jeremykubica: Added it as a member variable.

src/kbmod/fake_data_creator.py
jeremykubica: I made it configurable.

src/kbmod/fake_data_creator.py
jeremykubica: Good call. -1 is actually the indicator for no seed, so I will use that.

src/kbmod/fake_data_creator.py
jeremykubica: Changed.

src/kbmod/fake_data_creator.py
jeremykubica: I could see some instances where someone would want to change these, but it would require much of the set to be changed together. Right now the goal is to have some valid values so KBMOD can load the WCS. So I think we are okay using fixed values for now.

delucchi-cmu (approved):

PR #205, opened 2023-01-27
Cleanup more notebooks

jeremykubica:
This PR looks a lot bigger than it really is due to moving and breaking up files. It makes a few changes:
1. Update `kbmod_analysis_demo` to reflect the current project structure and instructions
2. Splits `create_stamps.py` into two files so that we can load data without importing all of tensorflow and the visualization packages. No changes were made to the code in the newly created `visualize_results.py` file. It is an exact copy of the code that was previously in `create_stamps.py`
3. Removes deprecated `np.float` and `np.int` from the remaining functions in `create_stamps.py`.
4. Moves `kbmod_analysis_demo` notebook `examples` directory into the `notebooks` directory.
5. Remove the `test1` folder and its results from the `examples` directory.


delucchi-cmu (approved):

maxwest-uw (approved):
lgtm!

PR #207, opened 2023-02-01
Update README.md

jeremykubica:
Update the example code in the README.md file

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,35 +75,55 @@ A short example injecting a simulated object into a stack of images, and then re
 
 ```python
 
-from kbmodpy import kbmod as kb
+import kbmod.search as kb
 import numpy as np
 
 # Create a point spread function
 psf = kb.psf(1.5)
 
-# load images from list of file paths
-imgs =  [ kb.layered_image(file, psf) for file in example_files ]
+# Create fake data with ten 512x512 pixel images.
+from kbmod.fake_data_creator import *
+ds = FakeDataSet(512, 512, 10)
+imgs = ds.stack.get_images()
+
+# Alternatively, if you have real images you would like to use,
+# load them from files as layered_images:
+# imgs =  [ kb.layered_image(file, psf) for file in example_files ]
+
+# Get the timestamp of the first image.
+t0 = imgs[0].get_time()
+print(f"Image times start at {t0}.")
 
 # Specify an artificial object
-flux = 175.0
-position = (100.7, 150.3)
-velocity = (50, 35)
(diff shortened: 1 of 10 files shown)


delucchi-cmu (approved):
This looks good, but I'd suggest moving more example code into notebooks. Olivia is working on strategies to include notebook testing in continuous integration, so examples wouldn't fall too far behind. This is being implemented for RAIL, but would hopefully be propagated to other projects in time.

jeremykubica:
Thanks Melissa. Most of the examples are in notebooks, which is why we missed this one. We could in theory just remove it altogether.

jeremykubica:
It looks like there was a change to either the linter or the configs, so I needed to also update a lot of formatting throughout the project to get the presubmits to pass. It's all automatically generated changes from black.

README.md
```diff
@@ -75,41 +75,57 @@ A short example injecting a simulated object into a stack of images, and then re
...
+# Recover the object by searching a set of trajectories.
+search = kb.stack_search(stack)
+search.search(
+    5,  # Number of search velocities to try (0, 1, 2, 3, 4)
+    5,  # Number of search angles to try (-0.1, -0.05, 0.0, 0.05, 0.1)
+    -0.1,  # The minimum search angle to test
+    0.1,  # The maximum search angle to test
+    0,  # The minimum search velocity to test
+    4,  # The maximum search velocity to test
+    7,  # The minimum number of observations
```
DinoBektesevic, line 121:
```python
search.search(
    vSteps=5,      # Number of search velocities to try (0, 1, 2, 3, 4)
    aSteps=4,      # Number of search angles to try (-0.1, -0.05, 0.0, 0.05, 0.1)  
    minAngle=-0.1, # The minimum search angle to test
    maxAngle=0.1,  # The maximum search angle to test
    minVelocity=0, # The minimum search velocity to test
    maxVelocity=4, # The maximum search velocity to test
    minObservations=7 # The minimum number of observations
)
```
I will just note that the above also doesn't seem right, the signature of the function looks to be
https://github.com/dirac-institute/kbmod/blob/25456c6e048815fc9f8a56fa60802c33a84233ff/src/kbmod/search/KBMOSearch.h#L44
so the angles and velocity comments above seem flipped.

DinoBektesevic (approved):
Super confusing with the appearance of black differences, but otherwise approved.

README.md
jeremykubica: You are correct. I changed the comments to match.

PR #218, opened 2023-02-02
bump pybind11 to latest version

PWhiddy:
Didn't have a GPU machine handy to test the build with, but this is what updating pybind to the latest commit should look like.


PWhiddy:
Looks like Dino just opened the same

PR #233, opened 2023-02-10
Improve the file loading

jeremykubica:
- Ignore files in the input directory that do not match the given format
- Provide an option for verbose output during file loading
- Improve handling of the id strings in both the timestamp and psf files
- Add a test of load_files


src/kbmod/analysis_utils.py
```diff
@@ -71,36 +74,53 @@ def load_images(
...
+            print(f"Loaded {len(image_time_dict)} time stamps.")
 
         # Load a mapping from visit numbers to PSFs. This dictionary stays
         # empty if no time file is specified.
         image_psf_dict = OrderedDict()
         if psf_file:
             visit_nums, psf_vals = np.genfromtxt(psf_file, unpack=True)
             for visit_num, visit_psf in zip(visit_nums, psf_vals):
-                image_psf_dict[str(int(visit_num))] = visit_psf
+                visit_num_str = f"{int(visit_num):0{id_len}}"
```
delucchi-cmu, line 99: So does this spit out a string (e.g. for the 4th of 120) like "4:0120"? Or is there some extra fanciness happening here? If this is a format that is expected elsewhere, could it be constructed in a (documented) helper method?

src/kbmod/analysis_utils.py
```diff
@@ -71,36 +74,53 @@ def load_images(
...
+                image_psf_dict[visit_num_str] = visit_psf
+        if verbose:
+            print(f"Loaded {len(image_time_dict)} image PSFs stamps.")
```
delucchi-cmu, line 102:
```suggestion
            print(f"Loaded {len(image_psf_dict)} image PSFs stamps.")
```

src/kbmod/analysis_utils.py
jeremykubica: It spits out a zero padded string based on the integer `visit_num` where `id_len` is the length of the string. Since we allow the user to set the starting and ending indices for the id (and thus use variable length ids) we need the extra level of indirection when zero padding it.

delucchi-cmu: That *is* fancy!

delucchi-cmu (approved):

PR #234, opened 2023-02-10
Create a notebook of visualization examples

jeremykubica:
Also includes a set of fake data with smaller images.


cchris28 (approved):
I was able to run this under wsl2 so I am happy.

PR #235, opened 2023-02-10
Create a simple shift and stack algorithm for illustration purposes

jeremykubica:


cchris28 (approved):
Looks fine. Just in case you are not aware, namespaces are open in c++ and so using a common name like "search" increases the chances that some other code will also define something in that namespace. This is fine if that is your intent. But it is different from how modules work in python. If you are using namespaces to hide code then consider either static or the anonymous namespace.

PR #244, opened 2023-02-21
Rework CI tests, fix Bendis issues.

DinoBektesevic:


.github/workflows/canary_builds.yaml
```diff
@@ -45,12 +44,6 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-        # Run linters before attempting to compile and install
-        # the package, faster fails.
-      - name: Linter
-        uses: psf/black@stable
-        with:
-          jupyter: true
```
jeremykubica, line 53: Why not include the linter tests in the canary build?

DinoBektesevic: It's not in that workflow because when it is it reports no `python3` found so I'm trying to figure out why on `baldur` it sees `/epyc/anaconda2019/bin/python3` but not on `bendis`. This PR is not review ready yet.

jeremykubica (approved):

PR #245, opened 2023-02-21
Allow trajectories to start off chip

jeremykubica:
This PR adds the ability for the search to start off the chip using either explicit bounds on the x and y pixels (`x_pixel_bounds` and `y_pixel_bounds` respectively) or a constance number of pixel buffer around the edge (`x_pixel_buffer` and `y_pixel_buffer`). The trajectories returned look just like the normal results but have x and y values outside the image’s dimensions (including negative values).

Also:
- Updates the unit tests to include searches with different bounds and a trajectory starting off chip
- Updates the regression test to include two trajectories starting off the chip.
- Updates the documentation
- Modifies the trajectories in `test_search_encode.py` to provide more variety in the test cases (instead of using the same fake trajectory as `test_search.py`).


cchris28 (approved):

src/kbmod/search/kernels.cu
```diff
@@ -86,9 +86,19 @@ __device__ float readEncodedPixel(void *imageVect, int index, int numBytes, cons
...
+    const int search_width = params.x_start_max - params.x_start_min;
+    const int search_height = params.y_start_max - params.y_start_min;
+    if ((x_i >= search_width) || (y_i >= search_height)) {
+        return;
+    }
+
     // Get origin pixel for the trajectories.
-    const unsigned short x = blockIdx.x * THREAD_DIM_X + threadIdx.x;
-    const unsigned short y = blockIdx.y * THREAD_DIM_Y + threadIdx.y;
+    const int x = x_i + params.x_start_min;
```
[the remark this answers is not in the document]
jeremykubica: It does not appear that using the extra space matters too much in the diff test. The additional complexity of the offset notation is that `trajectory` is used throughout the code up to data export. At some point we will need to convert to absolute (from offset), which would mean creating a separate data structure for the GPU results. For now, I suggest we keep as is.

docs/source/user_manual/search_params.rst
```diff
@@ -173,6 +173,26 @@ Search parameters are set extensively via the :py:attr:`~kbmod.run_search.run_se
...
+|                        |                             | the image bounds to use for starting   |
+|                        |                             | coordinates. If `x_bounds` is provided |
+|                        |                             | that takes precedence.                 |
+|                        |                             | `None` uses the image bounds.          |
++------------------------+-----------------------------+----------------------------------------+
+| y_pixel_bounds         | None                        | A length two list giving the starting  |
+|                        |                             | and ending y pixels to use for the     |
+|                        |                             | search. `None` uses the image bounds.  |
++------------------------+-----------------------------+----------------------------------------+
+| y_pixel_buffer         | None                        | An integer length of pixels outside    |
```
DinoBektesevic, line 190: A thing I recently learned:

* double backtics render highlighted text, f.e. ` ``key`` ` --> `key`
* single backtics render italicized text, f.e, `key` --> _key_
* it's probably a good idea to render key names in highlight - f.e. ` ``x_pixel_bound`` ` --> `x_pixel_bound`
* it's probably a good idea to render key values in highlight with quotations - f.e. ` ``"a_string_value"`` ` --> `"a_string_value"` 

to produce the most consistent documentation.

docs/source/user_manual/search_space.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Given the linear sampling for both velocities and angles, the full set of candidate trajectories is computed as::
+
+
+    for (int a = 0; a < angleSteps; ++a) {
+        for (int v = 0; v < velocitySteps; ++v) {
+            searchList[a * velocitySteps + v].xVel = cos(angles[a]) * velocities[v];
+            searchList[a * velocitySteps + v].yVel = sin(angles[a]) * velocities[v];
+        }
+    }
```
DinoBektesevic, line 53: Wrapping this in 

```
.. code:: cpp

    for (int a=0....)
    
```

will result in correct syntax highlighting. Same for the other code blocks and languages like `python`

src/kbmod/search/kernels.cu
```diff
@@ -86,9 +86,21 @@ __device__ float readEncodedPixel(void *imageVect, int index, int numBytes, cons
...
-    // Get origin pixel for the trajectories.
-    const unsigned short x = blockIdx.x * THREAD_DIM_X + threadIdx.x;
-    const unsigned short y = blockIdx.y * THREAD_DIM_Y + threadIdx.y;
+    // Get the x and y coordinates within the search space.
+    const int x_i = blockIdx.x * THREAD_DIM_X + threadIdx.x;
+    const int y_i = blockIdx.y * THREAD_DIM_Y + threadIdx.y;
+
+    // Check that the x and y coordinates are consistent with the search space.
+    const int search_width = params.x_start_max - params.x_start_min;
+    const int search_height = params.y_start_max - params.y_start_min;
```
DinoBektesevic, line 95: Are the sizes of search area something we might want at a higher level of the code too? Should we move this above?

DinoBektesevic (approved):
Looks good, nothing but minor documentation comments.

docs/source/user_manual/search_space.rst
jeremykubica: I can't get this formatting to work out. It looks good with the current formatting, so I will leave for now and investigate more later.

src/kbmod/search/kernels.cu
jeremykubica: The sizes are used in other parts of the code. Since it is just a pair of subtractions, I don't think it is a huge overhead to recompute them from the params in a few different functions.

PR #255, opened 2023-03-06
Get rid of hardcoded stamp size

jeremykubica:
The stamp size output was hard coded for the default stamp setting, but would cause a problem if another size is used.

### DIFF
--- a/src/kbmod/result_list.py
+++ b/src/kbmod/result_list.py
@@ -428,7 +428,7 @@ class ResultList:
         if np.any(stamps_list == None):
             stamps_list = np.array([])
 
-        stamp_size = 0
+        stamp_size = 441
         if len(stamps_list) > 0:
             stamp_size = stamps_list[0].size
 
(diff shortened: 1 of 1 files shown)


src/kbmod/result_list.py
```diff
@@ -423,12 +423,18 @@ def save_to_files(self, res_filepath, out_suffix):
             fmt="%.4f",
         )
 
+        # Output the co-added stamps.
         stamps_list = np.array([x.stamp for x in self.results])
-        if np.any(stamps_list is None):
+        if np.any(stamps_list == None):
             stamps_list = np.array([])
+
+        stamp_size = 0
```
delucchi-cmu, line 431: Maybe default to 441 instead of 0?

delucchi-cmu (approved):

src/kbmod/result_list.py
jeremykubica: Done

PR #258, opened 2023-03-07
Fix a bug when FileUtils tries to load a single trajectory.

jeremykubica:
- Fix a bug where FileUtils crashes when it tries to load a single trajectory because it received a 1-d array instead of a 2-d array.
- Add a new test to cover this case
- Add a result trajectory writer  `save_results_file` to simplify the test.
- Use `save_results_file` in `ResultList`'s `save_to_files` function for consistency.


src/kbmod/file_utils.py
```diff
@@ -183,6 +196,12 @@ def load_results_file(filename):
             usecols=(1, 3, 5, 7, 9, 11, 13),
             names=["lh", "flux", "x", "y", "vx", "vy", "num_obs"],
         )
+
+        # If there is only one line, numpy will return a 1-d array.
+        # Make it 2-d for consistency.
+        if len(results.shape) == 0:
```
delucchi-cmu, line 208: Could you use the `ndmin` parameter when calling `genfromtxt`? That can force the `results` object to have at least 2 dimensions.

jeremykubica: Good call. That makes the function a lot cleaner.

delucchi-cmu (approved):

PR #259, opened 2023-03-07
Create new stamp plotting class

jeremykubica:
This is the first PR in breaking up the functionality in `create_stamps.py`. This PR creates a new `ResultsVisualizer` class and updates `kbmod_analysis_demo.ipynb` to use that instead of `create_stamps.py`.

We also move `create_stamps.py` back into the analysis directory to keep the top level directory cleaner.


maxwest-uw (approved):
looks good, the only thing that seems kind of weird to me is that you have `imshow` being called in `plot_single_stamp`, even when it's being called multiple times like in `plot_stamps` (which in my mind should create multiple different plot images) but mpl has a lot of magic going on under the hood and it seems like it works fine in the notebook so 🤷‍♀️

Issue #254, opened 2023-03-04
Remove visit_in_filename


jeremykubica:
Remove the manual visit_in_filename flags and replace with more general parsing, such as extracting a sequence of 4+ numbers directly from the filename.

PR #260, opened 2023-03-08
Improve extraction of Visit ID

jeremykubica:
Change the logic of how the visit ID is extracted to first look for it in the `IDNUM` field and second to extract the first sequence of 5+ digits in the filename. This allows us to remove the `visit_in_filename` parameter.

Also change how the file loading function is structured to simplify the handling of timestamps. We load the header first so we can use that to compute both visit ID and timestamp.


jeremykubica:
Closes #254

PR #263, opened 2023-03-14
Give ResultList the ability to track filtered results

jeremykubica:
Allow ResultList to optionally track results that is has filtered out with a string label. This allows efficiency analyses of different filtering steps by looking at which trajectories are removed at which step. Tracking filtered results should not be used in most production runs as it will increase overall memory usage.

This is the first PR in refactoring the filtering system to make it more modular.

### DIFF
--- a/src/kbmod/result_list.py
+++ b/src/kbmod/result_list.py
@@ -211,9 +211,23 @@ class ResultRow:
 class ResultList:
     """This class stores a collection of related data from all of the kbmod results."""
 
-    def __init__(self, all_times):
+    def __init__(self, all_times, track_filtered=False):
+        """Create a ResultList class.
+
+        Parameters
+        ----------
+        all_times : list
+            A list of all time stamps.
+        track_filtered : bool
+            Whether to track (save) the filtered trajectories. This will use
+            more memory and is recommended only for analysis.
+        """
         self.all_times = all_times
         self.results = []
+
+        # Set up information to track which row is filtered at which round.
+        self.track_filtered = track_filtered
+        self.filtered = {}
 
     def num_results(self):
         """Return the number of results in the list.
@@ -228,6 +242,7 @@ class ResultList:
     def clear(self):
         """Clear the list of results."""
         self.results.clear()
+        self.filtered.clear()
 
     def append_result(self, res):
         """Add a single ResultRow to the result set.
@@ -248,6 +263,14 @@ class ResultList:
             The data structure containing additional `ResultRow`s to add.
         """
         self.results.extend(result_list.results)
+
+        # When merging the filtered results extend lists with the
+        # same key and create new lists for new keys.
+        for key in result_list.filtered.keys():
+            if key in self.filtered:
+                self.filtered[key].extend(result_list.filtered[key])
+            else:
+                self.filtered[key] = result_list.filtered[key]
 
     def trajectory_list(self, indices_to_use=None):
         """Create and return a list of just the trajectories.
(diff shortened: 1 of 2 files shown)


src/kbmod/result_list.py
```diff
@@ -359,16 +382,50 @@ def append_result_dict(self, res_dict):
...
+        if not self.track_filtered:
+            self.results = [self.results[i] for i in indices_to_keep]
+        else:
+            keep_set = set(indices_to_keep)
+
+            # Divide the current results into a set of rows to keep
+            # and a set to filter.
+            keep_rows = []
+            filter_rows = []
+            for i, row in enumerate(self.results):
```
delucchi-cmu, line 411: I don't know what `len(self.results)` typically looks like, but python `list.append` can get pretty slow for a lot of elements. I guess this path isn't running in high-performance runs, but some kinda list comprehension could speed up this stage.

src/kbmod/result_list.py
```diff
@@ -382,13 +439,41 @@ def filter_on_stats(self, lh_threshold=-1.0, min_valid_indices=-1):
...
-        for x in self.results:
+        indices_to_keep = []
+        for i, x in enumerate(self.results):
             keep = min_valid_indices == -1 or len(x.valid_indices) >= min_valid_indices
             keep = keep and (lh_threshold < 0.0 or x.final_likelihood >= lh_threshold)
             if keep:
-                tmp_results.append(x)
-        self.results = tmp_results
+                indices_to_keep.append(i)
+        self.filter_results(indices_to_keep, "filter_on_stats")
```
delucchi-cmu, line 448: Could the user provide their own filter label?

If, say, they do something like `resultList.filter_on_stats(easy_stats).filter_on_more(more_stats).filter_on_stats(expensive_stats)`, it would be tough to distinguish what gets filtered because of the first call to `filter_on_stats` vs the second call.

Maybe pre-pend some filter counter to the label? Or maybe no one cares?

src/kbmod/file_utils.py
```diff
@@ -22,6 +22,32 @@ class FileUtils:
...
+        ----------
+        s : string
+            The input string
+
+        Returns
+        -------
+        res : string
+            The output string
+        """
+        separators = set([" ", ".", ",", ";", "\t", "\n", "_"])
```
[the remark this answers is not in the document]
jeremykubica: Yes. It is technically a no-op. But to make it cleaner I changed the logic.

I added the other separators.

src/kbmod/result_list.py
```diff
@@ -359,16 +382,50 @@ def append_result_dict(self, res_dict):
...
+        Returns
+        -------
+        self : ResultList
+            Returns a reference to itself to allow chaining.
         """
-        self.results = [self.results[i] for i in indices_to_keep]
+        if not self.track_filtered:
+            self.results = [self.results[i] for i in indices_to_keep]
+        else:
+            keep_set = set(indices_to_keep)
```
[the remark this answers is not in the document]
jeremykubica: Good call. I added some logic to de-duplicate and added a corresponding test.

src/kbmod/result_list.py
jeremykubica: Changed

src/kbmod/result_list.py
jeremykubica: You are ahead of the game. In the next PR the filter class will dynamically generate the labels based on both the filter and the parameters.

PR #264, opened 2023-03-15
Allow trajectory to be pickled

jeremykubica:
Since trajectory is based on a C++ struct, it does not automatically work with pickle. This augments the pybind setting to allow that.

This will enable future parallelizations.

### DIFF
--- a/src/kbmod/search/bindings.cpp
+++ b/src/kbmod/search/bindings.cpp
@@ -255,7 +255,23 @@ PYBIND11_MODULE(search, m) {
                 return "lh: " + to_string(t.lh) + " flux: " + to_string(t.flux) + " x: " + to_string(t.x) +
                        " y: " + to_string(t.y) + " x_v: " + to_string(t.xVel) + " y_v: " + to_string(t.yVel) +
                        " obs_count: " + to_string(t.obsCount);
-            });
+            })
+            .def(py::pickle(
+                [](const tj &p) { // __getstate__
+                    return py::make_tuple(p.xVel, p.yVel, p.lh, p.flux, p.x, p.y, p.obsCount);
+                },
+                [](py::tuple t) { // __setstate__
+                    if (t.size() != 7)
+                        throw std::runtime_error("Invalid state!");
+                    tj trj = {t[0].cast<float>(),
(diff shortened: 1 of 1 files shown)


src/kbmod/search/bindings.cpp
```diff
@@ -255,7 +255,23 @@ PYBIND11_MODULE(search, m) {
                 return "lh: " + to_string(t.lh) + " flux: " + to_string(t.flux) + " x: " + to_string(t.x) +
                        " y: " + to_string(t.y) + " x_v: " + to_string(t.xVel) + " y_v: " + to_string(t.yVel) +
                        " obs_count: " + to_string(t.obsCount);
-            });
+            })
+            .def(py::pickle(
```
DinoBektesevic, line 259: Add documentation. See [how](https://github.com/dirac-institute/kbmod/blob/main/src/kbmod/search/bindings.cpp#L39). See [rendered](https://epyc.astro.washington.edu/~kbmod/api_reference/search_referenceapi.html#kbmod.search.psf).


I am kind of interested in seeing what you think of this file, in the eventuality a full fledged python docs need to be added for all methods classes and constants with examples. There are no good natural ways of breaking out the bindings in separate files.

One is naming everything in C++ with underscore and having explicitly inherited, basically no-op classes, but then the docs can be broken out into them. 

The other is to do what Rubin does - but that requires cpp-utils module they use and various pretty heavy mechanisms most of which we won't use. 

I think the simplest one would be to breakout the search namespace into multiple different ones - each can have its own bindings file then, and make sure we keep these neatly formatted. 

Or keep it all in one file and keep that neatly formatted.

DinoBektesevic (approved):

src/kbmod/search/bindings.cpp
jeremykubica: I added a docstring for the trajectory overall. I also tried for the pickle function, but that wouldn't render. I also don't know if that will be useful in most cases as it just enables expected behavior.

In terms of overall documentation, I would like to redo the C++ documentation from scratch. I had been thinking of the inherited, no-op classes. But I would be happy to discuss more.

PR #265, opened 2023-03-15
Initial Filter implementation

jeremykubica:
Create a new `Filter` class which can be subclassed to perform filtering over the results. This PR adds:
- Adds a `Filter` base class
- Adds two filter implementations `LHFilter` and `NumObsFilter` along with corresponding tests
- Adds an `apply_filter` method to the results with the option for parallelization.
- Replaces the current likelihood and number-of-observations filtering in `analysis_utils` with the new filters.


src/kbmod/filters.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class NumObsFilter(Filter):
+    """A filter for result's number of valid observations."""
+
+    def __init__(self, min_obs):
+        """Create a ResultsNumObsFilter.
+
+        Parameters
+        ----------
```
DinoBektesevic, line 88: I think this is fine by numpydoc, I think it allows documenting it in class or init, but it has to be formatted like this (`Parameters` undelines, type, newline, four spaces, text). Just mentioning this in case you feel like you want to move it to class docstring - I believe it's optional.

src/kbmod/result_list.py
```diff
@@ -427,25 +428,34 @@ def filter_results(self, indices_to_keep, label=None):
...
 
         Parameters
         ----------
-        threshold : float
-            The minimum likelihood a row needs to pass the filtering.
-            Use -1 to ignore this field.
-        min_valid_indices : int
-            The minimum number of valid indices a row needs to pass the filtering.
-            Use -1 to ignore this field.
+        filter_obj : ResultListFilter
```
DinoBektesevic, line 438: Just `Filter` as type?

DinoBektesevic (changes requested):
Make `Filters` an abstract base class and merge approved. The rest are comments on documentation.

src/kbmod/filters.py
```diff
@@ -0,0 +1,118 @@
...
+        row : ResultRow
+            The row to evaluate.
+
+        Returns
+        -------
+        bool
+           An indicator of whether to keep the row.
+        """
+        return True
+
```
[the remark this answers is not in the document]
jeremykubica: Sounds good. I made Filter an abstract base class.

I would like to keep the `get_filter_name` function because I think it is easier for non-python experts to understand what is going on and how it is used.

src/kbmod/filters.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(self, **kwargs):
+        """Create a ResultsLHFilter.
+
+        Takes the following optional parameters:
+        min_lh : float - The minimum likelihood.
+        max_lh : float - The maximum likelihood.
+        """
+        self.min_lh = kwargs.get("min_lh", None)
+        self.max_lh = kwargs.get("max_lh", None)
```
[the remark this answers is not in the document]
jeremykubica: Changed the signature and the doc string.

I am keeping the option for either boundary to be `None` in order to allow one sided filters.  I know we could use something like -inf and inf, but I think this provides a reasonable interface.,

src/kbmod/filters.py
```diff
@@ -0,0 +1,118 @@
...
+        Returns
+        -------
+        bool
+           An indicator of whether to keep the row.
+        """
+        lh = row.final_likelihood
+        if self.min_lh is not None and lh < self.min_lh:
+            return False
+        if self.max_lh is not None and lh > self.max_lh:
+            return False
```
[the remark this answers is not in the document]
jeremykubica: As noted above, keeping the option for the boundaries to be `None` in order to support one sided filters.

src/kbmod/result_list.py
jeremykubica: Fixed.

src/kbmod/filters.py
```diff
@@ -0,0 +1,128 @@
+import abc
+
+from kbmod.result_list import *
+
+
+class Filter(abc.ABC):
```
delucchi-cmu, line 6: From a code organization angle, I had thought these might live in a subdirectory. As you start to build up filters, this file will get annoyingly big. e.g.

```
src/kbmod/filters/
  __init__.py
  base_filter.py
  lh_filter.py
  num_obs_filter.py
```
And a similar structure to the tests.

src/kbmod/filters.py
```diff
@@ -0,0 +1,128 @@
...
+        """Create a ResultsLHFilter.
+
+        Parameters
+        ----------
+        min_lh : ``float``
+            Minimal allowed likelihood. Use None for no min filter.
+        max_lh : ``float``
+            Maximal allowed likelihood. Use None for no max filter.
+
+        Examples
```
delucchi-cmu, line 34: +1. This is a great use of an `Examples` docstring.

src/kbmod/filters.py
```diff
@@ -0,0 +1,128 @@
...
+        """Create a ResultsNumObsFilter.
+
+        Parameters
+        ----------
+        min_obs : ``int``
+            The minimum number of valid observations needed.
+
+        Examples
+        --------
+        # r is a ResultRow with 5 valid indices.
```
delucchi-cmu, line 94: Could as-yet-unimplemented filters chop down the number of valid indices in a result list? Is it worth noting that order of filtering may matter for this one?

src/kbmod/filters.py
jeremykubica: I think we might want to do that eventually. But let's hold off until we see how many filters we have.

src/kbmod/filters.py
jeremykubica: My current plan is to have Filters not modify any of the rows. So it wouldn't change anything. Instead I want to create a second class (something like PostProcessor) that runs over each row and modifies it. There will be order dependent.

src/kbmod/filters.py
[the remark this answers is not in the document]
jeremykubica: Sounds good. I extended the doc string.

PR #269, opened 2023-03-16
Remove region search

jeremykubica:
Removes the experimental region search KBMOD algorithm, which we are not using including:
- The `KBMORegionSearch` class
- The `trajRegion` struct
- The `PooledImage` class
- The `extremeInRegion` function from `RawImage`.
- The `trajRegion` helper functions from `TrajectoryUtils` (`convertTrajRegion`, `subdivideTrajRegion`, and `filterTrajRegionsLH`)
- All the corresponding tests.

This PR leaves the RawImage pooling functions, which we might want to clean up in a later change if they are not being used.


delucchi-cmu (approved):
I love the smell of code deletion in the morning.

PR #272, opened 2023-03-16
removed docs/source/user_manual/#overview.rst#

cchris28:


DinoBektesevic:
When you're deleting this can you also remove the `docs/source/examples/_notebooks`? These are coped at documentation build time (see [here](https://github.com/dirac-institute/kbmod/blob/main/docs/source/conf.py#L65)) so we don't actually have to maintain duplicates. 

Could you also please add `_notebooks`  and `#*` to `.gitignore` too. I seem to have added the [build directories](https://github.com/dirac-institute/kbmod/blob/main/.gitignore#L15) and Emacs tilde `*~` files, but forgot to add these.

Realistically the notebooks and all the test data should live in a separate repo. The library itself should be the absolute bare minimum size required for it to work as expected, but we'll get there.

jeremykubica (approved):

PR #274, opened 2023-03-16
Clean up accidentally commited files.

DinoBektesevic:
I felt like deleting stuff too.


cchris28 (approved):
Yes! I was about to do this myself.
I have myriad other doc changes so I'll make a followup issue

PR #279, opened 2023-03-17
Remove kbmod_info

jeremykubica:
This PR moves the remaining functionality of `kbmod_info.py` into other files:
- The file functions, including MPC reading and writing, are moved into `file_utils.py`
- Tests are added for all the MPC functions
- The `get_searched_radec` is already implemented as the `trajectory_to_skycoords` function in `image_info.py`


DinoBektesevic (approved):
Always happy to remove confusingly named classes. 

I think however, the `FileUtils` is becoming a bit of a god-class itself. It made sense when you were originally refactoring out the old code. What you were refactoring was a class that had some class-methods related to some file IO stuff. I think if we have much more of such functionality in KBMOD left to refactor out a better approach is to leave them in `utils.py` as functions. 

```python
import kbmodutils
kbmodutils.format_result_mps
```

for example. Some of this functionality (like reading or writing MPC/IPAC/IRSA/WHATEVER formats are really best punted to the [AstroPy's IO module](https://docs.astropy.org/en/stable/io/unified.html#built-in-table-readers-writers) or `Table` object. This is a no-op now, something to think about (perhaps) if the class keeps growing in the future.

jeremykubica:
Good call. I will merge this in for now, but I will look at breaking up FileUtils in the future.

PR #281, opened 2023-03-20
Move the filters into their own directory and break into multiple files

jeremykubica:
- Break the filters into multiple files in the new `filters/` subdirectory
- Add an initial stamp filter and related tests
- Update the documentation

### DIFF
--- a/src/kbmod/__init__.py
+++ b/src/kbmod/__init__.py
@@ -12,6 +12,8 @@ from . import analysis_utils
 from . import jointfit_functions
 from . import result_list
 from . import file_utils
+
+# import the filters subdirectory
 from . import filters
 
 # lazy import analysis to arrest
--- a/src/kbmod/analysis_utils.py
+++ b/src/kbmod/analysis_utils.py
@@ -14,8 +14,8 @@ from scipy.special import erfinv  # import mpmath
 from sklearn.cluster import DBSCAN, OPTICS
 
 from .file_utils import *
+from .filters.stats_filters import *
 from .image_info import *
-from .filters import *
 from .result_list import *
 import kbmod.search as kb
 
--- a/src/kbmod/filters/base_filter.py
+++ b/src/kbmod/filters/base_filter.py
@@ -1,0 +1,18 @@
+import abc
+
+from kbmod.result_list import *
+
+
+class Filter(abc.ABC):
+    """The base class for derived filters on the ResultList."""
+
+    def __init__(self, *args, **kwargs):
+        pass
+
+    @abc.abstractmethod
+    def get_filter_name(self):
+        pass
+
+    @abc.abstractmethod
+    def keep_row(self, row: ResultRow):
+        pass
--- a/src/kbmod/filters/base_filter.py
+++ b/src/kbmod/filters/base_filter.py
@@ -11,8 +11,28 @@ class Filter(abc.ABC):
 
     @abc.abstractmethod
     def get_filter_name(self):
+        """Get the name of the filter.
+
+        Returns
+        -------
+        str
+            The filter name.
+        """
         pass
 
     @abc.abstractmethod
     def keep_row(self, row: ResultRow):
+        """Determine whether to keep an individual row based on
+        the row's information and the parameters of the filter.
+
+        Parameters
+        ----------
+        row : ResultRow
+            The row to evaluate.
+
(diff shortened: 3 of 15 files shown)


src/kbmod/filters/stamp_filters.py
```diff
@@ -0,0 +1,68 @@
...
+from kbmod.filters.base_filter import *
+from kbmod.search import *
+from kbmod.result_list import *
+
+
+class StampPeakFilter(Filter):
+    """A filter on how far the stamp's peak is from the center."""
+
+    def __init__(self, stamp_radius, x_thresh, y_thresh, *args, **kwargs):
+        """Create a ResultsNumObsFilter.
```
delucchi-cmu, line 12:
```suggestion
        """Create a StampPeakFilter.
```

src/kbmod/filters/stamp_filters.py
```diff
@@ -0,0 +1,68 @@
+import abc
```
delucchi-cmu, line 1: This file has three different wordings for the filter:

- `stamp_filter` (the file name)
- `StampPeakFilter` (class name)
- `PeakOffset` (filter name)

Can we converge on something?

src/kbmod/__init__.py
```diff
@@ -12,6 +12,8 @@
 from . import jointfit_functions
 from . import result_list
 from . import file_utils
+
+# import the filters subdirectory
 from . import filters
```
delucchi-cmu, line 17: I think you'll want to remove `src/kbmod/filters.py`

jeremykubica: It has been removed, but github is showing it as a file move into filters/stats_filter.py.

src/kbmod/filters/stamp_filters.py
jeremykubica: The file name is meant to be more general than the class name since I am planning on having a few basic stamp filters in this file.

I changed the filter name to match the class name though.

src/kbmod/analysis_utils.py
```diff
@@ -14,8 +14,8 @@
 from sklearn.cluster import DBSCAN, OPTICS
 
 from .file_utils import *
+from .filters.stats_filters import *
```
DinoBektesevic, line 17: I think that it's permissible to import `*` in ``__init__.py`` files because it's a significant hassle otherwise, and I do understand we import `*` everywhere in our code, but I really don't think we should continue doing that forward. 

It's ok not to change this right, I think that at some point I will go and redo all of the files at the same time.

src/kbmod/filters/base_filter.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self, *args, **kwargs):
+        pass
+
+    @abc.abstractmethod
+    def get_filter_name(self):
+        pass
+
+    @abc.abstractmethod
+    def keep_row(self, row: ResultRow):
+        pass
```
DinoBektesevic, line 18: These should be documented and targeted to the developer - they need to know what the expected returns are to successfully overwrite the `Filter`. If there are any particular errors in filtering we are willing to catch and ignore (or log) that would make sense to document too. 

We should mention if they are or are not allowed to raise errors and need to handle their own errors as fail-or-pass of the filter. For example `FilterLogLikelihood`, and somehow the likelihood is negative, would raise an `RuntimeError`. This is obviously easily marked as a fail and indicates that something's going wrong with likelihood calculations. While we don't have any logs at the moment, later we need to extend the class to pass it a one - so that the user can at least log the error somewhere. This would be a good place to add that note in the future.

src/kbmod/search/bindings.cpp
```diff
@@ -86,6 +86,7 @@ PYBIND11_MODULE(search, m) {
             .def("get_height", &ri::getHeight)
             .def("get_width", &ri::getWidth)
             .def("get_ppi", &ri::getPPI)
+            .def("get_all_pixels", &ri::getPixels)
```
DinoBektesevic, line 89: Documentation of any new C++ function is not a lot given the context, but it's better to do than not to do, even (or rather especially) when they're one-liners.

DinoBektesevic (approved):
This all looks good to me, two missing docs fixes (`Filters` and `get_all_pixels`) and some worth to think about in the future notes.

PR #280, opened 2023-03-18
Sphinx doc tweaks

cchris28:
I noticed these while building and examining all the notebooks and documentation.
The 1st commit is for warnings I saw when building sphinx.
The 2nd is for errors I got while building the notebooks using jupyter nbconvert.

PR #282, opened 2023-03-21
Docs/sphinx tweaks

cchris28:
Same as pull request https://github.com/dirac-institute/kbmod/pull/280 but with branch name and commit message as Dino requested. Should test documentation better.


DinoBektesevic (approved):
Thanks for the effort, this looks good. Don't sweat the docs on run_search - that documentation is super awkward to do.

PR #283, opened 2023-03-22
Add more stamp filters

jeremykubica:
This PR adds the existing stamp filters as `Filter` subclasses. All of these filters already exist in `image_kernels.cu` and reuse that code in their core. Adding these filters as `Filter` subclasses will allow users more flexibility to run the stamp filtering in different order or multiple times with different parameters.

### DIFF
--- a/src/kbmod/filters/stamp_filters.py
+++ b/src/kbmod/filters/stamp_filters.py
@@ -223,7 +223,7 @@ class StampCenterFilter(Filter):
         # Check the stamp's width is correct.
         width = 2 * self.stamp_radius + 1
         if len(row.stamp) != width * width:
-            raise ValueError("Expected stamp of size {width} by {width}.")
+            raise ValueError(f"Expected stamp of size {width} by {width}.")
 
         # Find the value of the center pixel. Filter pixels with no data.
         center_index = width * self.stamp_radius + self.stamp_radius
--- a/src/kbmod/filters/stamp_filters.py
+++ b/src/kbmod/filters/stamp_filters.py
@@ -27,7 +27,7 @@ class StampPeakFilter(Filter):
         super().__init__(*args, **kwargs)
 
         if stamp_radius <= 0:
-            raise ValueError("Invalid stamp radius {stamp_radius}.")
+            raise ValueError(f"Invalid stamp radius {stamp_radius}.")
         self.stamp_radius = stamp_radius
 
         self.x_thresh = x_thresh
--- a/src/kbmod/filters/stamp_filters.py
+++ b/src/kbmod/filters/stamp_filters.py
@@ -1,4 +1,8 @@
-import abc
+"""A series of Filter subclasses for processing basic stamp information.
+
+The filters in this file all operate over simple statistics based on the
+stamp pixels.
+"""
 
 from kbmod.filters.base_filter import *
 from kbmod.search import *
@@ -22,7 +26,10 @@ class StampPeakFilter(Filter):
         """
         super().__init__(*args, **kwargs)
 
+        if stamp_radius <= 0:
+            raise ValueError("Invalid stamp radius {stamp_radius}.")
         self.stamp_radius = stamp_radius
+
         self.x_thresh = x_thresh
         self.y_thresh = y_thresh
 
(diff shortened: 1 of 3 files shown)


src/kbmod/filters/stamp_filters.py
```diff
@@ -22,7 +26,10 @@ def __init__(self, stamp_radius, x_thresh, y_thresh, *args, **kwargs):
         """
         super().__init__(*args, **kwargs)
 
+        if stamp_radius <= 0:
+            raise ValueError("Invalid stamp radius {stamp_radius}.")
```
delucchi-cmu, line 30:
```suggestion
            raise ValueError(f"Invalid stamp radius {stamp_radius}.")
```

src/kbmod/filters/stamp_filters.py
```diff
@@ -66,3 +73,174 @@ def keep_row(self, row: ResultRow):
...
+            The threshold for the j=1, k=1 moment.
+        m02_thresh : ``float``
+            The threshold for the j=0, k=2 moment.
+        m20_thresh : ``float``
+            The threshold for the j=2, k=0 moment.
+        """
+        super().__init__(*args, **kwargs)
+
+        if stamp_radius <= 0:
+            raise ValueError("Invalid stamp radius {stamp_radius}.")
```
delucchi-cmu, line 111:
```suggestion
            raise ValueError(f"Invalid stamp radius {stamp_radius}.")
```

src/kbmod/filters/stamp_filters.py
```diff
@@ -66,3 +73,174 @@ def keep_row(self, row: ResultRow):
...
+            The filter name.
+        """
+        return (
+            f"StampMomentsFilter_m01_{self.m01_thresh}_m10_{self.m10_thresh}"
+            f"_m11_{self.m11_thresh}_m02_{self.m02_thresh}_m20_{self.m20_thresh}"
+        )
+
+    def keep_row(self, row: ResultRow):
+        """Determine whether to keep an individual row based on
+        the offset of the stamp's peak.
```
delucchi-cmu, line 135:
```suggestion
        how well the stamp's moments match that of a Gaussian.
```

src/kbmod/filters/stamp_filters.py
```diff
@@ -66,3 +73,174 @@ def keep_row(self, row: ResultRow):
...
+           An indicator of whether to keep the row.
+        """
+        # Filter any row without a stamp.
+        if row.stamp is None:
+            return False
+
+        # Check the stamp's width is correct.
+        width = 2 * self.stamp_radius + 1
+        if len(row.stamp) != width * width:
+            raise ValueError("Expected stamp of size {width} by {width}.")
```
delucchi-cmu, line 154:
```suggestion
            raise ValueError(f"Expected stamp of size {width} by {width}.")
```

src/kbmod/filters/stamp_filters.py
```diff
@@ -66,3 +73,174 @@ def keep_row(self, row: ResultRow):
...
+
+        # Find the peack in the image.
```
delucchi-cmu, line 156:
```suggestion
        # Find the peak in the image.
```

src/kbmod/filters/stamp_filters.py
```diff
@@ -66,3 +73,174 @@ def keep_row(self, row: ResultRow):
...
+           An indicator of whether to keep the row.
+        """
+        # Filter any row without a stamp.
+        if row.stamp is None:
+            return False
+
+        # Check the stamp's width is correct.
+        width = 2 * self.stamp_radius + 1
+        if len(row.stamp) != width * width:
+            raise ValueError("Expected stamp of size {width} by {width}.")
```
delucchi-cmu, line 226:
```suggestion
            raise ValueError(f"Expected stamp of size {width} by {width}.")
```

src/kbmod/filters/stamp_filters.py
```diff
@@ -66,3 +73,174 @@ def keep_row(self, row: ResultRow):
...
+        row : ResultRow
+            The row to evaluate.
+
+        Returns
+        -------
+        bool
+           An indicator of whether to keep the row.
+        """
+        # Filter any row without a stamp.
+        if row.stamp is None:
```
delucchi-cmu, line 220: There's some duplicated code between these methods, namely, checking that there's a stamp and is the correct size. Would it make sense to either create some `BaseStampFilter` that has this stamp checking, or at least break out into a common method?

src/kbmod/filters/stamp_filters.py
delucchi-cmu: And does it make sense to just return false here, instead of throwing an error?

src/kbmod/filters/stamp_filters.py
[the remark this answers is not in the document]
jeremykubica: Fixed.

cchris28 (approved):
minor (as in you can ignore if you want) comments.

src/kbmod/filters/stamp_filters.py
```diff
@@ -66,3 +73,174 @@ def keep_row(self, row: ResultRow):
...
+            and (moments.m20 < self.m20_thresh)
+            and (moments.m02 < self.m02_thresh)
+        )
+
+
+class StampCenterFilter(Filter):
+    """A filter on whether the center of the stamp is a local
+    maxima and the percentage of the stamp's total flux in this
+    pixel.
+    """
```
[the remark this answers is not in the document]
jeremykubica: I'm not sure what is the correct approach here. I'm happy either way.

src/kbmod/filters/stamp_filters.py
jeremykubica: That's fair. Created an abstract base class and moved the validation steps there.

src/kbmod/filters/stamp_filters.py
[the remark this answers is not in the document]
jeremykubica: Sounds good. I moved them.

cchris28 (commented):
I think that https://peps.python.org/pep-3119/ makes specific recommendations for ABC and I am not familiar enough to say this conforms. Mostly it seems that the ABC is for inspection and not as a way to provide default implementations. But I could be wrong. Also maybe you mea this as a base class. I am unsure how inheritance really works in python, it doesn't seem to be straight OOP.

delucchi-cmu (approved):

PR #284, opened 2023-03-22
Update trajectory object during row updates

jeremykubica:
This change probably requires some discussion. 

Previously KBMOD passes through the trajectory object as set during the on-GPU search. This means that if there is later filtering of observations by the sigmaG filter (lh, num_obs, and flux) these are NOT output to the `results_` file. This can lead to numbers in the `results_` file that are not the same as those derived from the other files (e.g. `lc_index_`). This is not a problem if all files are used, but might cause confusion down the line.

This change updates the trajectory to stay consistent with the rest of the values in the row.

A consequence is that this PR changes the results of the diff_test and may impact other analysis scripts.

### DIFF
--- a/src/kbmod/result_list.py
+++ b/src/kbmod/result_list.py
@@ -209,8 +209,12 @@ class ResultRow:
 
         if phi_sum <= 0.0:
             self.final_likelihood = 0.0
+            self.trajectory.lh = 0.0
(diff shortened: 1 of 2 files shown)


DinoBektesevic (approved):
Since this was discussed and everyone approved - so is this PR!

PR #288, opened 2023-03-24
Cleans up more legacy code from KBMOD

jeremykubica:
This PR makes three major clean ups:
1) It removes the `SharedTools` class and support for legacy result dictionaries. All code has been converted to use `ResultList` instead, so this code is all unused.
2) It removes the largely unused `kbmodpy.py` file. The only place this was being used was `test_bilinear_interp.py`. That code has been updated to use the RawImage in order to be consistent with the rest of KBMOD.
3) Removes the `save_results` function from the `Interface` class. This is another (small) step towards getting rid of that class. Replacement of the loader function should wait until the new data interface design is complete though.


DinoBektesevic (approved):
:partying_face: Very happy we are removing these classes. I agree - `Interface` should wait until we agree on, well, the interface.

PR #289, opened 2023-03-26
Update graph library dependency from pylab

### LINKED ISSUE #267: %pylab is deprecated
%pylab is deprecated, use %matplotlib inline and import the required libraries.
Populating the interactive namespace from numpy and matplotlib

This appears in notebooks/kl_divergence_filter.ipynb

A-Chaudhary:
Resolves #267 

Updated `%pylab inline`  with `%matplotlib inline` in notebooks/kl_divergence_filter.ipynb
Imported `matplotlib.pyplot` in notebooks/kl_divergence_filter.ipynb


jeremykubica:
Hi Abhishek. Thanks for submitting a patch for KBMOD. I'm having some trouble with the diffs because of the size of the images. I'm not sure why, but can you clear the notebook output and resubmit? That will allow me to look at the additions in isolation.

PR #291, opened 2023-04-05
Fix broken filtering on different GPU

jeremykubica:
When running KBMOD on one of the local CMU machines the GPU is not running the stamp filtering kernel correctly and as a result failing the unittests. This is a quick patch to get them to pass by running them on CPU.

Longer term the plan is to move stamp filtering off GPU and use the new filtering classes. This will remove some parallelization, but (hopefully) not be a huge slow down (though we will need to bench mark this).

### DIFF
--- a/src/kbmod/search/image_kernels.cu
+++ b/src/kbmod/search/image_kernels.cu
@@ -317,11 +317,7 @@ __global__ void device_get_coadd_stamp(int num_images, int width, int height, fl
     results[trj_offset + pixel_index] = result;
 }
 
-__global__ void device_filter_stamp(int num_stamps, stampParameters params, float *stamps_vect) {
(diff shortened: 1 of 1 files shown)


DinoBektesevic (approved):
I'm approving the changes because we can't really have unittests failing and that this seems to do the same things as before. I will note recent guests have had complaints about how kbmod runs in 5 minutes and then filtering, specifically of stamps, in dense fields can take up to 5h. 

It was insisted that as much as possible of the current filtering stays on GPU. Interestingly they noted that when they're searching for slow moving objects it takes the search significantly longer than for fast moving ones - although this could be when we quit after we left the CCD. 

I'll open an issue to remind us that we should come back and take a look at this.

PR #294, opened 2023-04-06
developer instructions for building local documentation

cchris28:
the required pandoc does not actually get installed with the dependency in setup.cfg so removed that dependency and added documentation to README.md for local documentation build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,17 @@ cmake --build src/kbmod --clean-first
 ```
 To rebuild, it is sufficient to just re-run the `cmake --build` command. Optionally, invoke the cmake generated `Makefile` as `make clean && make` from the `src/kbmod` directory.
 
+If you want to buid the documentation you must have pandoc which seems not installable by pip.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -67,6 +67,17 @@ cmake --build src/kbmod --clean-first
...
+If you want to buid the documentation you must have pandoc which seems not installable by pip.
+See [Pandoc](https://pandoc.org/installing.html), or if you are using conda:
+```
+conda install pandoc
+```
+Building the documentation in docs/build/html
+```
+pip install .[docs]
+cd docs
+make clean html
```
DinoBektesevic, line 79: I think it would be good to mention 

```
sphinx-build -t [html|pdf|xml] src dst
```
or
```
sphinx-build -t html docs docs_build_dst
```

and a [link to the sphinx-build](https://www.sphinx-doc.org/en/master/man/sphinx-build.html) docs. It's equivalent to the make command since the makefiles mostly just call the `sphinx-build` anyhow.

DinoBektesevic (approved):

PR #296, opened 2023-04-06
used example from numba to perform python version check in setup.py

cchris28:


DinoBektesevic (approved):
Thanks so much for fixing this. Sorry for the run-around, this was my bad.

PR #298, opened 2023-04-10
Add end to end tests

jeremykubica:
This PR makes multiple changes:
1) It adds an end-to-end test corresponding to the demo notebook.
2) It makes the output directory optional. If the directory name is set to `None`, it skips the output.
3) It has run_search return the result list directly to facilitate testing and (in the future) allow interactive filtering.
4) It adds an end-to-end test with a bigger stamp size to confirm that `stamp_radius` is fully customizable.


cchris28 (approved):
lgtm

Issue #253, opened 2023-03-03
File-based Configuration


jeremykubica:
Provide KBMOD the ability to read configuration information in from a file so that this information can be source controlled for future reference.

PR #301, opened 2023-04-11
YAML-based configuration

jeremykubica:
Move the configuration from a dictionary to a KBMODConfig object with stricter checking. Allows the user to load configuration from either a dictionary (current behavior) or a YAML file. Adds the YAML config as part of the output in order to create a record of exactly what was used for each run.


jeremykubica:
Closes #253

PR #304, opened 2023-04-12
Create framework for batch filtering

jeremykubica:
This PR creates a framework for batch filtering, adds a clustering-based batch filter, and replaces the current implementation with the filtering version. The unit tests and diff tests all pass.

In addition, the PR effectively drops support for using OPTICS for clustering as that it not implemented as a filter version. However this would be pretty easy to readd if needed.


cchris28 (approved):
looks fine

PR #308, opened 2023-04-18
added tests/test_calc_barycentric_corr.py

cchris28:
as built verification of code against a few scenarios.
I will file an issue to verify if _calc_barycentric_corr as written is sensible as I do not understand it completely
and am not yet convinced that it does what we want it to do.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,6 +45,7 @@ install_requires =
     scipy>=1.9.2
     scikit_learn>=1.0.0
     koffi>=0.1.1
+    jplephem
     PyYAML>=6.0
 
 [options.extras_require]
--- a/tests/test_calc_barycentric_corr.py
+++ b/tests/test_calc_barycentric_corr.py
@@ -59,7 +59,7 @@ class test_calc_barycentric_corr(unittest.TestCase):
         check_ok = True
         for i in range(0, shape_coeff[0]):
             for j in range(1, shape_coeff[1]):
-                if not np.isclose(baryCoeff[i, j], baryExpected[i, j], rtol=1e-8, atol=1e-16):
+                if not np.isclose(baryCoeff[i, j], baryExpected[i, j], rtol=1e-5, atol=1e-14):
                     exception_strings.append(
                         f"baryCoeff[{i},{j}] = {baryCoeff[i,j]:.16f} != {baryExpected[i,j]:.16f}"
                     )
--- a/tests/test_calc_barycentric_corr.py
+++ b/tests/test_calc_barycentric_corr.py
@@ -1,0 +1,247 @@
+import unittest
+
+import astropy.wcs
+import numpy as np
+from astropy import units as u
+from astropy.coordinates import SkyCoord
+from astropy.time import Time
+import math
+
+import kbmod
+import kbmod.analysis_utils
+import kbmod.search
+
+
+class test_calc_barycentric_corr(unittest.TestCase):
+    def setUp(self):
+        # Define the path for the data.
+        # TODO: this should be ../data/demo but vscode does not get the message all the time.
+        # from time import sleep
+        # sleep(300)
+        im_filepath = "../data/demo"
+        # HACK: workaround for vscode not honoring the cwd specification.
+        from os.path import isdir
+
+        if not isdir(im_filepath):
+            im_filepath = "data/demo"
+        # The demo data has an object moving at x_v=10 px/day
+        # and y_v = 0 px/day. So we search velocities [0, 20]
+        # and angles [-0.5, 0.5].
+        v_arr = [0, 20, 21]
+        ang_arr = [0.5, 0.5, 11]
+
+        self.input_parameters = {
+            # Required
+            "im_filepath": im_filepath,
+            "res_filepath": None,
+            "time_file": None,
+            "output_suffix": "DEMO",
+            "v_arr": v_arr,
+            "ang_arr": ang_arr,
+            "bary_dist": 50.0,
+        }
+
+    def _expected_string(self, baryCoeff, shape_coeff):
+        """Return a string of the actual results in a form that can be copied into baryExpected in the test code. Used for as-built testing."""
+        value_string = f"# fmt: off\nbaryExpected = np.array(["
+        value_string += f"\n    #       dx,                      dxdx,                    dxdy,                    dy,                      dydx,                    dydy,"
+        for i in range(0, shape_coeff[0]):
+            value_string += f"\n    [{baryCoeff[i,0]:23.16f}"
+            for j in range(1, shape_coeff[1]):
+                value_string += f", {baryCoeff[i,j]:23.16f}"
+            value_string += f"],"
+        value_string += f"\n]\n# fmt: on"
+        return value_string
+
+    def _exception_strings(self, baryCoeff, baryExpected, shape_coeff):
+        """Return a list of strings of the actual results that violate the expected results and a boolean that is True if all the results are as expected."""
+        exception_strings = []
+        check_ok = True
+        for i in range(0, shape_coeff[0]):
+            for j in range(1, shape_coeff[1]):
+                if not np.isclose(baryCoeff[i, j], baryExpected[i, j], rtol=1e-8, atol=1e-16):
+                    exception_strings.append(
+                        f"baryCoeff[{i},{j}] = {baryCoeff[i,j]:.16f} != {baryExpected[i,j]:.16f}"
+                    )
+                    check_ok = False
+        return exception_strings, check_ok
+
+    def _check_barycentric(self, baryCoeff, baryExpected):
+        """Check the barycentric correction values against the expected values."""
+        # Check the barycentric correction values
+        shape_coeff = baryCoeff.shape
+        value_string = self._expected_string(baryCoeff, shape_coeff)
+        # print(value_string)
+        shape_expected = baryExpected.shape
+        assert shape_coeff == shape_expected
+        assert len(shape_coeff) == 2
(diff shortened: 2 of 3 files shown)


tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+            "im_filepath": im_filepath,
+            "res_filepath": None,
+            "time_file": None,
+            "output_suffix": "DEMO",
+            "v_arr": v_arr,
+            "ang_arr": ang_arr,
+            "bary_dist": 50.0,
+        }
+
+    def _expected_string(self, baryCoeff, shape_coeff):
```
jeremykubica, line 44: As noted below, I don't think this function is needed anymore.

tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+            for j in range(1, shape_coeff[1]):
+                value_string += f", {baryCoeff[i,j]:23.16f}"
+            value_string += f"],"
+        value_string += f"\n]\n# fmt: on"
+        return value_string
+
+    def _exception_strings(self, baryCoeff, baryExpected, shape_coeff):
+        """Return a list of strings of the actual results that violate the expected results and a boolean that is True if all the results are as expected."""
+        exception_strings = []
+        check_ok = True
```
jeremykubica, line 59: Technically you don't need `check_ok` here or below. If `exception_strings` is non-empty, then things are not okay.

tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+        # Check the barycentric correction values
+        shape_coeff = baryCoeff.shape
+        value_string = self._expected_string(baryCoeff, shape_coeff)
+        # print(value_string)
+        shape_expected = baryExpected.shape
+        assert shape_coeff == shape_expected
+        assert len(shape_coeff) == 2
+        exception_strings, check_ok = self._exception_strings(baryCoeff, baryExpected, shape_coeff)
+        if len(exception_strings) > 0:
+            print(f"Exception strings = {exception_strings}")
```
jeremykubica, line 80: Could could return `False` here and then return `True` if the if did not trigger.

tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+            [  -203.5233773925634750,     -0.0000707616325463,      0.0000001649572917,   -510.8750476689951370,      0.0000000748362879,     -0.0000703773895892],
+            [  -301.8886731240853010,     -0.0001138896682213,      0.0000002446735039,   -758.0578046608508203,      0.0000001110405992,     -0.0001133195204643],
+        ])
+        # fmt: on
+        assert self._check_barycentric(baryCoeff, baryExpected)
+
+    def test_single_image(self):
+        """Verifies that the barycentric corrections for a single image are zeros."""
+        run_search = kbmod.run_search.run_search(self.input_parameters)
+        assert run_search is not None
```
jeremykubica, line 173: `self.assertIsNotNone(run_search)`

tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+            [  -102.1013035398333670,     -0.0000324007246175,      0.0000000827567619,   -256.1955780511078160,      0.0000000375305764,     -0.0000322080313739],
+            [  -102.8886591105074473,     -0.0000326747631053,      0.0000000833949229,   -258.1719728533235525,      0.0000000378200863,     -0.0000324805833438],
+            [  -201.5607485205129024,     -0.0000699606225079,      0.0000001633666874,   -505.9449406539421261,      0.0000000741141381,     -0.0000695800875374],
+            [  -202.7383401704241805,     -0.0000704409557553,      0.0000001643210668,   -508.9030324275871067,      0.0000000745474367,     -0.0000700581959558],
+            [  -203.5233773925634750,     -0.0000707616325463,      0.0000001649572917,   -510.8750476689951370,      0.0000000748362879,     -0.0000703773895892],
+            [  -301.8886731240853010,     -0.0001138896682213,      0.0000002446735039,   -758.0578046608508203,      0.0000001110405992,     -0.0001133195204643],
+        ])
+        # fmt: on
+        assert self._check_barycentric(baryCoeff, baryExpected)
```
jeremykubica, line 168: `self.assertTrue(self._check_barycentric(baryCoeff, baryExpected))`

tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+        header_info.populate_from_fits_file(full_file_path)
+        time_obj = header_info.get_epoch(none_if_unset=True)
+        time_stamp = time_obj.mjd
+        visit_times = [time_stamp]
+        img_info = kbmod.image_info.ImageInfoSet()
+        img_info.append(header_info)
+        img_info.set_times_mjd(np.array(visit_times))
+        baryCoeff = run_search._calc_barycentric_corr(img_info, 50.0)
+        assert baryCoeff is not None
+        assert baryCoeff.shape == (1, 6)
```
jeremykubica, line 188: These asserts can be changed into unittest style as above.

setup.cfg
```diff
@@ -45,6 +45,7 @@ install_requires =
     scipy>=1.9.2
     scikit_learn>=1.0.0
     koffi>=0.1.1
+    jplephem
```
jeremykubica, line 48: Is this used anywhere?

tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+import kbmod.analysis_utils
+import kbmod.search
+
+
+class test_calc_barycentric_corr(unittest.TestCase):
+    def setUp(self):
+        # Define the path for the data.
+        # TODO: this should be ../data/demo but vscode does not get the message all the time.
+        # from time import sleep
+        # sleep(300)
```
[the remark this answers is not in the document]
cchris28: yeah. missed that. Thanks.

tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+        im_filepath = "../data/demo"
+        # HACK: workaround for vscode not honoring the cwd specification.
```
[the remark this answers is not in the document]
cchris28: Sort of. I think it is a bug in vscode. Interestingly I got the "run test" to work in vscode but the "debug test" just would not change to the tests directory before running. I searched and did not find a resolution so I just hacked it.
I will remove that hack.

tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def _check_barycentric(self, baryCoeff, baryExpected):
+        """Check the barycentric correction values against the expected values."""
+        # Check the barycentric correction values
+        shape_coeff = baryCoeff.shape
+        value_string = self._expected_string(baryCoeff, shape_coeff)
+        # print(value_string)
+        shape_expected = baryExpected.shape
+        assert shape_coeff == shape_expected
+        assert len(shape_coeff) == 2
```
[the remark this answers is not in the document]
cchris28: ok. Is that true across the board? Should I replace all
assert foo == bar
with self.assertEqual(foo, bar)
?

tests/test_calc_barycentric_corr.py
cchris28: i will change

tests/test_calc_barycentric_corr.py
cchris28: will change

tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+        exception_strings, check_ok = self._exception_strings(baryCoeff, baryExpected, shape_coeff)
+        if len(exception_strings) > 0:
+            print(f"Exception strings = {exception_strings}")
+        return check_ok
+
+    def _construct_wcs(
```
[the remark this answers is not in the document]
cchris28: noted

tests/test_calc_barycentric_corr.py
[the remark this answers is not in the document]
jeremykubica: That is at least the style that kbmod is using. So yes. We should replace them everywhere.

tests/test_calc_barycentric_corr.py
cchris28: done

tests/test_calc_barycentric_corr.py
cchris28: done

tests/test_calc_barycentric_corr.py
cchris28: yup. will do

setup.cfg
cchris28: yes. It is used in _calc_barycentric_corr when a distance is specified. Which it never was until these tests.
line 383
            obs_pos = get_body_barycentric("earth", Time(mjdlist[0], format="mjd"))

tests/test_calc_barycentric_corr.py
```diff
@@ -0,0 +1,247 @@
...
+                    )
+                    check_ok = False
+        return exception_strings, check_ok
+
+    def _check_barycentric(self, baryCoeff, baryExpected):
+        """Check the barycentric correction values against the expected values."""
+        # Check the barycentric correction values
+        shape_coeff = baryCoeff.shape
+        value_string = self._expected_string(baryCoeff, shape_coeff)
+        # print(value_string)
```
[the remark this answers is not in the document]
cchris28: I left it there because it was useful for the as-built and might be useful again for more tests.

jeremykubica (approved):

PR #310, opened 2023-04-18
Small cleanups

jeremykubica:
Remove functions from `image_info.py` that never ended up getting used and remove an unused parameter from `run_search._count_known_matches()`


cchris28 (approved):
lgtm

PR #312, opened 2023-04-19
Refactor Masking Functionality

jeremykubica:
Refactors KBMOD's masking functionality to:
1) Move masking out of `PostProcess`.
2) Make an abstract masking class `ImageMasker` that can be overridden to perform specific types of masking.
3) Move the current functionality into different masking classes.


DinoBektesevic (approved):
I was avoiding reviewing this because I thought it generalized functionality from `do_masking` but it abstracts masking functionality into separate classes. I think this is good because it matches how filtering was refactored so it improves consistency. 

This is good, sorry for the delays.

PR #316, opened 2023-05-02
test least squares linear fit of _barycentric_correction

cchris28:
refactored _calc_barycentric_corr to test least squares part
added test for least squares linear fit.
used benchmark to check that did not affect speed by refactoring.

### DIFF
--- a/src/kbmod/run_search.py
+++ b/src/kbmod/run_search.py
@@ -393,26 +393,26 @@ class run_search:
         # barycentric coordinate is observer position + r * line of sight
         cbary = astroCoords.SkyCoord(obs_pos + r * cobs.cartesian, representation_type="cartesian")
 
-        return run_search._calculate_barycoeff_list(xlist, ylist, wcslist, cbary, obs_pos_list)
+        return self._calculate_barycoeff_list(xlist, ylist, wcslist, cbary, obs_pos_list)
 
-    def _calculate_barycoeff_list(xlist, ylist, wcslist, cbary, obs_pos_list):
+    def _calculate_barycoeff_list(self, xlist, ylist, wcslist, cbary, obs_pos_list):
         """Function to calculate the least squares fit parameters for the barycentric correction.
         Requires that cbary, obs_pos_list and wcslist are defined.
         """
         baryCoeff = np.zeros((len(wcslist), 6))
-        A = np.stack([np.ones_like(xlist), xlist, ylist], axis=-1)
+        coefficients = np.stack([np.ones_like(xlist), xlist, ylist], axis=-1)
         xylist = np.stack([xlist, ylist], axis=-1)
         for i in range(1, len(wcslist)):
             # hold the barycentric coordinates constant and convert to new frame
             # by subtracting the observer's new position and converting to RA/DEC and pixel
             # [bary_to_obs_fast()]
-            baryCoeff[i, :] = run_search._least_squares_fit_parameters(
-                A, xylist, wcslist[i], cbary, obs_pos_list[i]
+            baryCoeff[i, :] = self._least_squares_fit_parameters(
+                coefficients, xylist, wcslist[i], cbary, obs_pos_list[i]
             )
 
         return baryCoeff
 
-    def _least_squares_fit_parameters(A, xylist, wcsi, cbary, obs_posi):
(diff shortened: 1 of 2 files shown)


src/kbmod/run_search.py
```diff
@@ -377,49 +377,61 @@ def _calc_barycentric_corr(self, img_info, dist):
...
         # barycentric coordinate is observer position + r * line of sight
-        cbary = SkyCoord(
-            obs_pos.x + r * cobs.x,
-            obs_pos.y + r * cobs.y,
-            obs_pos.z + r * cobs.z,
-            representation_type="cartesian",
-        )
+        cbary = astroCoords.SkyCoord(obs_pos + r * cobs.cartesian, representation_type="cartesian")
 
+        return run_search._calculate_barycoeff_list(xlist, ylist, wcslist, cbary, obs_pos_list)
```
maxwest-uw, line 396: small nitpick but it's probably more pythonic to have this be
```suggestion
        return self._calculate_barycoeff_list(xlist, ylist, wcslist, cbary, obs_pos_list)
```

src/kbmod/run_search.py
```diff
@@ -377,49 +377,61 @@ def _calc_barycentric_corr(self, img_info, dist):
...
+
+    def _calculate_barycoeff_list(xlist, ylist, wcslist, cbary, obs_pos_list):
+        """Function to calculate the least squares fit parameters for the barycentric correction.
+        Requires that cbary, obs_pos_list and wcslist are defined.
+        """
         baryCoeff = np.zeros((len(wcslist), 6))
-        for i in range(1, len(wcslist)):  # corections for wcslist[0] are 0
+        A = np.stack([np.ones_like(xlist), xlist, ylist], axis=-1)
```
maxwest-uw, line 403: is `A` a standard variable name for this type of calculation? I think a linter might complain that it's a single character and also a capital (python "reserves" FirstCharacterCapitals for classes and ALL_CAPS for constants)

src/kbmod/run_search.py
```diff
@@ -377,49 +377,61 @@ def _calc_barycentric_corr(self, img_info, dist):
...
+        xylist = np.stack([xlist, ylist], axis=-1)
+        for i in range(1, len(wcslist)):
             # hold the barycentric coordinates constant and convert to new frame
             # by subtracting the observer's new position and converting to RA/DEC and pixel
             # [bary_to_obs_fast()]
-            with solar_system_ephemeris.set("de432s"):
-                obs_pos = get_body_barycentric("earth", Time(mjdlist[i], format="mjd"))
-            c = SkyCoord(
-                cbary.x - obs_pos.x, cbary.y - obs_pos.y, cbary.z - obs_pos.z, representation_type="cartesian"
+            baryCoeff[i, :] = run_search._least_squares_fit_parameters(
```
maxwest-uw, line 409:
```suggestion
            baryCoeff[i, :] = self._least_squares_fit_parameters(
```

tests/test_calc_barycentric_corr.py
```diff
@@ -233,6 +241,114 @@ class Params(object):
         # fmt: on
         self.assertTrue(self._check_barycentric(baryCoeff, baryExpected))
 
+    def test_single_image_barycentric(self):
```
maxwest-uw, line 244: this test doesn't actually test for the correctness of the correction, would it be possible to add a check to make sure the corrected values make sense?

src/kbmod/run_search.py
cchris28: Changed. I also had to change the related signatures so I'll not commit the suggestion but commit a change. I don't know how else to do that.

src/kbmod/run_search.py
cchris28: good point. I changed it to "coefficients".

src/kbmod/run_search.py
cchris28: As mentioned I will do this locally and push.

tests/test_calc_barycentric_corr.py
cchris28: urp. I should have removed that test! Will do so now.

maxwest-uw (approved):
lgtm! thanks for the changes

PR #322, opened 2023-05-23
Add helper functions to ImageInfo

jeremykubica:
Add a few helper functions to the ImageInfo class to allow it to store the LayeredImage and to convert the sky position into pixel positions. These are both needed for future PRs which will enable ImageInfo to perform some basic image processing, such as resampling.


cchris28 (approved):
Looks good. You might consider a test that checks the error result if the image has no psf?
Not important enough to block pull though.

PR #324, opened 2023-05-24
Resample LayeredImages based on RA, dec information

jeremykubica:
This is a PR primarily for discussion on whether this technique works. The idea is to be able to create stamps of a `LayeredImage` centered around a given RA and dec.


DinoBektesevic (commented):
All of this looks reasonable and correct to me eyeball inspecting it. I guess one way to make sure to test numerical correctness would be to put down a plus sign and then rotate by 90 by making sure it's drawn perpendicular to the equatorial coordinate axes. Bilinear interpolation then shouldn't change the values in the pixels, because the pixel scale shouldn't change, in this implementation also with the correct choice of angle radius. 
           
I think a plot for this approach is that it's able to combine images with different pixel scales, so I guess good for analyzing different instruments. So I take it that's what your aim was in the example? Otherwise it's not completely clear to me why we need to resample the cutout so that its axes approx. align with the equatorial system. 

The cutout functionality, if we don't force the resampling, can just be outsourced to Astropy (see [here](https://docs.astropy.org/en/stable/nddata/utils.html#saving-a-2d-cutout-to-a-fits-file-with-an-updated-wcs)). It's mainly used as plotting utility I think, but I think the two key features it has that stood out to me is the ability to carry over the WCS and that it doesn't make a copy of the original array when instantiated. 

The last comment I have is that it's confusing to me, from this example, as to what the interface is. I take it, since this is an example, you made it so that it's easier on you - and avoided having to calculate ang and pixel scale. In real life though, what would we be doing? 

Taking a lead from [Reproject](https://reproject.readthedocs.io/en/stable/) by base instinct is that we should look for low-level interface that is based on a region on the sky and pixel-scale. The higher level interface is a collection of images. We determine the smallest bbox that contains all image bboxes and find the image(s) with the smallest pixscale and set that as our pixscale in the resulting reprojection (easier to bin down than to supersample upwards).       

I feel like a lot of that is in reproject package already and I would be inclined to support contributing optimized implementations to that package rather than rolling our own. Perhaps we can get in contact with some of the devs via GH and see what their plans for the package looked like. If it doesn't include implementing some of this stuff at a lower C/C++ level maybe we would be missing out on too much performance.

jeremykubica:
Closing this as #457 provides a better approach to resampling images.

PR #332, opened 2023-06-22
Remove the on-GPU grow mask option

jeremykubica:
More detailed profiling shows that growing the mask on GPU is not providing a significant speedup. Keeping two implementations adds complexity and maintenance overhead. Use the CPU implementation as that is simpler and more portable (no GPU required).

### DIFF
--- a/tests/test_layered_image.py
+++ b/tests/test_layered_image.py
@@ -205,7 +205,7 @@ class test_layered_image(unittest.TestCase):
         mask.set_pixel(10, 13, 1)
         self.image.set_mask(mask)
         self.image.apply_mask_flags(1, [])
-        self.image.grow_mask(1, False)
+        self.image.grow_mask(1)
 
         # Check that the mask has grown to all adjacent pixels.
         science = self.image.get_science()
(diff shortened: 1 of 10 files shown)


tests/test_layered_image.py
```diff
@@ -224,7 +224,7 @@ def test_grow_mask_mult(self):
         mask.set_pixel(10, 12, 1)
         self.image.set_mask(mask)
         self.image.apply_mask_flags(1, [])
-        self.image.grow_mask(3, True)
+        self.image.grow_mask(3)
```
delucchi-cmu, line 227: This case strikes me as odd, and I'm not sure why it's not combined with the one above it. Maybe it's just that the below line ("Check that the mask has grown to all applicable pixels") is not a good description of what's being verified with this case.

Outside the scope of this change, though.

delucchi-cmu (approved):

tests/test_layered_image.py
jeremykubica: That's a good point. There is the potential to combine these two in the future.

PR #333, opened 2023-06-22
Combine similar tests

jeremykubica:
Addressing Melissa's suggestion from #332


delucchi-cmu (approved):
I dig this change! I like that you're testing that you can apply multiple grows and get the expected behavior.

PR #334, opened 2023-07-07
Proposal for a new interface.

DinoBektesevic:
See example usage in [this gist](https://gist.github.com/DinoBektesevic/c93877bd73f3fd47e0094a21feb10ce7).

Note that the functionality within the classes themselves is not necessarily correct. It was a lot of work so some things have cut corners.

- [ ] Some classes are maybe not named the best, see: `DECamCPFits` for example. Rename
- [ ] `DECamCPFits` also uses the [`astro_metadata_translator`](https://github.com/lsst/astro_metadata_translator) package but only works for DECam - no bueno, either make the header translation yourself or make that class general enough to work for all instruments supported by metadata translator. 
- [ ] Everything about how `DECamCPFits` works is completely made up and does not represent the data correctly. Fix how variance is calculated, fix how masks are created. Find datasets to test on because that's the current difficulty.
- [ ] Nearly all `Standardizers` except the `ButlerStandardizer` have made up PSFs - figure out what to do with them
- [ ] `Standardizers` don't take in a config that sets how masks or PSFs are created
- [ ] Write tests, unittests, CI tests etc.
- [ ] Try to look into moving the C++ code onto ndarray or something, the cost of copying the Python arrays is super large and if we could just pass a pointer via numpy arrays API it would basically be a no-op. We could internalize the GPU representation in C++ code if we really need to flatten them on transfer to GPU via the `strides` or something.
- [ ] I'm sure there's more, I left comments strewn throughout the code where I got stuck on something. 

Use it, I'm sure there'll be more ideas on what we can fix or add or change. I'm still trying to work out how best to integrate in the repo.


DinoBektesevic:
I came up with a couple of more want-to's and good to-do's:

- [ ] Serializing `ImageStack`
- [ ] Serializing `LayeredImage`
- [ ] Serializing `RawImage`
- [ ] Potentially round-tripping `ImageCollection` to `ImageStack` and back
- [ ] Look into adopting `lsst.resource` as our internal handler of paths, URLs and URIs
- [ ] Delete the current C++ `load_file` stuff 
- [ ] Delete the `run_search` class
- [ ] Delete the `Interface` class
- [ ] Can we come up with a more Pythonic convention of naming and binding the C++ classes
- [ ] Clean up some of the method names and attributes in C++ (PPI is n_pixels in image, not pixels per inch for example)

These would all be good ways to create an `export` functionality to `ImageStack` in a way that would alleviate the need to constantly create and re-create the `ImageStack` required for processing with KBMOD. It would be a valid solution to the issue of the creation of the `ImageStack` from various data sources being slow but also it would make the staging of the data to process  much easier in batch-like processing environments. If we could make it roundtrip too then we could carry all the metadata along and make `ImageCollection` and `ImageStack` more equivalent, in content even when not in functionality. These files might be large, but they would still be smaller than transferring 100 DECam Community Pipelines processed FITS files with all their CCDs in them just to be able to create the `ImageStack` on the processing resource for 1 targeted CCD. 

The exported FITS file could have a primary header, or a BinTable HDU, containing the internal metadata and indices that could carry the metadata table with it. The challenge would be associating masks, variance and science exposure extensions with each other - which would need a mechanism (a lookup table of names or extension indices) in it. This could be based on the existing indices in the table-row-to-serializer-extension-index but they could also be unrelated to that.

DinoBektesevic:
- [ ] Time the masking code and make sure we haven't lost too much by moving that out of C++. Try to optimize, or just move to C++ if it's that bad.

src/kbmod/standardizers/butler_standardizer.py
```diff
@@ -0,0 +1,265 @@
...
+    @classmethod
+    def canStandardize(self, tgt):
+        # this is pretty hacky - but I'm not importing the entire stack to get
+        # a simple isinstance comparison
+        if "datasetref" in str(type(tgt)).lower():
+            return True, tgt
+        return False, []
+
+    def __init__(self, butler, datasetRefs=None, id=None, **kwargs):
+        # this requires an invovled logic sequence where
```
jeremykubica, line 38: Should we be more prescriptive throughout? A lot of the complexity here results from accepting a wide range of inputs. Is it worth considering less flexibility for more simplicity?

src/kbmod/standardizers/butler_standardizer.py
```diff
@@ -0,0 +1,265 @@
...
+
+        standardizedBBox["center_ra"] = centerSkyCoord.ra.deg
+        standardizedBBox["center_dec"] = centerSkyCoord.dec.deg
+
+        standardizedBBox["corner_ra"] = cornerSkyCoord.ra.deg
+        standardizedBBox["corner_dec"] = cornerSkyCoord.dec.deg
+
+        return standardizedBBox
+
+    def _maskSingleExp(self, exp):
```
jeremykubica, line 139: I really like how the structure works for preparing images. I think masking is a great example of the power of this approach.

src/kbmod/image_collection.py
```diff
@@ -0,0 +1,727 @@
...
+        A table of exposure metadata properties
+    serializers : `list`
+        A list of valid serializer names, used to extract the metadata in its
+        row.
+
+    Attributes
+    ----------
+    data : `~astropy.table.Table`
+        Table of exposure metadata properties.
+    standardizers : `list`
```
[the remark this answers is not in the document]
DinoBektesevic: Multiple standardizers in the same collection. Let's say we have the following situation:

```python
from kbmod import ImageCollection
from kbmod.standardizers import Std1, Std2

# at the moment possible only if the standardizers return the same columns
allStds = []
allStds.extend([Std1(fname) for fname in file_set1])
allStds.extend([Std2(fname) for fname in file_set2])

ic = ImageCollection.fromStandardizers(allStds)
```

then the list would be `Std1.name .... Std1.name, Std2.name .... Std2.name`. 

If our approach is that we require only the mandatory columns to exist for all standardizers then extending the metadata table for standardizers could be handled in two ways:
1) raise an error if they don't have the same columns (current status)
2) extend anyhow and fill with None's 

I think 2) is more intuitive, but I hadn't had the time to add the `add_row(row: Iterable, Mapping)`, `add_rows` and `extend(table: ImageCollection)`.

src/kbmod/image_collection.py
```diff
@@ -0,0 +1,727 @@
...
+            frame="geocentrictrueecliptic",
+        )
+        guess_pixel_coord = guess_ecliptic_coord.to_pixel(wcs)
+
+        # calculate the distance, in pixel coordinates, between the guess and
+        # the start pixel. Calculate the angle that represents in the image.
+        x_dist, y_dist = np.array(guess_pixel_coord) - start_pixel
+        return np.arctan2(y_dist, x_dist)
+
+    def run(self, config):
```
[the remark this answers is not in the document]
DinoBektesevic: Yes! This was more for demo and because this function drives requirements. 

Completely unrelated to this, I decided to go on a rant and write it down so I don't forget:

Looking at it, we can't really run processing because we can't avoid calling 

```python
suggested_angle = self._calc_suggested_angle(self.wcs[0], center_pixel)
```
This is where the assumptions that all the images are looking at the same point in the sky and that they're all the same sizes is coded. I think it's instructive to think about what the interface to the processing is. 

As is - processing basically doesn't require any state tracking, so a class wouldn't have any state keeping that the current `Results` don't already. So it doesn't feel like it should be a class. Really, it already is a class - the `KBMODSearch` or `search` in the code below.

So what the `run_search` would do is just unravel the config parameters and move the data from Python to C++. 
And either of those are really functions of things that already exist. So I do wonder whether or not this whole thing can be made to just be a `run(ImageCollection, config)` convenience function in the sense that it returns the filtered results and not the `KBMODSearch` object and that we should improve the `KBMODSearch` interface.

I think the hard thing about doing this refactoring is developing a meaningful, purposeful plan forward where we don't just "add a feature to support mapping multiple instruments to KBMOD abstraction", but holistically develop a cohesive interface that brings the code together and makes it act like a package and not a collection of functionalities. 

This includes reasoning about interfaces between the data abstraction code in C++, Python, their interaction with the data processing code (both in C++ and in filtering in Python) and the end user interface in the form of Results. I'm not for accepting this abstraction into the code if all we want is a mapper between different data sources and KBMOD processing. That can also be achieved by making https://github.com/dirac-institute/kbmod/blob/94cb6d75c139de8fbd03082e3b533e50cccefc0b/src/kbmod/image_info.py#L28
an abstract method and then providing a bunch of default mapper functions like `decam_mapper(file_path) --> dict` and then saying `ImageInfo(mapper=decam_mapper)` or `ImageInfo(callable=my_mapper)` and keep no states or complexity like we do here.

src/kbmod/standardizers/butler_standardizer.py
DinoBektesevic: I'm not sure what you mean exactly. There are a few moving parts here. The instantiation, the serialization and then future-proofing. 

The current implementation covers cases 1, 4.1 and 4.2 and spans lines 70-77 lines. So I don't know if I'd call that complicated. 

The comment is super long and mostly is a note to myself to pay attention how to map butler interface to our abstraction. The butler is complicated interface, most of which we don't support and I don't know which bits we should. Like if we have to find a way to support case 2) - I hadn't written it out but there's probably half a dozen different ways to query a butler down to `DatasetRefs`, or `DatasetIDs` and most of them can be combined. So I don't know what cases to support, or how to abstract the interface in a way that might not require us to rewrite it once we get some users to actually use it. This leaves us a bit exposed because we can't future-proof this implementation in the same way that  `FitsStandardizer` does (pretty much covers all possible cases).

Maybe you're asking why is `DataID` even here? That's the third problem: we're not always in control of the serialization of objects from the Rubin's Stack.                
We want to serialize `ImageCollection` for share-ability and "transport" (remember the ping-pong match between Hyak and Epyc for processing of large datasets). But the serialization happens in `ImageCollection`, which itself can't know about the individual `Standardizers`. Like, what happens in the cases when we can't construct them because data doesn't exist on that machine or the internet isn't available for a remote Butler - we couldn't make the Standardizers to get them to serialize themselves!                
Since we can `get_standardizer` of a row in `ImageCollection`, we need to serialize sufficient information to construct the Standardizer back out at the other end again.              
So we're in a situation where `ImageCollection` has to know how to serialize itself, that can't be interfaced by the `Standardizers`, and we can't can't guarantee they produce only serializable values, or that they standardize sufficient data to instantiate themselves. So in this case, to demonstrate that, I serialize "DataIds" in the default `standardizeMetadata` method and then resolve the `DataId`s to `DatasetRef`s on the round-trip. Is this a problem? I don't know. It certainly feels like an edge-case.            
Why just an edge-case? Because we provide the default standardizers for most use-cases, we can always guarantee that we will always standardize at least sufficient information to instantiate them by default; therefore for all extensions of them also. We might enforce that in tests, but if a user inherits from the most general `Standardizer` and attempts to implement things from scratch and skips the tests - they just might shoot themselves in the foot when they try to write it to a file. So this is like feature not a bug, in the sense "we take care of you so you don't have to"?          

The last way I think in which I can interpret your questions is that you generally find the `__init__`s not prescriptive enough? I don't know, I feel like this one is ok, `Standardizer` really just wants the `location`, `FitsStandardizer` and `SingleExtensionFits` also. `MultiExtensionFits` has a new magic kwarg - `set_exts`, which is sort of a micro-optimization for a default functionality because it's expensive and I don't want to have to run it.            
`ButlerStandardizer` takes `DatasetRef`s and `DataID`s but this is a consequence of the fact that we support retrieving `get_standardizer` from `ImageCollection`. Maybe that needs to go, but then how does `toImageStack` method work?                     
There is one complex init - the `ImageCollection`. I've seen worse, but that's not really an excuse. Unfortunately I'm not sure how avoidable that is. The issue there is that we serialize internals into a data-table and then have to unserialize them and set them as some class properties. This requires to check if all the columns exist, and then get the column values and properly construct these internal objects. Thankfully there's only a very small limited set of them that exist, but it does produce a cacophony of `if`s I'm not a fan of either. I'm happy to hear suggestions, since off the top of my head I wouldn't know how to reduce that. I can refactor it into multiple sub-inits for separation and readability?

jeremykubica: I was actually talking about the first part. I saw the detail in the comment and thought some of that was placeholder for code that needed to be added. If lines 70-77 already implement all of this, then I have no concerns.

src/kbmod/image_collection.py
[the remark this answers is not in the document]
jeremykubica: Sounds good.

src/kbmod/image_collection.py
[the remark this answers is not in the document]
jeremykubica: I strongly agree on needing to purposefully plan a cohesive interface.

src/kbmod/standardizers/standardizer.py
```diff
@@ -0,0 +1,589 @@
...
+__all__ = ["Standardizer", "StandardizerConfig", "ConfigurationError"]
+
+
+class ConfigurationError(Exception):
+    """Error that is raised when configuration parameters contain a logical error."""
+    pass
+
+
+class StandardizerConfig:
+    """Standardizer configuration.
```
jeremykubica, line 30: It might be worth pointing out that this is a base class for standardizer configs. The next line implies it.

src/kbmod/standardizers/standardizer.py
```diff
@@ -0,0 +1,589 @@
...
+        self.config = self.configClass(config)
+
+    def __str__(self):
+        return f"{self.name}({self.location}, {self.processable})"
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}({self.location})"
+
+    # all these should probably be named in plural - but lord allmighty that's
+    # a lot of references to update and it feels so unnatural - help? These lead
```
jeremykubica, line 388: In cases like this, I usually name the function something like `wcs_list` instead of `wcses` (is that even how you'd spell the plural). I don't think it is worth updating all the references though.

src/kbmod/standardizers/standardizer.py
```diff
@@ -0,0 +1,589 @@
...
+
+        Returns
+        -------
+        mask : `list[~np.array]`
+            Mask images.
+        """
+        raise NotImplementedError()
+
+    # no idea really what to do bout this one? AFAIK almost no images come
+    # with PSFs in them
```
jeremykubica, line 550: What is Rubin's plan for providing PSFs? Is is a separate query?

src/kbmod/image_collection.py
```diff
@@ -0,0 +1,684 @@
...
+        else:
+            raise ValueError(f"Metadata is {explanation}")
+
+        if "std_name" in metadata.columns:
+            self._standardizer_names = metadata["std_name"]
+        elif standardizers is not None:
+            self._standardizer_names = [std.name for std in standardizers]
+            self.data["std_names"] = self._standardizer_names
+
+        if standardizers is not None:
```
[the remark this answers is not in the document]
DinoBektesevic: Sorted, this whole section was cleaned up and all the extra test-stuff that never ended up being used was removed. Worth paying it another look.

src/kbmod/image_collection.py
```diff
@@ -0,0 +1,684 @@
...
+            n_entries = len(np.unique(metadata["location"]))
+            self.data.meta["n_entries"] = n_entries
+            self._standardizers = np.full((n_entries, ), None)
+
+        # hidden indices that track the unravelled lookup to standardizer
+        # extension index. I should imagine there's a better than double-loop
+        # solution and a flat lookup table.
+        # If they weren't added when standardizers were unravelled it's
+        # basically not possible to reconstruct them. Guess attempt is no good?
+        no_std_map = False
```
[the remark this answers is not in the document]
DinoBektesevic: Almost all of them are like this. DECam processed via the Science Pipelines will have 1 FITS file per CCD. DECam processed via Community pipelines will have 1 FITS file per 62-72 CCDs depending on the data product in question. Raws will have 72 (focus and tracking CCDs) HDUs, calibrated exposures will have 62 HDUs (just the science CCDs). 

This doesn't mean that the Rubin Sci. Pipes. will have only 1 HDU. They have ~16HDUs per FITS file processed to a calexp and ~30 for coadds. It's just that the other HDUs hold data like PSF, mask, variance etc. whereas for CommunityPipeline products these are separate data products (different FITS files for example).

More details are given in the following comment and the design document as well.

src/kbmod/image_collection.py
```diff
@@ -0,0 +1,684 @@
...
+        -------
+        ic : `ImageCollection`
+            Image Collection
+        """
+        unravelledStdMetadata = []
+        for i, stdFits in enumerate(standardizers):
+            # needs a "validate standardized" method here or in standardizers
+            stdMeta = stdFits.standardizeMetadata()
+
+            # unravel all standardized keys whose values are iterables unless
```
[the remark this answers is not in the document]
DinoBektesevic: Writing a different standardizer is like having a user define this for us. This proposal is assuming that all the data required can be located in the PRIMARY header and the selected HDU extension. This is not the case in general.

In a simplistic description, Rubin defines their data products as collection of a single CCD exposure alongside with their definitions of `SkyWcs`, `PSF`, mask, variance, catalog metadata etc. The DECam CommunityPipelines define their data products as collection of images from all CCDs at the same time. For CommunityPipe data products the variance, mask, PSF etc. are all different data products that live separately from their "image" data products (raw, calibrated, coadded imdiffed etc.). For example such as in a different FITS file or in an online service-like database thing. 

For different Sci Pipe and Community Pipe data products these "CCD"s include usable and not-usable pixels (such as overscan and bias estimation regions for example) and sometimes not. This sometimes makes it so that one can calculate the variance and bias planes from the data they already have, and sometimes they might just have to go and fetch the additional data they need. Sometimes they may have already fetched that data and have it on disk somewhere. A lot of the times, this is actually (annoyingly) encoded in names of files - for example SDSS relies on that to figure out what's what. This means that sometimes to figure out what file/access mechanism we need, we would need to look at more than just the primary header and the given HDU or provide a definition that's a convention in nature.

So to sum it up the data products we get we have to observe holistically because standardizers encode rules and conventions that differ from a data source to data source, and more specifically user to user. Standardizers encode not just the conventions, but the mechanism by which those conventions are translated to internal KBMOD abstractions (f.e. a numpy arrays of zeros and ones for masks).

src/kbmod/standardizers/fits_standardizers/multi_extension_fits.py
```diff
@@ -0,0 +1,109 @@
...
+            return False
+
+        # The problem is that all kinds of things are stored as ImageHDUs, say
+        # a 120k x 8000k table (I'm looking at you SDSS!). To avoid that we
+        # need to check the data shape. The problem is that causes the data to
+        # load from disk and that is very expensive
+        if len(hdu.shape) != 2:
+            return False
+
+        if hdu.shape[0] > 6000 or hdu.shape[1] > 6000:
```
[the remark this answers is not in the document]
DinoBektesevic: Not by default because this isn't called by default. It's a generic "is this an image" test. It's not executed by default because it's very costly as it reads the actual image into memory. I guess I was trying to provide a reasonable default implementation, regardless of speed, but I could remove this tbh since using this would be nearly impossible with any larger dataset. I could, if you want to leave it as is, also make the value a thing of the config class I guess?

src/kbmod/standardizers/standardizer.py
DinoBektesevic: Yeah, I've no idea either :man_shrugging: Happy to update at a later point if needed.

src/kbmod/standardizers/standardizer.py
DinoBektesevic: It's in the FITS file (that is, it's part of the object you can retrieve via the Butler), but it's their own internal representation of a PSF - so this isn't something we can just rely on as a generic thing. We can make the `ButlerStandardizer` use it to load up the object in Python - but then we need to do something with that object, like evaluate a realization of a PSF at some wavelength and position on the CCD as a numpy array for example. But even then we need to line up that realization with pixel size basically, so that the realization is basically over the same pixels as our images, i.e. so their physical sizes are comparable, because that's how the current PSF class works - the kernel dimension is directly related to the pixel size.  For other instruments I've no idea what to do tbh.

A part of the problem is that PSF class we have is not a good representation of a PSF. Not because Gaussians are bad functional representation of a PSFm but because it only stores a representation of a PSF as an array and not some functional expression.
But even then, as far as I have seen there are no ways the [Rubin PSF class](https://github.com/lsst/afw/blob/main/include/lsst/afw/detection/Psf.h#L75) can be "exported" as any of the to extract the functional PSF forms from AstroPy (f.e. like [moffat](https://docs.astropy.org/en/stable/api/astropy.convolution.Moffat2DKernel.html)) or [photutils](https://photutils.readthedocs.io/en/stable/psf.html) - which again wouldn't even help us that much because we still need an C++ representation of the same again if we want to evaluate it at points (in a way that is generic to where we can do things like PSFs that vary across the CCD) and so on. I guess we can evaluate it and then fit it - but the costs involved here are kinda crazy.

tests/test_standardizer.py
```diff
@@ -0,0 +1,357 @@
...
+        if not self.testing_kwargs:
+            self.required_flag = False
+        else:
+            self.required_flag = required_flag
+        self.optional_flag = optional_flag
+
+    def translateHeader(self):
+        # invoke the parent functionality to standardize the default values
+        metadata = super().translateHeader()
+        metadata["required_flag"] = False
```
jeremykubica, line 56: Could the next few lines just be:
```
metadata["required_flag"] = self.required_flag
metadata["optional_flag"] = self.optional_flag
```

tests/test_standardizer.py
```diff
@@ -0,0 +1,357 @@
...
+        if self.required_flag:
+            metadata["required_flag"] = True
+        if self.optional_flag:
+            metadata["optional_flag"] = True
```
jeremykubica, line 60: No value set if `self.optional_flag` is false?

tests/test_standardizer.py
```diff
@@ -0,0 +1,357 @@
...
+        # One of them will be the MyStd
+        warnings.filterwarnings(action="ignore", category=UserWarning, message="Multiple standardizers")
+
+    def tearDown(self):
+        # restore defaults
+        MyStd.resolveTarget = MyStd.noStandardize
+        MyStd.priority = 3
+        warnings.resetwarnings()
+
+        # self.fits.close(output_verify="ignore")
```
jeremykubica, line 79: Is this possibly needed in the future or can we remove it?

tests/test_standardizer.py
```diff
@@ -0,0 +1,357 @@
...
+# Test KBMODV1 more extensively than other standardizers to cover for the
+# possible code-paths through Standardizer itself. TODO: eventually update this
+class TestKBMODV1(unittest.TestCase):
+    """Test KBMODV1 Standardizer and Standardizer."""
+
+    def setUp(self):
+        self.fits = FitsFactory.mock_fits(spoof_data=True)
+
+    def tearDown(self):
+        # Note that np.int32 in setUp is necessary because astropy will raise
```
jeremykubica, line 162: `setUp` doesn't have `np.int32`. Should it?

tests/test_standardizer.py
```diff
@@ -0,0 +1,357 @@
...
+        # neighbors.
+        self.fits["MASK"].data[2, 2] = KBMODV1Config.bit_flag_map["BAD"]
+
+        conf = StandardizerConfig({"grow_mask": True, "grow_kernel_shape": (3, 3)})
+        std = Standardizer.get(self.fits, force=KBMODV1, config=conf)
+        mask = next(std.standardizeMaskImage())
+
+        # Note that this is different than masking via Manhattan neighbors -
+        # which can be implemented by using the C++ functions in KBMODV1, do I?
+        # the solution now is an masked square in the center of the array
```
jeremykubica, line 324: I think we asserted that it likely won't matter practically. So we can leave this approach for now and check if we see a massive drop in the number of candidate tracks.

src/kbmod/image_collection.py
```diff
@@ -0,0 +1,510 @@
...
+        if not metadata:
+            return False, "an emtpy table."
+
+        cols = metadata.columns
+        missing_keys = [key for key in self.required_metadata if key not in cols]
+        if missing_keys:
+            return False, f"missing required columns: {missing_keys}"
+
+        # check that none of the actual required column entries are empty in
+        # some way. perhaps we should be checking np.nan too?
```
jeremykubica, line 123: We are very inconsistent about checking for NaNs throughout KBMOD, but we should.

src/kbmod/image_collection.py
```diff
@@ -0,0 +1,510 @@
...
+    def get_zero_shifted_times(self):
+        """Returns a list of timestamps such that the first image
+        is at time 0.
+
+        Returns
+        -------
+        List of floats
+            A list of zero-shifted times (JD or MJD).
+        """
+        # what's the deal here - are we required to be sorted?
```
jeremykubica, line 477: It was assumed in the original KBMOD code. We should move off that, but we'd need to change things throughout.

tests/test_butlerstd.py
```diff
@@ -0,0 +1,346 @@
...
+        mocked.visitInfo.date.toAstropy.return_value = Time(hdul["PRIMARY"].header["DATE-AVG"], format="isot")
+        mocked.info.id = prim["EXPID"]
+        mocked.getWidth.return_value = hdul[1].header["NAXIS1"]
+        mocked.getHeight.return_value = hdul[1].header["NAXIS2"]
+        mocked.info.getFilter().physicalLabel = prim["FILTER"]
+
+        # Rubin Sci. Pipes. return their own internal SkyWcs object. We mock a
+        # Header that'll work with ButlerStd instead. It works because in the
+        # STD we cast SkyWcs to dict-like thing, from which we make a WCS. What
+        # happens if SkyWcs changes though?
```
jeremykubica, line 137: We could add a test that creates a SkyWcs and makes sure it can convert to a dict-like thing? Or, more realistically we can do a fix if it changes later.

src/kbmod/image_collection.py
```diff
@@ -0,0 +1,510 @@
...
+        filepath : `str`
+            Path to the file containing the serialized image collection.
+
+        Returns
+        -------
+        ic : `ImageCollection`
+            Image Collection
+        """
+        metadata = Table.read(*args, format=format, units=units, descriptions=descriptions, **kwargs)
+        metadata["wcs"] = [WCS(w) for w in metadata["wcs"] if w is not None]
```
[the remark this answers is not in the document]
DinoBektesevic: Great spot!

src/kbmod/standardizers/standardizer.py
```diff
@@ -0,0 +1,591 @@
...
+        Collection of configuration key-value pairs.
+    kwargs : optional
+        Keyword arguments, assigned as configuration key-values.
+    """
+
+    def __init__(self, config=None, **kwargs):
+        # This is a bit hacky, but it makes life a lot easier because it
+        # enables automatic loading of the default configuration and separation
+        # of default config from instance bound config
+        keys = list(set(dir(self.__class__)) - set(dir(StandardizerConfig)))
```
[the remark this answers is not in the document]
DinoBektesevic: I'm a bit sketchy about the idea of adding logic to a configuration class (that doesn't affect IO but the config values). 

In general I can see why it's useful, but it tends to complicate things a great deal and can usually be refactored into the standardizer itself at a cost of 1 more flag. For example `do_masking` and `grow_mask`. We don't have a method setting `grow_mask` value if `do_masking` is set - we just ignore it. My initial reaction was that if there's a `compute_file_offset` that would be the job of a Standardizer, who would look up a `self.config.byte_offset` value if it's set and do a thing; but maybe I'm missing something. Could you explain a bit more what do you think as a "standardizer helper function"? Standardizers are, by definition, the helper functions that help translate data on disk to KBMOD abstraction - what parameters would we be wanting to make "derived params" [parameters depending on value of other parameters] for a Standardizer?

To answer your question though - yes, if you add a method to a subclass of `StandardizerConfig` this will add it to the `self._conf` dict. If you update, overwrite or replace it - this will replace it. In the list of `keys` this will appear as a string as you'd expect. In the list of `values` it would appear as a callable. If a `config` is provided that contains the same named key - it will be overridden with that value. This is all in line with what generally you can do in Python. 
In an image-collection this might be a problem, as we serialize all of the configs and we can't serialize a callable. So it wouldn't roundtrip. It is possible to filter out `"method"` from a stringified representation of attributes and get rid of it and expect it to exist on the roundtrip. This can be done here or in the image-collection.

src/kbmod/standardizers/standardizer.py
```diff
@@ -0,0 +1,591 @@
...
+        Implementation is instrument-specific.
+        """
+        return cls.resolveTarget(tgt)[0]
+
+    def __init_subclass__(cls, **kwargs):
+        # Registers subclasses of this class with set `name` class
+        # parameters as availible standardizers. Note the specific
+        # standardizer has to be imported before this class since the
+        # mechanism is triggered at definition time.
+        name = getattr(cls, "name", False)
```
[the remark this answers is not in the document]
DinoBektesevic: Nice! Done.

src/kbmod/standardizers/standardizer.py
```diff
@@ -0,0 +1,591 @@
...
+        if name and name is not None:
+            super().__init_subclass__(**kwargs)
+            Standardizer.registry[cls.name] = cls
+
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
DinoBektesevic: In short, I don't want to allow the three following things:
* direct instantiation of Standardizer
* enforce setting of location (which I should probably rename `tgt` or something for generality) and `config`
* enforce existence of `processable`

If I make `__init__` non-abstract then the user could inherit `Standardizer` and not overwrite the `__init__`. Then they are not forced to handle the data source IO in it (tests should hopefully enforce that it roundtrips in `ImageCollection`). If they don't do that and don't call `super().__init__(location="something")` - the location won't be set to anything meaningful. Even if it isn't meaningful - the `config` doesn't get instantiated. This is the only place that sets `self.config = self.configClass(config)` so it must execute - otherwise each `Standardizer` wouldn't have a `config` by default either. 

I guess the core of your question is "how does abstractmethod execute anything at all"? There are no short answers to that, but I'm happy to chat irl. Python's `__init__` is just one step in the construction of an instance. Before it, the `metaclass` mechanism is executed where `__new__` is called, builds an object using [descriptors](https://docs.python.org/3/howto/descriptor.html#descriptor-howto-guide) and then various hooks can be called (like the `__init__subclass__` we use). These already see fully initialized descriptors so they can do whatever they want to them. After that which the `__init__` itself it called. In this process, `__init__`, is supposed to just initialize (not so much construct) the values of descriptors. 

`Standardzier` inherits from `abc.ABC` Abstract Base Class, which does part of this heavy-lifting in the `__init__` and we leverage `__init_subclass__` to keep the `registry` of subclasses working. So that's why declaring things in the `__init__` actually executes - as long as the user follows the golden rule of invoking the `super().__init__` once they inherit. For most part the complicated IO mechanisms are hidden from them, like in the `FitsStandardizer` for example that provides a generic IO for FITS files and HDUList objects, and then provides a reasonable best-guess default behavior for `processable` in `SingleExtensionFits` and `MultiExtensionFits` classes. Inheriting from those, they would just have to worry about writing the `.standardize*` methods.

src/kbmod/standardizers/butler_standardizer.py
```diff
@@ -0,0 +1,337 @@
...
+    do_bitmask = True
+    """Mask ``mask_flags`` from the mask plane in the FITS file."""
+
+    do_threshold = False
+    """Mask all pixels above the given count threshold."""
+
+    grow_mask = True
+    """Grow mask footprint by ``grow_kernel_shape``"""
+
+    brightness_treshold = 10
```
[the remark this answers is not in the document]
DinoBektesevic: The Rubin images are normalized to 0 I think, so the values should be reasonable? But I have no idea. This is copied directly from the old before mask refactoring and were set on probably 1 or 2 example images I had. We can adjust later.

https://github.com/dirac-institute/kbmod/blob/88afe4628570ff8033b0fad533d092fc9fead6a8/src/kbmod/run_search.py#L77

but the 

https://github.com/dirac-institute/kbmod/blob/88afe4628570ff8033b0fad533d092fc9fead6a8/src/kbmod/configuration.py#L61

so this would never execute. I did the same in this PR, I just made it a flag that is `False` by default: 

https://github.com/dirac-institute/kbmod/blob/c00519f20992f9a27294089dfc12c4ef29ccf7fc/src/kbmod/standardizers/butler_standardizer.py#L64

because I rather had multiple values to set than to mix logic with numerical parameters. Should I remove to match the masking refactor?

src/kbmod/standardizers/fits_standardizers/fits_standardizer.py
```diff
@@ -0,0 +1,441 @@
...
+    """
+
+    # Standardizers we don't want to register themselves we leave nameless
+    # Since FitsStd isn't usable by itself - we do not register it.
+    # name = "FitsStandardizer"
+    # priority = 0
+    configClass = FitsStandardizerConfig
+    """The default standardizer configuration."""
+
+    extensions = [".fit", ".fits", ".fits.fz"]
```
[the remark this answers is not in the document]
DinoBektesevic: That is confusing indeed. Thanks!

src/kbmod/standardizers/fits_standardizers/fits_standardizer.py
```diff
@@ -0,0 +1,441 @@
...
+        if hdulist is None and (location == ":memory:" or not isfile(location)):
+            raise FileNotFoundError("Given location is not a file, but no " "hdulist is given.")
+
+        # Otherwise it's pretty normal
+        # - if location is ok and no HDUList exists, read location
+        # - if HDUList exists and location doesn't, try to get loc from it, put
+        # :memory: otherwise
+        # - if hdulist and location exist - nothing to do.
+        if location is not None and hdulist is None:
+            hdulist = fits.open(location)
```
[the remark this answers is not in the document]
DinoBektesevic: Added. Although, there are no guarantees that will ever execute. From whatever I read, Python's `__del__` is basically useless.

src/kbmod/standardizers/fits_standardizers/fits_standardizer.py
```diff
@@ -0,0 +1,441 @@
...
+                 degrees, of the extracted, or calculated, on-sky poisiton of
+                 the data.
+        ======== ==============================================================
+
+        Returns
+        -------
+        metadata : `dict`
+            Required and optional metadata.
+        """
+        raise NotImplementedError("This FitsStandardizer doesn't implement a " "header standardizer.")
```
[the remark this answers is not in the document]
DinoBektesevic: Black slams multi-line comments to one line in a very frustrating way!

src/kbmod/standardizers/fits_standardizers/rubin_scipip_fits.py
```diff
@@ -0,0 +1,84 @@
+"""Class for standardizing FITS files produced by the Vera C. Rubin
+Science Pipelines as they were specified during the KBMOD V1.0 development.
+
+There are no guarantees that current Rubin Science Pipeline Data Products can
+still be standardized with this class. To ensure correct behavior for any
+version of the Rubin Stack, use the `ButlerStandardizer`.
+"""
+
+from .configurable_standardizer import ConfigurableStandardizer
```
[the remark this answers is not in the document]
DinoBektesevic: Uh oh. I seem to have made a mistake when rebasing and squashed two opposing/different commits into one. This is the cause of all these confusing errors. Several later comments are also solved by fixing this. Embarrassing :flushed:

src/kbmod/standardizers/fits_standardizers/rubin_scipip_fits.py
```diff
@@ -0,0 +1,84 @@
+"""Class for standardizing FITS files produced by the Vera C. Rubin
```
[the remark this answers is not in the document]
DinoBektesevic: Stupid mistake, see above, removed.

tests/test_collection.py
```diff
@@ -0,0 +1,10 @@
+import unittest
+
+from kbmod import ImageCollection, Standardizer
+
+from kbmod.analysis_utils import *
+from kbmod.fake_data_creator import add_fake_object
+from kbmod.result_list import *
+from kbmod.search import *
+
+from utils.utils_for_tests import get_absolute_data_path
```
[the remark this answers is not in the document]
DinoBektesevic: Deleted, again, see above. My bad.

tests/test_butlerstd.py
```diff
@@ -0,0 +1,346 @@
...
+        self.assertTrue(standardized["meta"]["bbox"])
+
+    def test_roundtrip(self):
+        """Test ButlerStandardizer can instantiate itself from standardized
+        data and a Data Butler."""
+        std = Standardizer.get(DatasetId(8), butler=self.butler)
+        standardized = std.standardize()
+
+        std2 = ButlerStandardizer(**standardized["meta"], butler=self.butler)
+        self.assertIsInstance(std, ButlerStandardizer)
```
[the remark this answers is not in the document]
DinoBektesevic: Right, testing `std2` roundtrips from standardized data was the point so this would only make sense. Fixed to test `std2` and added a comparison.

tests/test_standardizer.py
DinoBektesevic: That would make sense and would be much easier to read. Unfortunately I'm not sure I can do the `optional_flag` the same way. See below.

tests/test_standardizer.py
DinoBektesevic: It's not supposed to exists in the `metadata` at all, tests are `assertIn`. Leaving this unresolved if you have a better suggestion.

tests/test_standardizer.py
DinoBektesevic: Removed.

tests/test_standardizer.py
DinoBektesevic: Mocking used to be here, but was moved to `utils`. Moved the comment to an appropriate location there. Adjusted the comments there to include, or better explain really, what I was trying to say here.

tests/test_standardizer.py
DinoBektesevic: Right, I feel like a lot of work went into them so I'm loath about just deleting them. Much of the C_++ functionality would work just fine as a C++ wrapped python function, would likely be faster than what we have, and make our lives easier long-term too.

It would be kind of hard to stuff the current masking setup into standardizers so I appreciate the flexibility and patience!

src/kbmod/image_collection.py
DinoBektesevic: Should I open an issue?

src/kbmod/image_collection.py
DinoBektesevic: It wouldn't be that big of a deal to sort the table in `ImageCollection.__init__` or in `ImageCollection.fromStandardizer`. 

Former is easier as we just blindly always sort, but we pay that cost all the time, `fromStandardizer` require us to just be a bit careful about inserting things into the table. [Reading up on it](https://numpy.org/doc/stable/reference/generated/numpy.sort.html), it seems that while there are no guarantees, usually one should expect a smaller cost of sorting for an already sorted list.

Do you have any preferences?

tests/test_butlerstd.py
DinoBektesevic: `SkyWcs` comes from the [`lsst.afw`](https://github.com/lsst/afw), if I mock it I can mock it in a way that it dictifies, but that still doesn't tell me anything about the real `SkyWcs` from `AFW` C++ code used by Rubin. Thankfully this is not likely to change, but generally doing these tests is a pain. We need a canary CI that will build from Rubin weeklies, but the setup is hard so it's a big job to do and the CI run itself will probably go on for a few hours as it would need to build the weekly, run some data through it and then run the KBMOD tests.

This has been my long-time goal for a while though.

src/kbmod/image_collection.py
jeremykubica: Yes. Let's do that. It's a good first issue.

src/kbmod/image_collection.py
jeremykubica: Let's always sort then.

src/kbmod/standardizers/butler_standardizer.py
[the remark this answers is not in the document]
jeremykubica: If you would prefer to have a separate logic value and numeric threshold, I'm fine keeping it that way.

tests/test_butlerstd.py
jeremykubica: I'm also not sure how to create a `SkyWcs` either. I was just suggesting that, if we could create one, that we could use that to watch for breakages.

tests/test_standardizer.py
jeremykubica: No better suggestion.

DinoBektesevic:
Blocked by https://github.com/dirac-institute/kbmod/pull/440 
Do not merge without rebasing after #440 merges.

jeremykubica (approved):

PR #335, opened 2023-09-14
Make GPU Code optional

jeremykubica:
Changes the CMake and setup file to compile in CPU only mode in cases where there is no CUDA compiler or valid GPU. Adds `#ifdef` overrides to the C++ code to avoid GPU specific code in CPU-only mode.

However this currently only raises an error when we try to call a GPU function. We will need to replace the GPU code with CPU equivalents (where we want them) for tests to pass on non-CPU machines.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ endif()
 
 project(search LANGUAGES CXX)
 
-# Check if we can compile CUDA on this system and, if so, enable the options.
+# Check if we can compile CUDA on this system.                                                                             
 include(CheckLanguage)
 check_language(CUDA)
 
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -3,69 +3,86 @@ if(${CMAKE_VERSION} VERSION_LESS 3.12)
...
+      PREFIX "${PYTHON_MODULE_PREFIX}"
+      SUFFIX "${PYTHON_MODULE_EXTENSION}"
+  )
+  if(ipo_supported)
+      set_property(TARGET searchcu PROPERTY INTERPROCEDURAL_OPTIMIZATION TRUE)
+  endif()
+
+  target_link_libraries(search PRIVATE searchcu)
+else()
+  message(STATUS "Skipping CUDA Libraries")
```
DinoBektesevic, line 87: This never prints for me. From the manual it seems as it should. I think I remember reading that Python and C++ don't [deal with their prints very nicely](https://pybind11.readthedocs.io/en/stable/advanced/pycpp/utilities.html) in the Pybind docs. Perhaps something similar could be the cause of the lost message. We could punt that out to `setup.py` - in which case we can basically be sure it would work.

src/kbmod/search/RawImage.h
```diff
@@ -10,6 +10,7 @@
 #ifndef RAWIMAGE_H_
 #define RAWIMAGE_H_
 
+#include <algorithm>
```
DinoBektesevic, line 13: I can't see anything in your changes to RawImage that calls something from this list: https://cplusplus.com/reference/algorithm/

Is this required?

DinoBektesevic (approved):
This breaks tests. We need to protect tests that execute on GPU from doing so then and also we need to rework the CI to test the new build path. Currently, for the sake of saving time we build the per-PR tests in a pre-made environment that is always set up with the GPU in path.

I'm approving this, under condition the build message not printing out is handled before merge, because testing no-GPU build path is not high priority at the moment.

CMakeLists.txt
jeremykubica: If you build with `pip install .`, it doesn't show the output of either the cmake messages or python (setup.py).  You can get them if you do the individual steps (such as running cmake).

src/kbmod/search/RawImage.h
jeremykubica: The code uses std::sort below, which needs algorithm.

I saw this and the `std::isnan` problems in RawImage.cpp when I tried to build on a completely new platform.

PR #341, opened 2023-09-18
Prefactoring: Rename and move interfaces

jeremykubica:
This is an initial set of changes to prepare for the `ImageStack` redesign, including:
- Use a consistent name (`obstime`) throughout the stack for the time of observation.
- Move the storage of the observation time from `LayeredImage` to `RawImage`.
- Change `PPI` to `NPixels` throughout the stack.
- Move the majority of FITS file reading into `RawImage`, adding convenience functions for appending layers, reading in an image from a specific layer, and write a single layer file.
- Write out per-image timestamps (instead of a single one per `LayeredImage`)
- Add more tests for file io.


DinoBektesevic (approved):
It's a bit confused as to what naming convention we're using - but that's job for a different day. 

Feels pretty good to finally sort these annoyances out.

PR #347, opened 2023-09-26
Change pybind11 return policies

jeremykubica:
Change some of the pybind11 return policies to make the returned values mutable (and not do a copy). Clean up a bunch of the code (primarily tests) to use the new behavior.


src/kbmod/run_search.py
```diff
@@ -131,13 +131,13 @@ def do_gpu_search(self, search, img_info, suggested_angle, post_process):
         if self.config["x_pixel_bounds"] and len(self.config["x_pixel_bounds"]) == 2:
             search.set_start_bounds_x(self.config["x_pixel_bounds"][0], self.config["x_pixel_bounds"][1])
         elif self.config["x_pixel_buffer"] and self.config["x_pixel_buffer"] > 0:
-            width = search.get_ImageStack().get_width()
+            width = search.get_imagestack().get_width()
```
DinoBektesevic, line 134: whoops, my bad.

PR #349, opened 2023-09-27
Remove ImageInfo

jeremykubica:
The ImageInfo data structure was created to store a bunch of image related information that we needed for barycentric correction and checking for known objects. Most of this is not needed anymore because we use KOFFI instead. This removes ImageInfo and just returns lists of wcs, mjds instead. This makes the interface wider, but is a step towards cleaning all of this up and removing unused data/code.


src/kbmod/analysis_utils.py
```diff
@@ -54,8 +55,10 @@ def load_images(
         -------
             stack : `kbmod.ImageStack`
                 The stack of images loaded.
-            img_info : `ImageInfo`
-                The information for the images loaded.
+            wcs_list : `astropy.wcs.WCS`
```
DinoBektesevic, line 58: This is a great question I've been struggling with too. WCS is un-plural-izable, but I really dislike the Hungarian notation. No change requested here, but  as a point for the future, I am, however, appinionating that we can just use `wcs` for both and usually the context makes sense of it. There are no image collections or stack searches that do not hold multiple `wcs` and no raw images that can hold more than one `wcs`. I don't know if you've seen examples where we couldn't default to just `wcs` for both cases?

Since this is going to be refactored soon-ish again to put in image/metadata collection, it's fine as it is.

src/kbmod/analysis_utils.py
```diff
@@ -54,8 +55,10 @@ def load_images(
...
+                 A list of `astropy.wcs.WCS` objects for each image.
+            visit_times : list
```
DinoBektesevic, line 60: backtick on list

src/kbmod/analysis_utils.py
```diff
@@ -90,25 +93,40 @@ def load_images(
...
+            visit_id = None
+            with fits.open(full_file_path) as hdu_list:
+                curr_wcs = WCS(hdu_list[1].header)
+
+                # If the visit ID is in header (using Rubin tags), use for the visit ID.
+                # Otherwise extract it from the filename.
+                if "IDNUM" in hdu_list[0].header:
+                    visit_id = str(hdu_list[0].header["IDNUM"])
+                else:
+                    name = full_file_path.rsplit("/")[-1]
```
DinoBektesevic, line 106: if it's a path, better `os.path.split`, it accounts for escapes and os defined separators

src/kbmod/run_search.py
```diff
@@ -382,7 +384,7 @@ def _count_known_matches(self, result_list, search):
             print(matches_string)
         print("-----------------")
 
-    def _calc_barycentric_corr(self, img_info, dist):
+    def _calc_barycentric_corr(self, wcslist, mjds, x_size, y_size, dist):
```
DinoBektesevic, line 387: Can it be named `wcs_list` so it matches previous

src/kbmod/search/stack_search_docs.h
DinoBektesevic, line 1: This entire file exists already in `kbmod/pydocs/stack_search_docs.h. Remove it from `kbmod/search` and update the `kbmod/search/pydocs/` one please.

DinoBektesevic (changes requested):
Primarily, the docs header is in the wrong directory, everything else is fine.

src/kbmod/analysis_utils.py
jeremykubica: Done

src/kbmod/run_search.py
jeremykubica: Done

src/kbmod/search/stack_search_docs.h
jeremykubica: Fixed.

src/kbmod/analysis_utils.py
jeremykubica: Done.

src/kbmod/analysis_utils.py
```diff
@@ -54,8 +55,10 @@ def load_images(
         -------
             stack : `kbmod.ImageStack`
                 The stack of images loaded.
-            img_info : `ImageInfo`
-                The information for the images loaded.
+            wcs_list : `list`
+                 A `list` of `astropy.wcs.WCS` objects for each image.
+            visit_times : list
```
DinoBektesevic, line 60: haha, sorry, I guess I wasn't clear enough. 

```
visit_times : `list`
```

Things that are single back-ticked Sphinx interprets as a reference - in this case to the builtin list type - and things that are double-backticked Sphinx interprets as a literal. So backticking things like `kbmod.ImageStack` craetes a link to the class documentation and things that are double back-ticked are just `grayed_out_literals` like these. So the lists int he explanations don't need to be backticked, the lists in the type description should.

DinoBektesevic (approved):

src/kbmod/analysis_utils.py
jeremykubica: Fixed.

PR #366, opened 2023-10-03
Add input/output options for configuration

jeremykubica:
Rename the `KBMODConfig` to `SearchConfiguration` since this will be the configuration for the core search algorithm (and we will be adding configuration for retrieving images, etc.). Adds the ability to read a `SearchConfiguration` from or write it to either an astropy table or a FITS file (with some admittedly hacky support for `None` and `dict`).  This will allow us to include the core search configuration information as another table in the singular work unit for distributed KBMOD.

Also removes `im_filepath` as a required entry in preparation for reading in the data from alternate sources.

### DIFF
--- a/src/kbmod/configuration.py
+++ b/src/kbmod/configuration.py
@@ -182,7 +182,6 @@ class SearchConfiguration:
         params = {col.name: safe_load(col.value[0]) for col in t.values()}
         return SearchConfiguration.from_dict(params)
 
-
     @classmethod
     def from_yaml(cls, config, strict=True):
         """Load a configuration from a YAML file.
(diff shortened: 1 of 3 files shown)


DinoBektesevic (changes requested):
Annoyingly I went and reviewed the #368 beforehand so I didn't know. Now my review that is more appropriate here is bundled there. 

Basically make the `set_from` factory functions and then build them in  a way where they return objects, or strings, and write to file separately. This gets you two for the cost of one basically.

So my recommendation was

```
to_yaml -> str
to_header -> and then by extension `tofile`, `tostring`
to_file -> with open... + to_header/yaml 

@classmethod
from_yaml -> SearchConfig

@classmethod
from_header  -> SearchConfig

@classmethod
from_file --> SearchConfig
```

And using named headers when creating the FITS files and standardizing on the header name instead of index of users choice. I think I did it wrong in the other PR too and said using HDUs and HDULists, really, unless some parameter is too big or too long and doesn't fit I'd definitely consider writing it just as a header straight up. https://docs.astropy.org/en/stable/io/fits/api/headers.html#astropy.io.fits.Header

Other than that this looks fine to me.

jeremykubica:
I think I got all of your comments from both the PRs, including the use of factory functions and changing how the loading worked.

I also removed the FITS based files. Instead the `SearchConfiguration` object can read or create an HDU. This means we could centralize all the FITS file logic (or other serialization) in the `WorkUnit`.

DinoBektesevic (approved):
I've taken the `__DICT__` and `__NONE__` and replaced it with 

```
serialized_dict = {key: dump(val) for key, val in self._params.items()}
```

so we don't have to worry about casting types by string comparisons. Not much changes in terms of how the HDU is written internally, except that everything is a string now. On the way back we `safe_load` so they are ints/strings/somethings again.

Revert if you don't like it. It's looking good!

PR #368, opened 2023-10-04
Create a framework WorkUnit Class

jeremykubica:
This PR is an example of how we could structure the input reading and writing of a KBMOD `WorkUnit`. It builds off PR #366.

### DIFF
--- a/src/kbmod/work_unit.py
+++ b/src/kbmod/work_unit.py
@@ -5,13 +5,15 @@ from astropy.table import Table
 import numpy as np
 from pathlib import Path
 
-import kbmod.search as kb
+from kbmod.configuration import SearchConfiguration
+from kbmod.search import ImageStack, LayeredImage, PSF, RawImage
+
 
 class WorkUnit:
     """The work unit is a storage and I/O class for all of the data
     needed for a full run of KBMOD, including the: the search parameters,
     data files, and the data provenance metadata.
-    
+
     A WorkUnit file is a FITS file with the following extensions:
         0 - Primary header with overall metadata
         1 - The data provenance metadata
@@ -24,53 +26,15 @@ class WorkUnit:
         self.im_stack = im_stack
         self.config = config
 
-    @staticmethod
-    def _read_layered_image(hdul, img_num):
-        """Read in a single layered image from an open WorkUnit file.
-        
-        Reads from 4 separate extensions in the order:
-        Science, Variance, Mask, and PSF.
-        
-        Parameters
-        ----------
-        hdul : `astropy.io.fits.hdu.hdulist.HDUList`
-            The pointer to the FITS file data.
-        img_num : `int`
-            The image number to extract.
-
-        Returns
-        -------
-        result : `layered_image`
-            A constructed layered image.
-        """
-        ext_num = 3 + 4 * i
-
-        # Read in science and variance layers.
-        sci = hdu_to_raw_image(hdul[ext_num])
-        var = hdu_to_raw_image(hdul[ext_num + 1])
-        
-        # Read the mask layer if it exists.
-        msk = hdu_to_raw_image(hdul[ext_num + 2])
-        if mask is None:
-            msk = kb.RawImage(np.zeros((sci.get_height(), sci.get_width())))
-
-        # Check if the PSF layer exists.
-        if hdul[ext_num + 3].header["NAXIS"] == 2:
-            p = kb.PSF(hdul[ext_num + 3].data)
-        else:
-            p = kb.PSF(1e-8)
-
-        return LayeredImage(sci, var, msk, p)
-        
     @classmethod
     def from_file(cls, filename):
         """Create a WorkUnit from a single file.
-        
+
         Parameters
         ----------
         filename : `str`
             The file to load.
-        
+
         Returns
         -------
         result : `WorkUnit`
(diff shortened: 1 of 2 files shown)


src/kbmod/work_unit.py
```diff
@@ -0,0 +1,170 @@
...
+        with fits.open(filename) as hdul:
+            num_layers = len(hdul)
+            if num_layers < 5:
+                raise ValueError(f"WorkUnit file has too few extensions {len(hdul)}.")
+
+            # TODO - Read in provenance metadata from extension #1.
+
+            # Read in the search parameters from extension #2.
+            config = SearchConfiguration()
+            config_table = Table(hdul[2].data)
```
DinoBektesevic, line 56: I was kind of wondering earlier why you wouldn't just write the config as a header, is there a variable we have that breaks the length limit or something?

src/kbmod/work_unit.py
```diff
@@ -0,0 +1,170 @@
...
+                    p = PSF(hdul[ext_num + 3].data)
+                else:
+                    p = PSF(1e-8)
+
+                imgs.append(LayeredImage(sci, var, msk, p))
+
+            im_stack = ImageStack(imgs)
+
+            result = WorkUnit(im_stack=im_stack, config=config)
+        return result
```
DinoBektesevic, line 92: `return WorkUnit(im_stack=im_stack, config=config)`?

src/kbmod/work_unit.py
```diff
@@ -0,0 +1,170 @@
...
+            if len(hdul) != 4 * num_images + 3:
+                raise ValueError(
+                    f"WorkUnit wrong number of extensions. Expected "
+                    f"{4 * num_images + 3}. Found {len(hdul)}."
+                )
+
+            # Read in all the image files.
+            imgs = []
+            for i in range(num_images):
+                ext_num = 3 + 4 * i
```
DinoBektesevic, line 70: This convinces me we should just use named HDUs.

src/kbmod/work_unit.py
```diff
@@ -0,0 +1,170 @@
...
+                # Read the mask layer if it exists.
+                msk = hdu_to_raw_image(hdul[ext_num + 2])
+                if msk is None:
+                    msk = RawImage(np.zeros((sci.get_height(), sci.get_width())))
+
+                # Check if the PSF layer exists.
+                if hdul[ext_num + 3].header["NAXIS"] == 2:
+                    p = PSF(hdul[ext_num + 3].data)
+                else:
+                    p = PSF(1e-8)
```
DinoBektesevic, line 85: We are the sole creator of these files. I think we should just throw an error if we don't have everything we need before writing it to begin with. Or make these defaults the matter of RawImage/PSF constructors or something and write whatever they make to disk.

src/kbmod/work_unit.py
```diff
@@ -0,0 +1,170 @@
...
+    img : `RawImage`
+        The RawImage to convert.
+
+    Returns
+    -------
+    hdu : `astropy.io.fits.hdu.image.ImageHDU`
+        The image extension.
+    """
+    # Expensive copy. To be removed with RawImage refactor.
+    np_pixels = np.array(img.get_all_pixels()).astype("float32", casting="same_kind")
```
DinoBektesevic, line 144: img is `RawImage` no, a list of `float32` already. Why not just instantiate `np.array(img.get_all_pixels(), dtype=np.single)` and avoid a cast to double's and then to floats altogether.

src/kbmod/work_unit.py
```diff
@@ -0,0 +1,170 @@
...
+    hdu : `astropy.io.fits.hdu.image.ImageHDU`
+        The image extension.
+
+    Returns
+    -------
+    img : `RawImage` or None
+        The RawImage if there is valid data and None otherwise.
+    """
+    img = None
+    if hdu.header["NAXIS"] == 2:
```
DinoBektesevic, line 165: I think it's better to be checking `if isinstance(hdu, (ImageHDU, CompImageHDU)` because those are the two image HDU types.

DinoBektesevic (changes requested):
Using named header units would help a lot so I'd most recommend them.
I would restructure how SearchConfiguration (de)serializes itself, as in naming not in scope, so that we can create on-the-fly strings. 

I am very tepid about pickling anything, I've usually nothing but slight annoyance with it. I would remove support for that here, short of if there's anything we strictly need - in which case I would make it an internal job to write that to string in a way that it roundtrips. 

You expressed some concern about being on the right track here, but yes, this is basically exactly what we talked about regarding building serializable processable units.

src/kbmod/configuration.py
```diff
@@ -137,6 +141,62 @@ def set_from_dict(self, d, strict=True):
...
+        -------
+        t: `~astropy.table.Table`
+            The configuration table.
+        """
+        t = Table()
+        for col in self._params.keys():
+            val = self._params[col]
+            t[col] = [val]
+
+            # If Table does not understand the type, pickle it.
```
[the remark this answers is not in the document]
jeremykubica: That's fair. The only thing this is supporting is dictionaries and None. I originally had the code write these out as strings, but thought using pickle might reduce the code complexity. I'm happy to change it back.

src/kbmod/configuration.py
```diff
@@ -170,15 +230,43 @@ def load_from_file(self, filename, strict=True):
...
 
-    def save_configuration(self, filename, overwrite=False):
+    def load_from_fits_file(self, filename, layer=0, strict=True):
+        """Load a configuration from a FITS extension file.
+
+        Parameters
+        ----------
+        filename : `str`
+            The filename, including path, of the configuration file.
+        layer : `int`
```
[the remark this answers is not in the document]
jeremykubica: Nice. I wasn't aware that FITS supported named headers.

src/kbmod/configuration.py
```diff
@@ -137,6 +141,62 @@ def set_from_dict(self, d, strict=True):
         for key, value in d.items():
             self.set(key, value, strict)
 
+    def set_from_table(self, t, strict=True):
```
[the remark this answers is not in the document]
jeremykubica: Sounds good.

jeremykubica:
I think I got all of your requested changes. For not FITS is the only output format, but we can add `to_` and `from_` functions for other formats.

src/kbmod/work_unit.py
```diff
@@ -0,0 +1,182 @@
...
+        return WorkUnit(im_stack=im_stack, config=config)
+
+    def to_fits(self, filename, overwrite=False):
+        """Write the WorkUnit to a single FITS file.
+
+        Uses the following extensions:
+            0 - Primary header with overall metadata
+            1 or "metadata" - The data provenance metadata
+            2 or "kbmod_config" - The search parameters
+            3+ - Image extensions for the science layer ("SCI_i"),
```
DinoBektesevic, line 88: Remember that the docstrings are markdown rendered into documentation by Sphinx. To render a proper list that discriminates between references and literals try:

```
The FITS file will have at least the following extensions:

0. ``PRIMARY`` extension
1. ``METADATA`` extension containing provenance 
2. ``KBMOD_CONFIG`` extension containing search parameters
3. (+) any additional image extensions are named ``SCI_i``, ``VAR_i``, ``MSK_i`` 
and ``PSF_i`` for the science, variance, mask and PSF of each image respectively, 
where ``i`` runs from 0 to number of images in the `WorkUnit`.
```

DinoBektesevic (approved):
Looks great!

PR #371, opened 2023-10-10
Analysis utils refactor

### LINKED ISSUE #357: Break apart analysis_utils.py
`Analysis_utils.py` includes a variety of different functionality including data loading, some filtering, reading the results in batches, and clustering similar results. This should be refactored into coherently themed modules (e.g. see #355) for better maintainability.

maxwest-uw:
resolves #357 
- moves the image loading code into a new `data_interface` module
- moves the tests for the interface into a new file


src/kbmod/data_interface.py
```diff
@@ -0,0 +1,150 @@
...
+
+import kbmod.search as kb
+
+from .file_utils import *
+from .filters.clustering_filters import DBSCANFilter
+from .filters.stats_filters import *
+from .result_list import ResultList, ResultRow
+
+
+class Interface:
```
jeremykubica, line 18: Long term question (no changed needed in this PR): Do we need `Interface` to be a class? Or should it be a few static methods?

src/kbmod/data_interface.py
```diff
@@ -0,0 +1,150 @@
+import multiprocessing as mp
```
jeremykubica, line 1: I think you can clean out a bunch of the imports that are not used for the file loading, including; multiprocessing, result_list, stats_filter, clustering_filters.

jeremykubica (approved):

src/kbmod/data_interface.py
maxwest-uw: I agree, it would probably be better to just have them be static!

PR #376, opened 2023-10-16
Add basic masking benchmarks

jeremykubica:
Add manual benchmarks for the various masking functions.

### DIFF
--- a/benchmarks/bench_masking.py
+++ b/benchmarks/bench_masking.py
@@ -1,0 +1,150 @@
+import timeit
+import numpy as np
+
+from kbmod.masking import (
+    BitVectorMasker,
+    DictionaryMasker,
+    GlobalDictionaryMasker,
+    GrowMask,
+    ThresholdMask,
+    apply_mask_operations,
+)
+from kbmod.result_list import ResultRow
+from kbmod.search import ImageStack, PSF, RawImage, LayeredImage
+
+
+def set_up_image_stack():
+    """Create a fake image stack to use with the masking tests.
+
+    Returns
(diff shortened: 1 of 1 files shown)


benchmarks/bench_masking.py
```diff
@@ -0,0 +1,150 @@
...
+
+        res_times.append(timeit.timeit("stack.grow_mask(r)", number=1, globals=locals()))
+    return np.mean(np.array(res_times))
+
+
+def run_all_benchmarks():
+    print("Apply Mask Timings:")
+    print("    Method       |    Time")
+    print("-" * 30)
+    print(f" Python Dict     | {bench_dictionary_masker():10.7f}")
```
wilsonbb, line 136: Nit: I would print out the unit (seconds) as well.

benchmarks/bench_masking.py
```diff
@@ -0,0 +1,150 @@
...
+    for i in range(10):
+        stack = set_up_image_stack()
+        premasker = BitVectorMasker(1, [])
+        premasker.apply_mask(stack)
+
+        res_times.append(timeit.timeit("stack.grow_mask(r)", number=1, globals=locals()))
+    return np.mean(np.array(res_times))
+
+
+def run_all_benchmarks():
```
wilsonbb, line 132: Nit: since it's a manual tool run from the command line, it may be also helpful to print out the number of repetitions/trials so the user doesn't have to inspect the file.

wilsonbb (approved):
Looks good to me! Left two minor nits.

PR #381, opened 2023-10-19
A few small fixes to the stamp generation code

jeremykubica:
Pass the `ImageStack` by reference for efficiency.  Also applies clag-format to the files.


wilsonbb (approved):
Thanks for the catch!

PR #382, opened 2023-10-19
DRAFT PR for discussion

jeremykubica:
Makes several fundamental changes to the KBMOD core search:
- Shifts all predictions to be 0 centered instead of 0.5 centered for a pixel.
- Uses `floor` instead of `int` to convert from predicted position to pixel index (matters for pixels off the negative side of the grid)
- Removes pixel interpolation functions (add and get)
- Makes final stamps non-interpolated.


src/kbmod/search/kernels.cu
```diff
@@ -147,8 +147,8 @@ __global__ void searchFilterImages(int num_images, int width, int height, void *
         for (int i = 0; i < num_images; ++i) {
             // Predict the trajectory's position.
             float curr_time = image_data.image_times[i];
-            int current_x = x + int(curr_trj.vx * curr_time + 0.5);
-            int current_y = y + int(curr_trj.vy * curr_time + 0.5);
+            int current_x = x + (int)(floor(curr_trj.vx * curr_time));
+            int current_y = y + (int)(floor(curr_trj.vy * curr_time));
```
wilsonbb, line 151: Minor question: From the earlier discussion, it seems the "+ 0.5" is used so that the int casting will round the result rather than truncate/floor it.

We should definitely be consistent about whichever we do, but out of curiosity is there a reason to prefer flooring over rounding?

wilsonbb: After writing that, I'm realizing that I shouldn't have used truncate/floor interchangeably since they are only equivalent operations for positive values.

jeremykubica:
Closing this for now since the Eigen change went in. We can revisit pieces of this as needed.

PR #387, opened 2023-10-23
Move more stamp functions to StampCreator

jeremykubica:
Moves the remaining stamp-related functions from `StackSearch` to `StampCreator`.

### DIFF
--- a/src/kbmod/search/pydocs/stamp_creator_docs.h
+++ b/src/kbmod/search/pydocs/stamp_creator_docs.h
@@ -120,6 +120,14 @@ namespace pydocs {
       The co-added stamp.
   )doc";
 
+  static const auto DOC_StampCreator_get_coadded_stamps = R"doc(
+  todo
+  )doc";
+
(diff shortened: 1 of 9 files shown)


src/kbmod/search/pydocs/stamp_creator_docs.h
```diff
@@ -120,6 +120,14 @@ namespace pydocs {
...
 
+  static const auto DOC_StampCreator_get_coadded_stamps = R"doc(
+  todo
+  )doc";
+
+  static const auto DOC_StampCreator_filter_stamp = R"doc(
+  todo
+  )doc";
+
 } /* pydocs */
```
wilsonbb, line 131: Optional: Up to you, but it's probably nice to fill these out while we're here

wilsonbb (approved):
Thanks for moving these!

src/kbmod/search/pydocs/stamp_creator_docs.h
jeremykubica: Good call. I added them.

PR #389, opened 2023-11-03
A series of small cleanups

jeremykubica:
- Fix a couple breakages that started happening when I compiled on a fresh environment.
- Fix the regression test by increasing the flux. I'm not sure why the change to Eigen caused a small difference in `add_fake_object`, but it looks like the `interpolated_add` is spreading out the flux more than the previous implementation. This compensates without changing any of the code.
- Add a test for `interpolated_add`.
- Remove the exceptions list from `apply_map`. This was never used (except in a single test).

### DIFF
--- a/src/kbmod/search/stamp_creator.cpp
+++ b/src/kbmod/search/stamp_creator.cpp
@@ -190,8 +190,8 @@ std::vector<RawImage> StampCreator::get_coadded_stamps_gpu(ImageStack& stack,
         // should not accept images of different sizes being added to the stack
         // and then get rid of all these for loops in the code
         auto& sci = stack.get_single_image(t).get_science().get_image();
-        assertm(sci.get_width() == width, "Stack image " + t + " has different width than 0th image.");
-        assertm(sci.get_height() == height, "Stack image " + t + " has different width than 0th image.");
+        assertm(sci.cols() == width, "Stack image has different width than 0th image.");
+        assertm(sci.rows() == height, "Stack image has different width than 0th image.");
         data_refs[t] = sci.data();
     }
 
--- a/src/kbmod/masking.py
+++ b/src/kbmod/masking.py
@@ -63,14 +63,11 @@ class BitVectorMasker(ImageMasker):
     ----------
     flags : `int`
         A bit vector of masking flags to apply.
-    exception_list : `list`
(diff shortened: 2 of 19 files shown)


src/kbmod/search/stamp_creator.cpp
```diff
@@ -190,8 +190,8 @@ std::vector<RawImage> StampCreator::get_coadded_stamps_gpu(ImageStack& stack,
         // should not accept images of different sizes being added to the stack
         // and then get rid of all these for loops in the code
         auto& sci = stack.get_single_image(t).get_science().get_image();
-        assertm(sci.get_width() == width, "Stack image " + t + " has different width than 0th image.");
-        assertm(sci.get_height() == height, "Stack image " + t + " has different width than 0th image.");
+        assertm(sci.cols() == width, "Stack image has different width than 0th image.");
```
DinoBektesevic, line 193: Above you replaces `cols` and `rows` with `get_width` and `get_height` but now you're doing the opposite. Is there a particular logic you're applying? Asking so I can do the same.

DinoBektesevic (approved):
You're mentioning `interpolated_add` spreading the given value more than previously, but it's not immediately obvious from the added test how big is this problem?

Love getting rid of the `exception` flags. Cleans up so many things.

src/kbmod/search/stamp_creator.cpp
jeremykubica: `sci` is an Eigen image (from the `get_image()`) and the statement was giving me a compilation error that the eigen image didn't have `get_height()` and `get_width()`.

Above it was the opposite. I was getting a compile error that RawImage did not have `cols()` or `rows()`.

jeremykubica:
In terms of interpolate_add spreading out the values differently, it was not a huge difference. In checking the images after adding the object the difference in flux values was ~10%. It was just enough for filtering to drop some of them. I'm guessing some of this has to do with the 0.5 centering vs not being inconsistent in the old code.

src/kbmod/search/stamp_creator.cpp
jeremykubica: These two compile errors were the ones where it was compiling fine for me all morning and then I refreshed something and it started giving me errors on unchanged lines.

PR #390, opened 2023-11-03
Remove interpolation from stamp generation

jeremykubica:
The only time interpolation was used in stamp generation is when creating the stamps for visualization. This changes the behavior to return the actual pixel values of the stamp (instead of interpolated), so the user is looking at a cutout of the image.

This PR also adds a bunch of stamp tests, which uncovered a small bug with the new implementation of `create_stamps` (only overwriting `NO_DATA` values on the margin). This fixes that behavior as well.


wilsonbb (approved):
Looks good!

PR #393, opened 2023-11-09
A draft proposal of a simple pointings search

jeremykubica:
Create a PointingTable structure and allow us to ask queries against it (in terms of heliocentric points). This is an initial version and not feature complete. In particular it is not optimized, only supports CSV inputs, and does not include all the corrections we might want to use.


jeremykubica:
Since this is an example, I moved it out of KBMOD to its own repository

PR #397, opened 2023-12-01
Add a series of functionality to the ResultList object

jeremykubica:
This PR adds a bunch of helper methods to the `ResultRow` and `ResultList` data structures to make analysis easier, including:
- Adds the ability to compute and save predicted (RA, dec) positions given a list of times and a WCS.
- Adds ability to revert a filtering step, re-adding the previously filtered results to the `ResultList`.
- Adds the ability to sort the result list by properties of the `ResultRow`

In addition includes a few clean ups:
- Adds a lot of comments about the design decisions behind the current implementation.
- Makes some of the attributes of `ResultRow` private and adds read-only properties. This is done to prevent users from accidentally causing an inconsistent state.
- Remove now unused `zip_phi_psi_idx`
- Expands the tests.

### DIFF
--- a/tests/test_stats_filters.py
+++ b/tests/test_stats_filters.py
@@ -3,6 +3,7 @@ import unittest
 from kbmod.filters.stats_filters import *
 from kbmod.result_list import *
 from kbmod.search import *
+from kbmod.trajectory_utils import make_trajectory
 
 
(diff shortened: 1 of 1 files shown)


wilsonbb (approved):
LGTM!

PR #403, opened 2023-12-11
Redesign how KBMOD encodes data for on GPU use

jeremykubica:
This PR moves the encoding of the Psi and Phi arrays into an object to provide more flexibility and better modularity. The major goals of this change include:
- Restore the interleaved psi/phi approach from the original code to improve memory locality on the GPU.
- Allow the code to copy the psi/phi data to GPU memory once and run multiple searches. This is not currently used anywhere, but will allow us to reduce the memory needed to store results by breaking the search into batches.
- Provide direct hooks for reading the encoding that can be used for testing (previous code was tested indirectly via a full search)

The code is a bit complicated due to the fact that CUDA cannot call non-device functions from a device function (including the getters/setters for a class).

### DIFF
--- a/src/kbmod/search/kernels.cu
+++ b/src/kbmod/search/kernels.cu
@@ -19,10 +19,35 @@
 
 #include "common.h"
 #include "cuda_errors.h"
-
-//#include "image_stack.h"
+#include "psi_phi_array_ds.h"
 
 namespace search {
+
+extern "C" void device_allocate_psi_phi_array(PsiPhiArray& data) {
+    assertm(data.get_cpu_array_ptr() != nullptr, "CPU array not allocated.");
+    assertm(data.get_gpu_array_ptr() == nullptr, "GPU array already allocated.");
+
+    void* array_ptr;
+    checkCudaErrors(cudaMalloc((void **)&array_ptr, data.get_total_array_size()));
+    checkCudaErrors(cudaMemcpy(array_ptr,
+                               data.get_cpu_array_ptr(),
+                               data.get_total_array_size(),
+                               cudaMemcpyHostToDevice));
+    data.set_gpu_array_ptr(array_ptr);
+}
+
+extern "C" void device_free_psi_phi_array(PsiPhiArray& data) {
+    if (data.get_gpu_array_ptr() != nullptr) {
+        checkCudaErrors(cudaFree(data.get_gpu_array_ptr()));
+        data.set_gpu_array_ptr(nullptr);
+    }
+}
+
+// Read the pixel from the compressed array.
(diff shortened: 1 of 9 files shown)


src/kbmod/search/kernels.cu
```diff
@@ -19,11 +19,62 @@
...
+        throw std::runtime_error("GPU data is already allocated.");
+
+    void* device_array_ptr;
+    checkCudaErrors(cudaMalloc((void **)&device_array_ptr, data->get_total_array_size()));
+    checkCudaErrors(cudaMemcpy(device_array_ptr,
+                               data->get_cpu_array_ptr(),
+                               data->get_total_array_size(),
+                               cudaMemcpyHostToDevice));
+    data->set_gpu_array_ptr(device_array_ptr);
+}
```
maxwest-uw, line 39: this is the function that allocated and sets the array on the gpu device memory, and we enforce in the search that the `gpu_array_ptr` has been set, but it's not actually been used/read in at the moment right? so during runtime we just allocate the array memory, then pass the data in again for the process? (this is fine for the initial changes and if we'll actually read the pre-loaded array later I just wanted to make sure I understand)

maxwest-uw (commented):
have a few comments to make sure I understand how the kernel changes work, but otherwise looks good!

src/kbmod/search/kernels.cu
```diff
@@ -19,11 +19,62 @@
...
+
+extern "C" void device_free_psi_phi_array(PsiPhiArray* data) {
+    if (data->gpu_array_allocated()) {
+        checkCudaErrors(cudaFree(data->get_gpu_array_ptr()));
+        data->set_gpu_array_ptr(nullptr);
+    }
+}
+
+__forceinline__ __device__ PsiPhi read_encoded_psi_phi(PsiPhiArrayMeta &params, void *psi_phi_vect, int time,
```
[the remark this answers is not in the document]
jeremykubica: Yes. I broke this out for two reasons:
1) It makes the code a little easier to read.
2) It keeps all the decoding logic in a single place in case we want to change it later.
The `__forceinline__` inlines this code in the function, so we don't pay the overhead of a function call.

src/kbmod/search/kernels.cu
jeremykubica: It is used in the search. The `PsiPhiArray` is passed into the search function `deviceSearchFilter` and the pointer to the GPU data is extracted in line 315:

```
searchFilterImages<<<blocks, threads>>>(psi_phi_array.get_meta_data(), psi_phi_array.get_gpu_array_ptr(), device_img_times,
                                            params, num_trajectories, device_tests, device_search_results);
```

PR #404, opened 2023-12-13
Add WCS information to the WorkUnit

jeremykubica:
Adds two forms of WCS information to the `WorkUnit` class. A global `wcs` attribute that should be set when all the images are projected into the same pixel space and a `per_image_wcs` for images that live in different pixel spaces. This information is automatically stored in and read from the FITS headers.

### DIFF
--- a/src/kbmod/work_unit.py
+++ b/src/kbmod/work_unit.py
@@ -60,7 +60,7 @@ class WorkUnit:
             config = SearchConfiguration.from_hdu(hdul["kbmod_config"])
 
             # Read in the global WCS from extension 0 if the information exists.
-            self.wcs = extract_wcs(hdul[0])
+            global_wcs = extract_wcs(hdul[0])
 
             # Read the size and order information from the primary header.
             num_images = hdul[0].header["NUMIMG"]
@@ -71,9 +71,10 @@ class WorkUnit:
                 )
 
             # Read in all the image files.
+            per_image_wcs = []
             for i in range(num_images):
                 # Extract the per-image WCS if one exists.
-                self.per_image_wcs.append(extract_wcs(hdul[f"SCI_{i}"]))
+                per_image_wcs.append(extract_wcs(hdul[f"SCI_{i}"]))
 
                 # Read in science, variance, and mask layers.
                 sci = hdu_to_raw_image(hdul[f"SCI_{i}"])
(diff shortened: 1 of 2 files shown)


tests/test_work_unit.py
```diff
@@ -40,11 +41,37 @@ def setUp(self):
         self.config.set("mask_bits_dict", {"A": 1, "B": 2})
         self.config.set("repeated_flag_keys", None)
 
+        # Creater a fake WCS
```
delucchi-cmu, line 44:
```suggestion
        # Create a fake WCS
```

tests/test_work_unit.py
```diff
@@ -93,6 +121,9 @@ def test_save_and_load_fits(self):
                     for x in range(p1.get_dim()):
                         self.assertAlmostEqual(p1.get_value(y, x), p2.get_value(y, x))
 
+                # No per-image WCS
```
wilsonbb, line 124: Can we test generating a per-image WCS as well?

src/kbmod/work_unit.py
```diff
@@ -77,7 +93,9 @@ def from_fits(cls, filename):
                 imgs.append(LayeredImage(sci, var, msk, p))
 
         im_stack = ImageStack(imgs)
-        return WorkUnit(im_stack=im_stack, config=config)
+        result = WorkUnit(im_stack=im_stack, config=config, wcs=global_wcs)
+        result.per_image_wcs = per_image_wcs
```
wilsonbb, line 97: I probably would have set `self.per_image_wcs` via an optional parameter to the `WorkUnit`'s constructor.

This seems fine but curious if there was a specific reason we chose not to do that

tests/test_work_unit.py
jeremykubica: Good call. I added a mix of WCS and non-WCS to test the parsing functions.

src/kbmod/work_unit.py
jeremykubica: I changed it to use an optional parameter.

Originally the motivation is the most `WorkUnit`s should use the single wcs. The `per_image_wcs` is only for ones that have not been reprojected yet. But I think it is clearer the way you proposed.

wilsonbb (approved):
LGTM!

PR #423, opened 2024-01-05
Add Orbit-based fake data to a WorkUnit

jeremykubica:
Create helper functions for generating ephemerides of an object given its orbital parameters and then adding them into a `WorkUnit`.  The function automatically tracks which images were able to get a fake detection (inside the image bounds and not masked) so that we can determine whether this is a useful fake result or we should use a different orbit.


maxwest-uw (approved):
sorry for the delay, lgtm !

Issue #352, opened 2023-09-28
Consolidate Masking Logic


jeremykubica:
The logic for computing masks is spread over multiple C++ and python files including `raw_image.cpp`, `layered_image.cpp`, `image_stack.cpp`, and `masking.py`. As we look to move much of the masking out of the core KBMOD search (and make it a preprocessing step), it would be cleaner to consolidate this logic in a single location outside of (at least) the image data structures.

One example implementation would be to create a masking class in C++ (modeled off the python class) or extend the current python class that takes an `ImageStack` and performs masking on all images in it. This class could be overloaded to perform the different types of masking (threshold masking, bitmask masking, etc.). Note the dictionary -> bitmask code should stay in python.

jeremykubica:
Closed by #424

PR #424, opened 2024-01-11
Clean up masking logic

### LINKED ISSUE #415: Support BitMask Masks
The core KBMOD search should be redesigned to take masks as bitmaps and apply them during stages like create Psi and Phi. The logic for generating the bit mask from the full mask (a bitvector of reasons), including growing the masks, should be moved to preprocessing (standardizer).

The core search logic will need to be updated to handle inserting NO_DATA into psi curves, phi curves, light curves, and stamps. This could be done by applying the bit mask once on load or during each action.

jeremykubica:
This PR refactors the masking code to clean up the interfaces, but does not change end-to-end behavior. A major part of this change is that most masking operations will now work on the `LayeredImage`'s mask layer. For example, `grow_mask` will now expand the masked pixels in the mask layer instead of the science layer.  `LayeredImage`'s `apply_mask()` function will then apply the masks to the science and variance layers.

The motivations for these changes are:
- Support binary masks coming via the standardizers and `WorkUnit`s. These will be cases where all the masking logic is applied when the `WorkUnit` is constructed.
- Maintain backwards compatibility with current input files.
- Consolidate most masking logic in `LayeredImage` instead of across `RawImage`, `LayeredImage`, and `ImageStack`.
- Reduce logic in `RawImage` for possible eventual removal of that class.

More detail on the changes:

`LayeredImage`: Move the majority of the masking logic into `LayeredImage` and perform most changes on the mask layer (instead of directly onto science and variance).

`RawImage`: Remove `grow_mask` function. All mask changes should be done through `LayeredImage` and then applied to `RawImage` with `apply_mask()`.

`ImageStack`:
- Remove masking functions that just loop over the individual `LayeredImage`s, such as `apply_mask_flags()`.
- Remove global mask as an object attribute of `ImageStack`. Replace it with a function to create a new global mask on-demand. This allows the code to skip allocating a global mask in cases where no global masking is done.

`masking.py`:
- Remove the class-based logic. This was added to support use cases that will be covered in the standardizer. We only need to apply the individual functions from `LayeredImage` and `ImageStack`. The new flow is much simpler.

`run_search.py`:
- Move the masking logic into `masking.py`.

Closes #415 and #352.

### DIFF
--- a/src/kbmod/masking.py
+++ b/src/kbmod/masking.py
@@ -3,6 +3,7 @@
 
 import kbmod.search as kb
 
+
 def mask_flags_from_dict(mask_bits_dict, flag_keys):
     """Generate a bitmask integer of flag keys from a dictionary
     of masking reasons and a list of reasons to use.
@@ -25,7 +26,7 @@ def mask_flags_from_dict(mask_bits_dict, flag_keys):
         bitmask += 2 ** mask_bits_dict[bit]
     return bitmask
 
-            
+
 def apply_mask_operations(config, stack):
     """Perform all the masking operations based on the search's configuration parameters.
 
@@ -79,6 +80,5 @@ def apply_mask_operations(config, stack):
     # Apply the masks to the images.
     for i in range(stack.img_count()):
         stack.get_single_image(i).apply_mask(0xFFFFFF)
-            
-    return stack
 
+    return stack
--- a/src/kbmod/masking.py
+++ b/src/kbmod/masking.py
@@ -1,95 +1,11 @@
-"""Classes for performing masking on images from FITS files.
-
-ImageMasker provides an abstract base class that can be overridden to define masking
-algorithms for specific studies, instruments, or FITS headers. Specific masking classes
-are provided to support common masking operations including: masking based on a bit vector,
-masking based on a dictionary, masking based on a threshold, and growing a current mask.
+"""Functions for performing masking operations as specified in the configuration.
 """
-
-import abc
 
 import kbmod.search as kb
 
-
-def apply_mask_operations(stack, mask_list):
-    """Apply a series of masking operations defined by a list of
-    ImageMasker objects.
-
-    Parameters
-    ----------
-    stack : `kbmod.ImageStack`
-        The stack before the masks have been applied.
-    mask_list : `list`
-        A list of mask_list objects.
-
-    Returns
-    -------
-    stack : `kbmod.ImageStack`
-        The same stack object to allow chaining.
-    """
-    for mask in mask_list:
-        stack = mask.apply_mask(stack)
-    return stack
-
-
-class ImageMasker(abc.ABC):
-    """The base class for masking operations."""
-
-    def __init__(self, *args, **kwargs):
-        pass
-
-    @abc.abstractmethod
-    def apply_mask(self, stack):
-        """Apply the mask to an image stack.
-
-        Parameters
-        ----------
-        stack : `kbmod.ImageStack`
-            The stack before the masks have been applied.
-
-        Returns
-        -------
-        stack : `kbmod.ImageStack`
-            The same stack object to allow chaining.
-        """
-        pass
-
-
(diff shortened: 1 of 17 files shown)


src/kbmod/masking.py
```diff
@@ -98,126 +15,70 @@ class DictionaryMasker(BitVectorMasker):
...
+            stack.get_single_image(i).union_masks(global_binary_mask)
+
+    # Grow the masks.
+    if config["mask_grow"] and config["mask_grow"] > 0:
+        for i in range(stack.img_count()):
+            stack.get_single_image(i).grow_mask(config["mask_grow"])
+
+    # Apply the masks to the images.
+    for i in range(stack.img_count()):
+        stack.get_single_image(i).apply_mask(0xFFFFFF)
```
wilsonbb, line 82: Nit: can you add a comment here to note that 0xFFFFFF is being used to apply all flags?

src/kbmod/search/layered_image.h
```diff
@@ -76,7 +93,7 @@ class LayeredImage {
     PSF psf;
     RawImage science;
     RawImage mask;
-    RawImage variance;
+    RawImage variance;
```
wilsonbb, line 96: Nit: remove trailing whitespace

src/kbmod/search/pydocs/image_stack_docs.h
```diff
@@ -32,57 +32,25 @@ static const auto DOC_ImageStack_build_zeroed_times = R"doc(
   in the stack and the first image.
   ")doc";
 
-static const auto DOC_ImageStack_apply_mask_flags = R"doc(
-  Applies a mask to each image by comparing the given bit vector with the
-  values in the mask layer and marking pixels NO_DATA. Modifies the image in-place.
+static const auto DOC_ImageStack_make_global_mask = R"doc(
+  Create a new global mask that from a set of flags and a threshold.
```
wilsonbb, line 36: Nit: s/global mask that from/global mask from

tests/test_layered_image.py
```diff
@@ -126,36 +145,78 @@ def test_overwrite_PSF(self):
...
         self.assertLess(science_pixel_psf1, science_pixel_psf2)
 
-    def test_mask_threshold(self):
+    def test_binarize_mask(self):
+        # Mask out a range of pixels.
+        mask = self.image.get_mask()
+        for x in range(9):
+            mask.set_pixel(10, x, x)
+
+        # Only keep the mask at for pixels with flags at
```
wilsonbb, line 154: Nit: s/mask at for/mask for

tests/test_layered_image.py
```diff
@@ -126,36 +145,78 @@ def test_overwrite_PSF(self):
...
                 if value > threshold:
                     index = self.image.get_width() * y + x
                     masked_pixels[index] = True
-
-        # Do the masking and confirm we have masked
-        # at least 1 pixel.
-        self.image.apply_mask_threshold(threshold)
         self.assertGreater(len(masked_pixels), 0)
 
+        # Reset the mask an perform threshold masking.
```
wilsonbb, line 207: Nit: s/mask an perform/mask and perform

wilsonbb (approved):
LGTM, just a few nits

src/kbmod/masking.py
jeremykubica: Added

src/kbmod/search/layered_image.h
jeremykubica: Removed

src/kbmod/search/pydocs/image_stack_docs.h
jeremykubica: Fixed

tests/test_layered_image.py
jeremykubica: Fixed

tests/test_layered_image.py
jeremykubica: Fixed

PR #432, opened 2024-01-12
Consolidate fake WCS generation logic

jeremykubica:
Clean up the testing code by consolidating the specification of fake WCSs into a single location and calling that.  Also expands some of the tests for WCS generation to make sure they are reasonable as fake data.

### DIFF
--- a/tests/test_wcs_utils.py
+++ b/tests/test_wcs_utils.py
@@ -51,18 +51,18 @@ class test_wcs_conversion(unittest.TestCase):
             else:
                 wcs_info = self.wcs
 
-        # Run the dictionary and full WCS tests as separate subtests
-        with self.subTest(i=use_dictionary):
-            pri = fits.PrimaryHDU()
-            self.assertFalse("CRVAL1" in pri.header)
-            self.assertFalse("CRVAL2" in pri.header)
-            self.assertFalse("CRPIX1" in pri.header)
-            self.assertFalse("CRPIX2" in pri.header)
+            # Run the dictionary and full WCS tests as separate subtests
+            with self.subTest(i=use_dictionary):
+                pri = fits.PrimaryHDU()
+                self.assertFalse("CRVAL1" in pri.header)
+                self.assertFalse("CRVAL2" in pri.header)
+                self.assertFalse("CRPIX1" in pri.header)
+                self.assertFalse("CRPIX2" in pri.header)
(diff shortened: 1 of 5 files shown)


tests/test_wcs_utils.py
```diff
@@ -45,16 +45,54 @@ def test_wcs_to_dict(self):
...
+            pri = fits.PrimaryHDU()
+            self.assertFalse("CRVAL1" in pri.header)
+            self.assertFalse("CRVAL2" in pri.header)
+            self.assertFalse("CRPIX1" in pri.header)
+            self.assertFalse("CRPIX2" in pri.header)
+
+            append_wcs_to_hdu_header(wcs_info, pri.header)
+            for key in self.header_dict:
+                self.assertTrue(key in pri.header)
+                self.assertAlmostEqual(pri.header[key], self.header_dict[key])
```
wilsonbb, line 65: I'm a bit confused here. I presume the subTests are supposed to be within scope of the the `for use_dictionary in [True, False]` loop?

tests/test_wcs_utils.py
```diff
@@ -45,16 +45,54 @@ def test_wcs_to_dict(self):
...
+        self.assertEqual(wcs_dict["CRPIX1"], 100.0)
+        self.assertEqual(wcs_dict["CRPIX2"], 50.0)
+        self.assertEqual(wcs_dict["CDELT1"], 0.01)
+        self.assertEqual(wcs_dict["CDELT2"], 0.01)
+        self.assertEqual(wcs_dict["CTYPE1A"], "LINEAR  ")
+        self.assertEqual(wcs_dict["CTYPE2A"], "LINEAR  ")
+        self.assertEqual(wcs_dict["CUNIT1A"], "PIXEL   ")
+        self.assertEqual(wcs_dict["CUNIT2A"], "PIXEL   ")
+
+        # Test the we can convert to a WCS and extra predictions.
```
wilsonbb, line 84: Nit: s/Test the we/Test that we

tests/test_wcs_utils.py
jeremykubica: Good catch. It's off that this worked during testing.

jeremykubica: I fixed this.

wilsonbb (approved):

PR #433, opened 2024-01-17
Fix integer overflow

jeremykubica:
Fix an integer overflow that can happen when encoding a psi/phi array with a large number of large images. Changes two integers to long unsigned integers.


wilsonbb (approved):
LGTM

PR #434, opened 2024-01-17
Restructure the package layout to enable automatic module detection

DinoBektesevic:
The current layout of the library is

```
kbmod
... src
... ... kbmod
... ... ... [python code and modules]
... ... ... search
... ... ... .... [C++ code and modules in]
```

And something about it is breaking the module auto-detection from setuptools as mentioned in the comments [in the code](https://github.com/dirac-institute/kbmod/blob/main/setup.cfg#L28). It's also partly the reason why pyproject.toml couldn't be the only source of configuration and we still use setup.cfg.

This draft PR proposes flattening the structure of the repository into

```
kbmod
... kbmod
... ... [Python code and modules]
... src
... ... [C++ code and modules]
```

Which then let's us specify all configuration in just the pyproject.toml file and get rid of setup.cfg and have the setuptools module autodetection work. This relies on [namespace packages](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-namespace-packages) mechanism.


DinoBektesevic:
Closing this draft proposal because I like https://github.com/dirac-institute/kbmod/pull/435 better.

PR #438, opened 2024-01-19
Remove manual path construction

jeremykubica:
Use os.path.join.


delucchi-cmu:
AND an import sorting? Yes, please.

delucchi-cmu (approved):

PR #445, opened 2024-02-01
Consolidate GPU data loading

jeremykubica:
Extend the `PsiPhiArray` data structure to handle memory management of both the psi-phi array and the time array. 

This is step one in allowing individual searches by allocating `PsiPhiArray` on CPU+GPU in a single object. We can pass this object to the grid search (current approach) or to search individual trajectories (coming in a future PR).


jeremykubica:
Step 1 of N in #447

maxwest-uw (approved):

PR #446, opened 2024-02-01
Break out core evaluation into its own function

jeremykubica:
This is step 2 in enabling runs on individual trajectories and largely parallel of step 1 (#445). Breaks the core evaluation of an individual `Trajectory` into its own function.

Surprisingly timing using the diff test show this runs faster than the monolithic version.

Also deletes a function that has been commented out.


jeremykubica:
Step 2 of N in #447

PR #448, opened 2024-02-02
PixelPos cleanup

jeremykubica:
Clean up the (now unused) `PixelPos`. KBMOD now uses `Point` to pass around this information. 

Also remove two unused functions from the `StackSearch` (`get_trajectory_position` and `get_trajectory_positions`). Instead the users should be using `Trajectory`'s functions directly to consolidate logic.


DinoBektesevic (approved):
This all looks good to me.

PR #449, opened 2024-02-02
Build instructions for KBMOD in conda/virtual environments

stevenstetzler:
I had trouble building KBMOD source on the UW Hyak/Klone GPU nodes into a conda environment. The problem was that CMake's `FindPython` was finding the wrong version of Python (even though the conda environment was activated and the correct Python installation was the first on the `PATH` etc.).

It was necessary to specify:
```
cmake -DPython3_FIND_VIRTUALENV=ONLY -B src/kbmod -S .
```
when installing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,12 @@ cmake --build src/kbmod --clean-first
 ```
 To rebuild, it is sufficient to just re-run the `cmake --build` command. Optionally, invoke the cmake generated `Makefile` as `make clean && make` from the `src/kbmod` directory.
 
+If installing KBMOD into an activated conda/virtual environment, it may be necessary to specify a value for the the CMake `Python3_FIND_VIRTUALENV` configuration to allow CMake to find and use the Python installation in the conda/virtual environment.
+```
+cmake -DPython3_FIND_VIRTUALENV=ONLY B src/kbmod -S .
+cmake --build src/kbmod --clean-first
+```
+
 If you want to build the documentation you must have pandoc which seems not installable by pip.
 See [Pandoc](https://pandoc.org/installing.html), or if you are using conda:
 ```


DinoBektesevic:
The CMake `find_package` discovery system is well busted and more often doesn't work than it does. The use of that flag is by default set to `FIRST` - i.e. it should strongly prefer the `venv` (note this is specific tool) than the system since the version 3.15. It only started supporting `conda` since version 3.17 and the use before was patchy at best. 

The current latest version is 3.28, but we officially pin 3.9..3.17 in CMakeLists. The limiting factors here are the availability of new versions of cmake on different machines - especially the shared compute resources that tend to be older and do not offer a newer version of cmake. The difference of behavior between whatever version of cmake is available at a particular machine make the package discovery sometimes work and sometimes not. The fixes are individually different for many of these situations (f.e. sometimes `-DPython3_FIND_STRATEGY=LOCATION` needs to be set instead, or, ultimately, the `Python3_ROOT`). 

Same is with the GPU discovery mechanisms. It needs `nvcc`, but also the correctly matched driver versions. This is done by exporting the installation location into `PATH` and let the global discovery find them. 

So, I'm not sure this is the solution to the problem. I'm not sure there is a single correct fix. It would be beneficial to eventually write a general "installation FAQ/Debug" in the [documentation](https://epyc.astro.washington.edu/~kbmod/) but that is atm far from getting updated so I'm happy to add the commentary to the README. Please clarify the following before merging:

> Older versions of CMAKE (<3.17) sometimes do not correctly discover the Python installation in virtual environments, causing the installation to fail. In those cases, and if you have CMAKE version >3.15, [.... your sentences ...]

DinoBektesevic (approved):

stevenstetzler:
I'm using CMake version 3.27.6, so the error didn't occur with an older version. 

Maybe this best sits in an FAQ like you mention instead of the README like you mention.

DinoBektesevic:
It feels like this was said, but I just want to do another check, @stevenstetzler and others, is this PR set to be closed or merged?

DinoBektesevic:
Since there's been no answers here and it's obvious from other experiences that the way we build and test KBMOD in different environments needs to be improved I've opened an issue https://github.com/dirac-institute/kbmod/issues/503 and am closing this PR.

PR #457, opened 2024-02-06
reproject 

### LINKED ISSUE #411: Resample images from different WCSs
Step 1 of reprojection: Resample the images from a bunch of close WCSs into a common WCS.

maxwest-uw:
resolves #411 

step 1 of the reprojection changes
- given a common `astropy.wcs.WCS` object and a `WorkUnit` with `per_image_wcs`'s, create a new `WorkUnit` where all the images are reprojected into the `common_wcs`. also condenses images with the same `obstime` to be the same reprojected image.


src/kbmod/reprojection.py
```diff
@@ -0,0 +1,129 @@
...
+    image : `kbmod.search.RawImage`
+        The image data to be reprojected.
+    original_wcs : `astropy.wcs.WCS`
+        The WCS of the original image.
+    common_wcs : `astropy.wcs.WCS`
+        The WCS to reproject all the images into.
+    obs_time : float
+        The MJD of the observation.
+    Returns
+    ----------
```
jeremykubica, line 26: Small thing: Can you explicitly list the outputs?

```
    new_image : `numpy.ndarray`
      The image data reprojected...
    footprint : ...
```

src/kbmod/reprojection.py
```diff
@@ -0,0 +1,129 @@
...
+
+            footprint_add += footprint
+            # we'll enforce that there be no overlapping images at the same time,
+            # for now. We might be able to add some ability co-add in the future.
+            if np.any(footprint_add > 1.0):
+                raise ValueError("Images with the same obstime are overlapping.")
+
+            reprojected_variance, _ = reproject_raw_image(variance, original_wcs, common_wcs, time)
+
+            reprojected_mask, _ = reproject_raw_image(mask, original_wcs, common_wcs, time)
```
jeremykubica, line 91: Since reproject_raw_image is using interpolation, we should reset the mask to a bitmask. We can probably do that with something like `reprojected_mask[reprojected_mask > 0.0] = 1`

(That would need to be tested)

tests/test_reprojection.py
```diff
@@ -0,0 +1,64 @@
...
+
+        test_vals = np.array([231.61615, 113.59214, 166.82635]).astype("float32")
+        # make sure the PSF for the object hasn't been warped
+        # in the no-op case
+        assert data[0][0][10][43] == test_vals[0]
+
+        # test other object locations
+        assert data[1][0][15][46] == test_vals[1]
+        assert data[2][0][21][49] == test_vals[2]
+
```
jeremykubica, line 44: Can you add a test that the mask is all 0 or 1?

jeremykubica (changes requested):

src/kbmod/reprojection.py
```diff
@@ -0,0 +1,135 @@
...
+            variance_add += reprojected_variance
+            mask_add += reprojected_mask
+
+        # change all the values where there are is no corresponding data to `KB_NO_DATA`.
+        gaps = footprint_add == 0.0
+        science_add[gaps] = KB_NO_DATA
+        variance_add[gaps] = KB_NO_DATA
+        mask_add[gaps] = 1
+
+        mask_add = np.where(np.isclose(mask_add, 0.0, atol=1e-01), 0.0, 1.0)
```
jeremykubica, line 114: It is worth adding a short comment here that this is effectively transforming the mask back into a 0/1 bit mask.

jeremykubica (approved):

PR #461, opened 2024-02-07
Region searching

ColinOrionChandler:
Updating many components of the Region Searching Workbook.

### DIFF
--- a/notebooks/region_search/Region Searching Workbook.ipynb
+++ b/notebooks/region_search/Region Searching Workbook.ipynb
@@ -2815,7 +2815,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 74,
+   "execution_count": 77,
    "id": "bfb23aef",
    "metadata": {},
    "outputs": [],
@@ -2826,7 +2826,7 @@
     "    Implementing an extremely basic (RA, Dec) query functionality.\n",
     "    This will work within the overlapping_sets framework.\n",
     "    \"\"\"\n",
-    "    set_to_corners"
+    "    pass"
    ]
   },
   {


notebooks/region_search/Region Searching Workbook.ipynb
```diff
@@ -5,7 +5,7 @@
    "id": "d94d8d2b",
```
wilsonbb, line 5: <pre class="ql-syntax" spellcheck="false">Line #7.        set_to_corners
</pre><p>It looks like this function is broken? If you want to leave it in as an interface, you could replace <code>set_to_corners</code> with <code>pass</code> for now  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/dirac-institute/kbmod/pull/461/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='94'/>

notebooks/region_search/Region Searching Workbook.ipynb
```diff
@@ -2794,6 +2813,22 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def ra_dec_search_overlapping_sets(df, overlapping_sets):\n",
+    "    \"\"\"\n",
+    "    2/6/2024 COC\n",
+    "    Implementing an extremely basic (RA, Dec) query functionality.\n",
+    "    This will work within the overlapping_sets framework.\n",
+    "    \"\"\"\n",
+    "    set_to_corners"
```
wilsonbb, line 2829: It looks like this function is broken? If you want to leave it in as an interface, you could replace set_to_corners with pass for now

PR #466, opened 2024-02-08
Consolidate test logic

jeremykubica:
Modifies the interface of `FakeDataSet` to take a list of times to allow greater flexibility. Change a bunch of tests to use `FakeDataSet` (and also `make_trajectory` while we are at it). Fixes style issues (`imCount` -> `img_count` for consistency) and removes unneeded test parameters from before the tests were broken up.

PR #467, opened 2024-02-12
Reorganize the fake data creators

jeremykubica:
Put the all the fake data creation helpers in one directory. Also cleans up some more tests (using `FakeDataSet` and `make_trajectory`) that were not included in #466.


tests/test_analysis_utils.py
```diff
@@ -85,21 +85,20 @@ def setUp(self):
...
-        self.imlist = []
-        self.time_list = []
-        for i in range(self.img_count):
-            time = i / self.img_count
-            self.time_list.append(time)
-            im = LayeredImage(self.dim_x, self.dim_y, self.noise_level, self.variance, time, self.p, i)
-            self.imlist.append(im)
-        self.stack = ImageStack(self.imlist)
+        self.dim_x = 100
+        self.dim_y = 100
```
drewoldag, line 89: This looks inconsistent with what was replaced. Before self.dim_x = 15, here self.dim_x = 100. Same for dim_y. 

Is this intentional?

tests/test_search.py
```diff
@@ -312,20 +308,16 @@ def test_median_stamps_trj(self):
...
         self.assertAlmostEqual(np.median(pix_values1), medianStamps1.get_pixel(2, 2), delta=1e-5)
 
     def test_median_stamps_no_data(self):
         # Create a Trajectory that goes through the masked pixels.
-        trj = Trajectory()
-        trj.x = self.masked_x
-        trj.y = self.masked_y
-        trj.vx = 0
-        trj.vy = 0
+        trj = make_trajectory(x=self.masked_x, y=self.masked_y, vx=0.0, vy=0.0)
```
drewoldag, line 320: Just to confirm, no concerns about int vs floats here? Correct?

drewoldag (approved):
Looks pretty good. I left 2 questions about some _maybe_ inconsistencies, but I'm not well enough versed to know for sure.

tests/test_search.py
jeremykubica: Correct. The data for velocities is stored as floats, so they were being converted before.

tests/test_analysis_utils.py
jeremykubica: This is an intentional change so the fake data can be used by multiple tests. The second test needs the larger image. The first only needs a smaller one, but the larger one doesn't hurt.

PR #473, opened 2024-02-15
Additions to cmakelist and instructions for installation

jeremykubica:
Adds some more outputting to cmakelist.txt as well as compiler conditions to support the combination of GPU/CPU on baldur.

Adds some more details about installation to the documentation and another documentation link to the readme. Also cleans up an unused variable in `analysis_utils.py`.


DinoBektesevic (approved):
Debugging-wise I'm happy to have these messages in here, but pip suppresses STD outs so they won't appear via pip install.

jeremykubica:
> Debugging-wise I'm happy to have these messages in here, but pip suppresses STD outs so they won't appear via pip install.

Correct. That's why I added documentation about running cmake directly.

Issue #418, opened 2023-12-21
Add documentation for unified result format


jeremykubica:
Add documentation for the serialized ResultList including how it is stored and how a user could load and use the data.

jeremykubica:
Closed by #475

PR #475, opened 2024-02-15
Update the analysis demo notebook

jeremykubica:
Updates the `kbmod_analysis_demo.ipynb` notebook to read in a combined result file instead of individual files, addressing part of #418.

In the process updates `ResultList` to be able to write out the individual stamps to a table file given an override parameter and passes that through to run_search. 

This PR reused the `KBMOD_Demo.ipynb` file from #474, so some of the changes are the same. The new (additional) changes in `KBMOD_Demo.ipynb` reflect the generation of output with all stamps so we can use the output data directly for the analysis notebook.


src/kbmod/result_list.py
```diff
@@ -920,11 +920,14 @@ def write_table(self, filename, overwrite=True):
             The name of the result file.
         overwrite : `bool`
             Overwrite the file if it already exists.
+        keep_all_stamps : `bool`
+            Keep individual stamps for each result and time step.
+            This is very expensive.
```
wilsonbb, line 925: Optional nit: since this is so expensive maybe add the obvious note to the docstring that it is False by default.

wilsonbb (approved):
LGTM

PR #478, opened 2024-02-17
New Sky Patches and Regions sections

ColinOrionChandler:
The framework for dividing the sky into patches (with optional overlap) is here, along with some minimal LSST sphgeom region work (e.g., intersection checking).


wilsonbb (approved):
Thanks, Colin

PR #484, opened 2024-02-22
Remove C++ code for loading fits, saving fits, and generating random data

jeremykubica:
This PR removes the C++ code for working with legacy data and generating the random data allowing use to remove dependencies on fitsio and to consolidate the file and fake data generation in a few locations.

Specific changes include:
- Loader functions are removed from `RawImage`, `LayeredImage`, and `ImageStack` and are replaced by a single helper function `load_deccam_layered_image`
- Saving functions are removed from `RawImage` and `LayeredImage`  and are replaced by a single helper function `save_deccam_layered_image`
- All dependencies on cfitsio are removed.
- `LayeredImage` constructors that created random data are replaced by a helper function `make_fake_layered_image` in `fake_data/fake_data_creator.py`


src/kbmod/data_interface.py
```diff
@@ -3,12 +3,111 @@
...
+    if not Path(filename).is_file():
+        raise FileNotFoundError(f"{filename} not found")
+
+    img = None
+    with fits.open(filename) as hdul:
+        if len(hdul) < 4:
+            raise ValueError("Not enough extensions for legacy deccam format")
+
+        # Extract the obstime trying from a few keys and a few extensions.
+        obstime = 0.0
```
wilsonbb, line 49: 1) Is it ever possible for the obstime to be missing? Would we want to fail if so?

2) Currently we set the default obstime to -1 and check for that here (https://github.com/dirac-institute/kbmod/blob/1cc181adfba39c29d8ebc3c76f3e9746409622b7/src/kbmod/data_interface.py#L217),  so we should probably still be using `obstime = -1.0`

src/kbmod/data_interface.py
```diff
@@ -3,12 +3,111 @@
...
+    if Path(filename).is_file() and not overwrite:
+        raise ValueError(f"{filename} already exists")
+
+    hdul = fits.HDUList()
+
+    # Create the primary header.
+    pri = fits.PrimaryHDU()
+    pri.header["MJD"] = img.get_obstime()
+    if wcs is not None:
+        append_wcs_to_hdu_header(wcs, pri.header)
```
wilsonbb, line 92: Should all valid deccam images have a WCS? Might make sense to fail if this is missing

wilsonbb (approved):
LGTM, just some comments about handling invalid inputs

src/kbmod/data_interface.py
jeremykubica: I don't think they do for all the old data we have used historically. We will want to enforce a WCS in the future, but not in the legacy interface.

src/kbmod/data_interface.py
jeremykubica: For 1: yes. There were legacy files without times where we needed to read obstime from an auxiliary text file. So we don't want it to fail outright in this function.
For 2: Good call. I changed it.

Issue #486, opened 2024-02-23
Validate data coming into a LayeredImage


jeremykubica:
When creating a `LayeredImage`, check the input science and variance layers for invalid values such as `NaN` or infinity. These should be set to `NO_DATA` and the corresponding bit set in the mask layer. In addition the code should check that any pixels in those layers whose value is `NO_DATA` should have corresponding bits in the mask layer set to 1.

jeremykubica:
As per the discussion on the KBMOD pipelines document, it is better to have this happen in the standardizer.

Issue #441, opened 2024-01-26
Inconsistent checking of `NaN`s throughout the codebase


DinoBektesevic:
There are a [couple of places](https://github.com/dirac-institute/kbmod/pull/334/files/c2ec4dca1e2b41c84c895767160ccf25e6dd947e#r1457964401) within the codebase where we define "empty" or "non-existing" values. NaNs are sometimes handled as existing values and sometimes as non-existing ones. It would be good to go through the codebase and make sure the way we handle NaNs is consistent.

nemo-1999:
I would like to work on this issue

nemo-1999:
I would like to work on this

jeremykubica:
Sorry @nemo-1999, this ended up getting pulled into a larger set of changes. It will be addressed by a combination of #493 and #504.

jeremykubica:
Closed by #504

PR #487, opened 2024-02-23
Clean the values for a LayeredImage

jeremykubica:
When creating a `LayeredImage`, replace `NaN` and infinity with `NO_DATA`. Make the NO_DATA and mask bit consistent across all the layers.


jeremykubica:
Closes #486 and #441

src/kbmod/search/layered_image.cpp
```diff
@@ -17,6 +17,29 @@ LayeredImage::LayeredImage(const RawImage& sci, const RawImage& var, const RawIm
...
+    float* var_pixels = variance.data();
+    float* mask_pixels = mask.data();
+
+    for (int i = 0; i < num_pixels; ++i) {
+        // Mask NaNs, infinities, or NO_DATA tags in sci or var layers consistently
+        // in all layers.
+        // Does not apply masking to the information to sci or var.
+        if (std::isnan(sci_pixels[i]) || std::isnan(var_pixels[i])
+            || (sci_pixels[i] == NO_DATA) || (var_pixels[i] == NO_DATA) 
+            || std::isinf(sci_pixels[i]) || std::isinf(var_pixels[i])) {
```
DinoBektesevic, line 37: I did not know that `NaN` is an actual thing in a standard, so these CPP `isnan` checks behave equally to python is-nan checks. I just assumed it's like a singleton defined somewhere in Numpy and they're not comparable. 

Would it be better for us then to use nan as the mask instead of `NO_DATA`? Do they exist on the GPU too? This could be an issue, doesn't have to be sorted in this PR.

tests/test_layered_image.py
```diff
@@ -115,6 +116,36 @@ def test_create_from_layers(self):
...
+        var = RawImage(width, height, 1.0, -1.0)
+        msk = RawImage(width, height, 0, -1.0)
+
+        # Make a few of the values NaN, inf, or KB_NO_DATA.
+        sci.set_pixel(2, 5, KB_NO_DATA)
+        var.set_pixel(12, 5, KB_NO_DATA)
+        sci.set_pixel(2, 6, math.nan)
+        var.set_pixel(12, 6, math.nan)
+        sci.set_pixel(2, 7, math.inf)
+        var.set_pixel(12, 7, math.inf)
```
DinoBektesevic, line 132: Can we test this with `numpy.nan`, `np.inf` and `-np.inf`?

DinoBektesevic (commented):
Looks great to me but I'll defer the approval to Max because I know this impacts the reproject stuff. I just had two short questions.

src/kbmod/search/layered_image.cpp
maxwest-uw: also, since there's an option in `reproject` to ignore bad values, it might make more sense to do it this way ^^

jeremykubica:
Closing as we are no longer going with this approach

PR #490, opened 2024-02-26
Confirm data validity before encoding psi and phi arrays

jeremykubica:
Check that the psi and phi images do not have invalid values (NaN or inf).

### DIFF
--- a/src/kbmod/search/psi_phi_array.cpp
+++ b/src/kbmod/search/psi_phi_array.cpp
@@ -273,7 +273,7 @@ void fill_psi_phi_array(PsiPhiArray& result_data, int num_bytes, const std::vect
                     throw std::runtime_error("Invalid value found in PSI array");
                 }
 
-                float phi_val = psi_imgs[t].get_pixel({row, col});
+                float phi_val = phi_imgs[t].get_pixel({row, col});
                 if (std::isnan(phi_val) || std::isinf(phi_val)) {
                     throw std::runtime_error("Invalid value found in PHI array");
                 }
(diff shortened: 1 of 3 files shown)


src/kbmod/search/psi_phi_array.cpp
```diff
@@ -263,6 +263,24 @@ void fill_psi_phi_array(PsiPhiArray& result_data, int num_bytes, const std::vect
...
+    // before allocating space, encoding, etc.
+    for (int t = 0; t < num_times; ++t) {
+        for (int row = 0; row < height; ++row) {
+            for (int col = 0; col < width; ++col) {
+                float psi_val = psi_imgs[t].get_pixel({row, col});
+                if (std::isnan(psi_val) || std::isinf(psi_val)) {
+                    throw std::runtime_error("Invalid value found in PSI array");
+                }
+
+                float phi_val = psi_imgs[t].get_pixel({row, col});
```
delucchi-cmu, line 276:
```suggestion
                float phi_val = phi_imgs[t].get_pixel({row, col});
```

tests/test_psi_phi_array.py
```diff
@@ -174,6 +175,35 @@ def test_fill_psi_phi_array(self):
...
+            arr,
+            4,
+            [self.psi_1, self.psi_2],
+            [self.phi_1, self.phi_2],
+            self.zeroed_times,
+            False,
+        )
+
+        # Fails with inf
+        self.psi_1.set_pixel(4, 3, math.inf)
```
delucchi-cmu, line 195: It would also be good to test for invalid phi values (might have caught the above typo). A coverage check might have also revealed that the line that raises an error was not covered by tests?

src/kbmod/search/psi_phi_array.cpp
jeremykubica: Good catch. That's embarrassing.

delucchi-cmu (approved):

PR #493, opened 2024-02-27
Add support for NaNs in kernel functions

jeremykubica:
Extends the NO_DATA behavior to account for NaNs. Moves the test for whether a pixel's value is valid to a helper function `pixel_value_valid` so we only have a single place to change behavior if we want to extend this for things like handling inf.

This addresses most of #441. The next step would be determining if we should take out the NO_DATA tag altogether and use NaN's to represent that.


src/kbmod/search/kernels.cu
```diff
@@ -23,6 +23,8 @@
 
 namespace search {
 
+__host__ __device__ bool device_pixel_valid(float value) { return ((value != NO_DATA) && !isnan(value)); }
```
DinoBektesevic, line 26: I'm just a little bit confused why we're keeping the NO_DATA around. If we set NO_DATA to nan and check isnan does that break things. This can be a matter for a different PR though.

DinoBektesevic (approved):
Looks ok to me.

src/kbmod/search/kernels.cu
jeremykubica: The main reason is that removing NO_DATA as a flag the code sets will be more extensive. I will work on that in a later PR. But I wanted to get the code to support checking for NaNs in the short term.

PR #504, opened 2024-03-04
Fully replace old NO_DATA values

jeremykubica:
Replaces the `NO_DATA = -9999.0` with setting `NO_DATA = NAN`. This breaks all direct `==` comparisons, so we make use of the new `pixel_value_valid()` function for testing data validity.  Unfortunately this breaks the vectorization of a few Eigen functions (although there is probably a better way to structure them to keep the vectorization intact).

The `NO_DATA` and `KB_NO_DATA` flags are kept so we can use them to set data throughout the program.

The structure of the change is also meant to provide flexibility down the road if we want to change how we represent masked data. We only need to change a few locations within `common.h` instead of changing values throughout the files.

### DIFF
--- a/src/kbmod/filters/stamp_filters.py
+++ b/src/kbmod/filters/stamp_filters.py
@@ -12,7 +12,6 @@ from kbmod.configuration import SearchConfiguration
 from kbmod.result_list import ResultRow
 from kbmod.search import (
     HAS_GPU,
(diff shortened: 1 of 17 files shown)


maxwest-uw:
this looks good to me : )

maxwest-uw (approved):

PR #506, opened 2024-03-05
Remove analysis_utils.py

jeremykubica:
Removes `analysis_utils.py` by moving the batch loading into `run_search.py`. Includes a few small changes to clean up the interface including:
- Set everything in the loading function by a `SearchConfiguration` object instead of a combination of object attributes and function parameters.
- Change `do_gpu_search` to take an `ImageStack` object instead of a `StackSearch` object. It now creates and sets the `StackSearch` object itself.


jeremykubica:
Closes #507

maxwest-uw (approved):

PR #513, opened 2024-03-08
Add More TrajectoryList Filtering

jeremykubica:
Add the ability to filter a `TrajectoryList` by the `Trajectory`'s `valid` bit.

Also make the vector-based constructor accessible via python.


wilsonbb (approved):
LGTM

PR #514, opened 2024-03-08
Add initial not-that-generic plotting utilities.

DinoBektesevic:
There seems to be at least a slight need for common plotting utilities. I added some into analysis module. Feel free to edit and push new code to this PR, not sure what are the things we really need, but we can build to the requirements over time. 

This initial commit covers the example plotting functions that underpin the shared notebook that runs the imdiff recipe data through processing.

### DIFF
--- a/src/kbmod/analysis/plotting.py
+++ b/src/kbmod/analysis/plotting.py
@@ -192,7 +192,7 @@ def plot_footprints(ax, wcs_list):
     return ax
 
 
-def plot_all_objs(ax, objects, count=0, show_field=False, center=None, radius=1.1, lw=0.9, ms=1):
+def plot_all_objs(ax, objects, count=-1, show_field=False, center=None, radius=1.1, lw=0.9, ms=1):
     """Plots object tracks on the given axis.
 
     Parameters
@@ -200,10 +200,11 @@ def plot_all_objs(ax, objects, count=0, show_field=False, center=None, radius=1.
     ax : `matplotlib.pyplot.Axes`
         Axis.
     objects : `astropy.table.Table`
-        Table of objects. Must contain column ``Name`` identifying
-        unique objects.
+        Table of objects. Must contain columns ``Name``, ``RA`` and ``DEC``
+        identifying unique objects, their right ascension and declination
+        coordinates.
     count : `int`
-        Number of tracks to plot. Default 0.
+        Number of tracks to plot. Default -1, plots all objects.
     show_field : `bool`
         `False` by default, when `True` requires ``center`` and
         ``radius``to be specified. Adds an circle around the center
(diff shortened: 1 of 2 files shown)


src/kbmod/analysis/plotting.py
```diff
@@ -0,0 +1,382 @@
...
+
+def plot_all_objs(ax, objects, count=0, show_field=False, center=None, radius=1.1, lw=0.9, ms=1):
+    """Plots object tracks on the given axis.
+
+    Parameters
+    ----------
+    ax : `matplotlib.pyplot.Axes`
+        Axis.
+    objects : `astropy.table.Table`
+        Table of objects. Must contain column ``Name`` identifying
```
jeremykubica, line 203: Also need to contain "RA" and "DEC" columns.

src/kbmod/analysis/plotting.py
```diff
@@ -0,0 +1,382 @@
...
+        unique objects.
+    count : `int`
```
jeremykubica, line 205: You should mention that if count == 0 then it plots all of the objects. I figured it would plot zero until I got to the code below.

src/kbmod/analysis/plotting.py
```diff
@@ -0,0 +1,382 @@
...
+            ax = plot_footprint(ax, wcs)
+            pt = wcs.pixel_to_world(0, 0)
+            # units vs quantities
+            xoffset = txt_x_offset.to(u.deg).value
+            yoffset = txt_y_offset.to(u.deg).value
+            # we need to move diagonally to the right and down to center the text
+            x, y = pt.ra.deg + xoffset, pt.dec.deg + yoffset
+            ax.text(x, y, hdu.header.get("EXTNAME", None), clip_on=True)
+
+    # ax = plot_footprints(ax, wcss)
```
jeremykubica, line 286: Can this line be removed?

src/kbmod/analysis/plotting.py
```diff
@@ -0,0 +1,382 @@
...
+    -------
+    figure : `matplotlib.pyplot.Figure`
+        Modified Figure.
+    ax : `matplotlib.pyplot.Axes`
+        Modified Axes.
+    """
+    if ax is None and figure is None:
+        fig, ax = plt.subplots(figsize=(15, 10))
+    elif ax is not None and figure is not None:
+        raise ValueError("Provide both figure and axis, or provide none.")
```
jeremykubica, line 367: I don't under stand this error. The phrasing makes it sound like you need to provide both axis and figure, but it only triggers when both are provided. Should it say "Provide either a figure or axis"?

jeremykubica (approved):
A few of small changes requested before merging.

src/kbmod/analysis/plotting.py
DinoBektesevic: Whoops, that should've been an `or`. Thanks

src/kbmod/analysis/plotting.py
DinoBektesevic: I made that `-1` to make it less confusing (ironically)

Issue #500, opened 2024-02-29
Add a mask trajectory function


jeremykubica:
Add a function to `masking.py` that takes a trajectory, a `WorkUnit` or `ImageStack`, and radius then proceeds to add a mask to each image that is centered on the trajectory's predicted position and has the given radius. This can be used to support iterative search (remove the brightest objects first) or to remove stationary objects (with the upcoming `SingleVelocitySearch` generator in #485 and a setting of zero velocity).

vlnistor:
Hi @jeremykubica, can I be assigned to this?

jeremykubica:
closed by #515

PR #515, opened 2024-03-10
Mask based on a trajectory

vlnistor:
This PR tackles #500. It adds a function `mask_trajectory` that will mask based on the input trajectory and radius. 

An open question I still have is if the function needs to explicitly apply all active masks (like in the `apply_mask_operations` function):

```    
for i in range(stack.img_count()):
        stack.get_single_image(i).apply_mask(0xFFFFFF)
```

### DIFF
--- a/src/kbmod/masking.py
+++ b/src/kbmod/masking.py
@@ -5,7 +5,7 @@ import kbmod.search as kb
 from kbmod.search import RawImage
 
 
-def mask_trajectory(trj, stack, r, bitmask=1):
+def mask_trajectory(trj, stack, r):
     """Will apply a circular mask of radius `r` around
     the trajectory's predicted position in each image.
 
@@ -15,7 +15,7 @@ def mask_trajectory(trj, stack, r, bitmask=1):
         The trajectory along which to apply the mask.
     stack : `ImageStack`
         The stack of images to apply the mask to.
-    r : `float`
+    r : `int`
         The radius of the circular mask to apply.
 
     Returns
@@ -28,17 +28,17 @@ def mask_trajectory(trj, stack, r, bitmask=1):
     height = stack.get_height()
 
     for i in range(stack.img_count()):
-        time = stack.get_single_image(i).get_obstime() - stack.get_single_image(0).get_obstime()
+        img = stack.get_single_image(i)
+        time = img.get_obstime() - stack.get_single_image(0).get_obstime()
+        origin_of_mask = (int(trj.get_x_pos(time) + 0.5), int(trj.get_y_pos(time) + 0.5))
 
-        origin_of_mask = (trj.get_x_pos(time), trj.get_y_pos(time))
-        trajectory_mask = RawImage(width, height)
-
-        for y in range(height):
-            for x in range(width):
-                if (x - origin_of_mask[0]) ** 2 + (y - origin_of_mask[1]) ** 2 <= r**2:
-                    trajectory_mask.set_pixel(y, x, bitmask)
-        stack.get_single_image(i).union_masks(trajectory_mask)
-
+        for dy in range(-r, r + 1):
+            for dx in range(-r, r + 1):
+                if dx**2 + dy**2 <= r**2:
+                    x = origin_of_mask[0] + dx
+                    y = origin_of_mask[1] + dy
+                    if x >= 0 and x < width and y >= 0 and y < height:
+                        img.mask_pixel(y, x)
     return stack
 
 
--- a/src/kbmod/masking.py
+++ b/src/kbmod/masking.py
@@ -2,6 +2,44 @@
 """
 
 import kbmod.search as kb
+from kbmod.search import RawImage
+
+
+def mask_trajectory(trj, stack, r, bitmask=1):
+    """Will apply a circular mask of radius `r` around
+    the trajectory's predicted position in each image.
+
+    Attributes
+    ----------
+    trj : `Trajectory`
+        The trajectory along which to apply the mask.
+    stack : `ImageStack`
+        The stack of images to apply the mask to.
+    r : `float`
+        The radius of the circular mask to apply.
+
+    Returns
+    -------
+    stack : `ImageStack`
+        The stack after the masks have been applied.
+    """
+
+    width = stack.get_width()
+    height = stack.get_height()
+
+    for i in range(stack.img_count()):
+        time = stack.get_single_image(i).get_obstime() - stack.get_single_image(0).get_obstime()
+
+        origin_of_mask = (trj.get_x_pos(time), trj.get_y_pos(time))
+        trajectory_mask = RawImage(width, height)
+
+        for y in range(height):
+            for x in range(width):
+                if (x - origin_of_mask[0]) ** 2 + (y - origin_of_mask[1]) ** 2 <= r**2:
+                    trajectory_mask.set_pixel(y, x, bitmask)
+        stack.get_single_image(i).union_masks(trajectory_mask)
+
+    return stack
 
 
 def mask_flags_from_dict(mask_bits_dict, flag_keys):
--- a/tests/test_masking.py
+++ b/tests/test_masking.py
@@ -2,8 +2,9 @@ import unittest
 
 from kbmod.configuration import SearchConfiguration
 from kbmod.fake_data.fake_data_creator import make_fake_layered_image
-from kbmod.masking import apply_mask_operations
+from kbmod.masking import apply_mask_operations, mask_trajectory
 from kbmod.search import *
+from kbmod.trajectory_utils import make_trajectory
 
 
 class test_run_search_masking(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


src/kbmod/masking.py
```diff
@@ -2,7 +2,42 @@
...
+    r : `float`
+        The radius of the circular mask to apply.
+
+    Returns
+    -------
+    stack : `ImageStack`
+        The stack after the masks have been applied.
+    """
+
+    dt = stack.get_single_image(1).get_obstime() - stack.get_single_image(0).get_obstime()
```
jeremykubica, line 26: This should be computed within the loop as the time steps may vary between images. So it should be:

dt = stack.get_single_image(i).get_obstime() - stack.get_single_image(0).get_obstime()

src/kbmod/masking.py
```diff
@@ -2,7 +2,42 @@
...
+    width = stack.get_single_image(0).get_width()
```
jeremykubica, line 27: You could also use the stack's `get_width()` and `get_height()` functions directly.

src/kbmod/masking.py
```diff
@@ -2,7 +2,42 @@
...
+    height = stack.get_single_image(0).get_height()
+
+    for i in range(stack.img_count()):
+        origin_of_mask = (trj.x + dt * trj.vx * i , trj.y + dt * trj.vy * i)
```
jeremykubica, line 31: Can you use `trj.get_x_pos()` and `trj.get_y_pos()` here? Both are defined in `search/common.h`. Using that will keep the prediction logic centralized in case we make a change in the future.

src/kbmod/masking.py
```diff
@@ -2,7 +2,42 @@
...
+    dt = stack.get_single_image(1).get_obstime() - stack.get_single_image(0).get_obstime()
+    width = stack.get_single_image(0).get_width()
+    height = stack.get_single_image(0).get_height()
+
+    for i in range(stack.img_count()):
+        origin_of_mask = (trj.x + dt * trj.vx * i , trj.y + dt * trj.vy * i) 
+        trajectory_mask = RawImage(width, height)
+
+        for x in range(width):
+            for y in range(height):
```
jeremykubica, line 35: Small comment: Since we store the data row first, this would be a bit more efficient with the `y` loop first.

tests/test_masking.py
```diff
@@ -26,6 +26,28 @@ def setUp(self):
             self.imlist.append(im)
         self.stack = ImageStack(self.imlist)
 
+    def test_apply_trajectory_mask(self):
+        starting_pixel = (20,20) 
+        velocity = (50.0, -10.0)
+        radius = 5.0
+        dt = self.stack.get_single_image(1).get_obstime() - self.stack.get_single_image(0).get_obstime()
```
jeremykubica, line 33: Same comment about `dt` as in `masking.py`

tests/test_masking.py
```diff
@@ -26,6 +26,28 @@ def setUp(self):
...
+            msk = masked_stack.get_single_image(i).get_mask()
+        
+            for x in range(self.dim_x):
+                for y in range(self.dim_y):
+                    distance = (x - origin_of_mask[0])**2 + (y - origin_of_mask[1])**2
+                    if distance <= radius**2:
+                        self.assertEqual(msk.get_pixel(y, x), bitmask)
+            
+
+
```
jeremykubica, line 50: Can you run black formatting on this change (`black .`). This will pick up some style guides which will be checked for in the presubmit tests.

jeremykubica (changes requested):

jeremykubica:
In reference to the question of applying all masks, the answer is no. We are moving to a set up where the mask bits are preprocessed earlier in the code and the mask layer becomes binary. So using the approach you implemented works well.

src/kbmod/masking.py
```diff
@@ -2,7 +2,42 @@
...
+                if (x - origin_of_mask[0])**2 + (y - origin_of_mask[1])**2 <= r**2:
+                    trajectory_mask.set_pixel(y,x,bitmask)
```
jeremykubica, line 37: You are going to need to bounds check here. `set_pixel` will throw an error if you try to set a pixel that is off the image and trajectories might go off image.

jeremykubica (changes requested):

src/kbmod/masking.py
vlnistor: Moved inside loop

src/kbmod/masking.py
vlnistor: Moved to `get_x_pos()` and `get_y_pos`

src/kbmod/masking.py
vlnistor: Reordered the loops

tests/test_masking.py
vlnistor: Refactored

tests/test_masking.py
vlnistor: Ran formatting. Is there any reason we don't configure [husky](https://typicode.github.io/husky/) to run formatting before commits?

src/kbmod/masking.py
vlnistor: Won't `x` and `y` always be within the image here because of `y in range(height)` and `x in range(width)` where the `width` and `height` are the actual dimensions of the image?

jeremykubica: Yes. Sorry about that. You are correct.

I was thinking about a loop in the format of iterating over the dx and dy for the radius. I've added a comment about an alternate way to do this.

src/kbmod/masking.py
```diff
@@ -2,6 +2,44 @@
...
+
+    for i in range(stack.img_count()):
+        time = stack.get_single_image(i).get_obstime() - stack.get_single_image(0).get_obstime()
+
+        origin_of_mask = (trj.get_x_pos(time), trj.get_y_pos(time))
+        trajectory_mask = RawImage(width, height)
+
+        for y in range(height):
+            for x in range(width):
+                if (x - origin_of_mask[0]) ** 2 + (y - origin_of_mask[1]) ** 2 <= r**2:
```
jeremykubica, line 38: In general, I think we are going to have `r` much smaller than `width` or `height`. So it might be more efficient to do something like:

```
for dy in range(-r, r+1):
    for dx in range(-r, r+1):
        if dx**2 + dy**2 <= r**2:
            x = origin_of_mask[0] + dx
            y = origin_of_mask[1] + dy
            if (x >= 0) and (x < width) and (y >= 0) and (y < height):
...
```

This is what I was thinking when I made the comment about bounds checking.

src/kbmod/masking.py
```diff
@@ -2,6 +2,44 @@
...
+                    trajectory_mask.set_pixel(y, x, bitmask)
```
jeremykubica, line 39: Based on this PR, I added a helper function that will allow you to directly mask a value in a `LayeredImage` (#516). It would be more efficient to not create an entire new mask. If you switch the iteration (as described in the comment above), can you also switch to using `mask_pixel()`?

jeremykubica (commented):
Some optional speedups for discussion.

src/kbmod/masking.py
vlnistor: This is much better. Switched to this

src/kbmod/masking.py
vlnistor: Making use of it

src/kbmod/masking.py
vlnistor: Switched to this

src/kbmod/masking.py
```diff
@@ -2,6 +2,44 @@
...
+    stack : `ImageStack`
+        The stack after the masks have been applied.
+    """
+
+    width = stack.get_width()
+    height = stack.get_height()
+
+    for i in range(stack.img_count()):
+        time = stack.get_single_image(i).get_obstime() - stack.get_single_image(0).get_obstime()
+        origin_of_mask = (int(trj.get_x_pos(time) + 0.5), int(trj.get_y_pos(time) + 0.5))
```
vlnistor, line 32: Note I'm assuming I need to add a `0.5` offset.

src/kbmod/masking.py
```diff
@@ -2,6 +2,44 @@
 """
 
 import kbmod.search as kb
+from kbmod.search import RawImage
+
+
+def mask_trajectory(trj, stack, r, bitmask=1):
```
jeremykubica, line 8: You can remove bitmask from the arguments.

src/kbmod/masking.py
jeremykubica: That works.

jeremykubica (approved):

src/kbmod/masking.py
vlnistor: Removed bitmask

vlnistor:
@jeremykubica I reformatted both files using `black`. Should pass now

tests/test_masking.py
```diff
@@ -26,6 +27,27 @@ def setUp(self):
...
         self.stack = ImageStack(self.imlist)
 
+    def test_apply_trajectory_mask(self):
+        starting_pixel = (20, 20)
+        velocity = (50.0, -10.0)
+        radius = 5
+        trj = make_trajectory(starting_pixel[0], starting_pixel[1], velocity[0], velocity[1])
+        bitmask = 1
+
+        masked_stack = mask_trajectory(trj, self.stack, radius, bitmask)
```
jeremykubica, line 37: You also need to remove the bitmask argument here (and can delete the definition on line 35).

src/kbmod/masking.py
```diff
@@ -2,6 +2,44 @@
...
+        The stack after the masks have been applied.
+    """
+
+    width = stack.get_width()
+    height = stack.get_height()
+
+    for i in range(stack.img_count()):
+        time = stack.get_single_image(i).get_obstime() - stack.get_single_image(0).get_obstime()
+        origin_of_mask = (int(trj.get_x_pos(time) + 0.5), int(trj.get_y_pos(time) + 0.5))
+        img = stack.get_single_image(i)
```
jeremykubica, line 33: As a very small point. You could switch the order to make things slightly shorter and more efficient by:
- Moving `img = stack.get_single_image(i)` to the front of the loop.
- Computing time as `time = img.get_obstime() - stack.get_single_image(0).get_obstime()`

jeremykubica (changes requested):
The last change broke the test

tests/test_masking.py
vlnistor: Ah yes, my bad. Removed both

src/kbmod/masking.py
vlnistor: Changed. Where do we get efficiency from with the new approach?

jeremykubica: The efficiency would be in getting the reference to the image once instead of twice. To be honest, I don't expect it to really matter at all.

PR #519, opened 2024-03-12
Reorganize the CUDA code

jeremykubica:
Reorganize the CUDA code by moving the CUDA functions into a `kernels` subdirectory (with the appropriate change to `CMakeLists.txt`) and changing how memory management is wrapped for the other C++ functions. Specifically:
- Remove the type specific functions, such as `move_floats_to_gpu`
- Add thin wrapper on the CUDA functions for `void*` pointers only (e.g. `allocate_gpu_block`). Put these in their own file `kernel_memory.cu` with a corresponding header file.
- Add a new templated class for handling types like `float` or `Trajectory`. The `GPUArray` class handles allocating space on the GPU, freeing space on the GPU (including an auto free when the array object is destroyed), and transferring data between the CPU and GPU.


wilsonbb (approved):
This looks reasonable to me, but am not as familiar with CUDA and this part of the code so there may be something subtle I"m missing. Up to you if you want to wait for @maxwest-uw to have a pass at reviewing as well

jeremykubica:
Thanks @wilsonbb. I will have @maxwest-uw take a look as well.

maxwest-uw (approved):

Issue #495, opened 2024-02-29
Incorporate (optional) image normalization into the stamp plotting utils


jeremykubica:
Extend the functions in `src/kbmod/analysis` that plot multiple stamps to optionally perform normalization on the images. The code can be based off (`projects/kbmod/jupyter/notebooks/lincc_demo/plotting.py` internal access only).

jeremykubica:
Replaced by the more comprehensive #523

PR #520, opened 2024-03-14
Adding optional normalize on the `plot_stamps` utility function

vlnistor:
This issue solves #495 

I'm assuming the only utility function we care about is the `plot_stamps` inside `plot_results`. Seems from #259 the `ResultsVisualizer` class is where all relevant plotting of results is being moved to

### DIFF
--- a/src/kbmod/analysis/plot_results.py
+++ b/src/kbmod/analysis/plot_results.py
@@ -1,11 +1,12 @@
 import math
 import matplotlib.pyplot as plt
 import numpy as np
+from astropy.visualization import ZScaleInterval, AsinhStretch, ImageNormalize
 
 
 class ResultsVisualizer:
     @staticmethod
-    def plot_single_stamp(stamp, axes=None):
+    def plot_single_stamp(stamp, axes=None, norm=None):
         """Plot a single stamp image.
 
         Parameters
@@ -14,6 +15,8 @@ class ResultsVisualizer:
             The numpy array containing the stamp's pixel values.
         axes : matplotlib.axes.Axes
             The axes on which to draw the figure.
+        norm : `astropy.visualization.ImageNormalize` or None
+            The normalization to use for the image
         """
         # If there is nothing to plot, skip.
         if stamp is None or stamp.size == 0:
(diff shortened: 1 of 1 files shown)


src/kbmod/analysis/plot_results.py
```diff
@@ -56,7 +63,12 @@ def plot_stamps(stamps, fig=None, columns=3):
 
         for i in range(num_stamps):
             ax = fig.add_subplot(num_rows, columns, i + 1)
-            ResultsVisualizer.plot_single_stamp(stamps[i], axes=ax)
+            norm = (
+                ImageNormalize(stamps[i], interval=ZScaleInterval(), stretch=AsinhStretch())
+                if normalize
+                else None
+            )
```
DinoBektesevic, line 70: Can't this just be 

```
for i, stamp in enumerate(stamps):
    ax = fig.add_subplot(n_rows, n_cols, i+1)
    ax.set_title(f"Time {i}")
    norm = None
    if normalize:
        norm = ImageNormalize(stamp, interval=ZScaleInterval(), stretch=AsinhStretch())
    ResultsVisualizer.plot_single_stamp(stamp, axes=ax, norm=norm)
```

https://github.com/dirac-institute/kbmod/issues/495 doesn't specify whether or not normalization should happen for each image individually or whether they should all be normalized to the same baseline. If it's the latter, then the norm needs to be shared by all the images. 

We could calculate it for the first stamp in the list, or some kind of statistic across the stamps (like a mean, median, min, max or something), we could also let users supply their own `norm` since `plot_single_stamp` allows for it already. Addressing this would also probably fix the awkward `norm` if conditional too. 

This is fine as-is, but we should probably raise this as a point(s) of discussion in https://github.com/dirac-institute/kbmod/issues/495 or https://github.com/dirac-institute/kbmod/issues/523

DinoBektesevic (approved):
Looks fine to me, minor details and some larger structural issues out of scope of this PR.

src/kbmod/analysis/plot_results.py
vlnistor: Changed it to the more readable version you suggested. We can hold off merging this until we address it in either #495 or #523 ? Alternatively, if you prefer to merge as is and then remove later then that also works (but I cannot merge as I'm not a contributor).

DinoBektesevic: I'm ok with merging this now, the core functionality would remain the same, it's really just the func signatures and module reorganization that we need a better plan for I think. I added you to contributors, hopefully this becomes less of a blocker for you now. We can't use the github Runners for the CI because we need GPUs, so we had to deploy our own; please don't mine bitcoins on it.

PR #524, opened 2024-03-14
Add Logging module to the code.

DinoBektesevic:
Following some discussion the issue that was the cause for reverting the original logging module merge seems to be environment inconsistency. Cuda can build packages with a different standard library implementation than the system default. Merging of logging seems to have updated the env packages, or included something not previously required, which resulted in an ABI incompatibility. 

There are many instances of discussions around the internet as how to properly handle this issue (see for example https://github.com/pybind/pybind11/issues/3623), but it seems the issue is that ultimately C/C++ have no way of propagating, validating or enforcing compiler requirements across packages resulting in the fact that one can build two packages in the same environment, that link against 2 different standard library implementations. The culprit seems to have been somewhere in SciPy. 

Consequently, I've removed the reverted logging PR and removed the SciPy patch from https://github.com/dirac-institute/kbmod/pull/506. I'm unable to reproduce the error, so merging of this PR should wait until the environment is fixed.


src/kbmod/__init__.py
```diff
@@ -1,15 +1,18 @@
...
 
 try:
-    from ._version import version as __version__
+    from ._version import version as __version__  # noqa: F401
 except ImportError:
     warnings.warn("Unable to determine the package version. " "This is likely a broken installation.")
 
-# This is needed for compatibility with some older compilers: erfinv needs to be
-# imported before other packages though I'm not sure why.
-from scipy.special import erfinv
```
jeremykubica, line 10: When this line is removed, the tests no longer run on Baldur for me. I get the error:

```
  File "/astro/users/jkubica/kbmod_logs/kbmod/src/kbmod/__init__.py", line 14, in <module>
    from kbmod.search import Logging
ImportError: /astro/users/jkubica/kbmod_logs/kbmod/src/kbmod/search.cpython-312-x86_64-linux-gnu.so: undefined symbol: _ZNSt15__exception_ptr13exception_ptr10_M_releaseEv
```
But as soon as I put it back in all the tests pass fine.  I have no idea why or what is happening . But the PR works fine with the include.

src/kbmod/fake_data/fake_data_creator.py
```diff
@@ -279,7 +280,7 @@ def save_fake_data_to_dir(self, data_dir):
         # Make the subdirectory if needed.
         dir_path = Path(data_dir)
         if not dir_path.is_dir():
-            print("Directory '%s' does not exist. Creating." % data_dir)
+            logger.info(f"Directory {data_dir} does not exist. Creating.")
```
jeremykubica, line 283: Don't we need to include a 
```
logger = kb.Logging.getLogger(__name__)
```
line in this file?

src/kbmod/search/logging.h
```diff
@@ -0,0 +1,236 @@
...
+#endif  // Py_PYTHON_H
+
+// Logging is a singleton keeping the registry of all registered Loggers and
+// their default configuration. Use `getLoger(name)` to get or create a new
+// logger. When called, it will check if the logger exists and return a
+// reference if it does. If it doesn't exists, and the method was called from
+// Python's, it creates a new Python-side logger and returns it. When called
+// from C++, without being able to provide a reference to a, or a name of an
+// already existing, Python logger, it creates a default logger on the C++ side.
+// This logger will share at leas the `logging.basicConfig`-uration with any
```
jeremykubica, line 156:
```suggestion
// This logger will share at least the `logging.basicConfig`-uration with any
```

jeremykubica (commented):
Mostly looks good. One discussion point about an import that is needed to make it build on my config.

src/kbmod/__init__.py
jeremykubica: I followed the chain of dependencies back and it turns out there is an easy fix. If we include `import warnings` the error also goes away. It looks like this was being imported through one of the `__init__.py` files in scipy.

Can you add `import warnings` to the PR in place of this scipy import?

DinoBektesevic:
Ok, we sorted this out so I'm merging.

src/kbmod/__init__.py
wilsonbb: Now that the PR is merged, I'm seeing the same error on baldur. Though need for a rollback since

`from scipy.special import erfinv` in `__init__.py` does seem to fix it for me as well. 

Wondering however if a better solution for the environment issue was found?

PR #525, opened 2024-03-19
Have Standardizers add obstime to LayeredImage

wilsonbb:
Fix both the`ButlerStandardizer` and the`FitsStandardizer` to correctly set the observation time when generating standardized `LayeredImages`. 

Currently, when an `ImageCollection` [creates a `WorkUnit`](https://github.com/dirac-institute/kbmod/blob/f3882ff5e33deef7976eed1598d807df60ce1ed5/src/kbmod/image_collection.py#L519), it uses the `ButlerStandardizer` and the`FitsStandardizer` implementations of `Standardizer.toLayeredImage()` which are not correctly calling `LayeredImage.set_obstime()`
 
Note that for testing the `FitsStandardizer`, we used the same test suite as for the KBMODV1 standardizer since it inherits `FitsStandardizer.toLayerImage`.


src/kbmod/standardizers/butler_standardizer.py
```diff
@@ -331,8 +331,16 @@ def toLayeredImage(self):
...
+        else:
+            mjds = (meta["mjd"] for e in self.processable)
+
         imgs = []
         for sci, var, mask, psf, t in zip(sciences, variances, masks, psfs, mjds):
-            imgs.append(LayeredImage(RawImage(sci), RawImage(var), RawImage(mask), t, psf))
+            # Converts nd array mask from bool to np.float32
+            mask = mask.astype(np.float32)
+            img = LayeredImage(RawImage(sci, t), RawImage(var, t), RawImage(mask, t), psf)
+            img.set_obstime(t)
```
DinoBektesevic, line 344: I thought that `get_obstime` would defer to the `RawImage.get_obstime`, i.e. the `RawImage(sci, t)` already sets the obstime?

src/kbmod/standardizers/butler_standardizer.py
```diff
@@ -331,8 +331,16 @@ def toLayeredImage(self):
...
+        if isinstance(meta["mjd"], (list, tuple)):
+            mjds = meta["mjd"]
+        else:
+            mjds = (meta["mjd"] for e in self.processable)
+
         imgs = []
         for sci, var, mask, psf, t in zip(sciences, variances, masks, psfs, mjds):
-            imgs.append(LayeredImage(RawImage(sci), RawImage(var), RawImage(mask), t, psf))
+            # Converts nd array mask from bool to np.float32
+            mask = mask.astype(np.float32)
```
DinoBektesevic, line 342: Why are we explicitly casting the mask to float32? Is Eigen complaining to you? If so then we should address this in the `standardizeMaskImage` method.

src/kbmod/standardizers/fits_standardizers/fits_standardizer.py
```diff
@@ -441,6 +441,9 @@ def toLayeredImage(self):
         # copy. TODO: fix when/if CPP stuff is fixed.
         imgs = []
         for sci, var, mask, psf, t in zip(sciences, variances, masks, psfs, mjds):
-            imgs.append(LayeredImage(RawImage(sci), RawImage(var), RawImage(mask.astype(np.float32)), psf))
-
+            # Converts nd array mask from bool to np.float32
+            mask = mask.astype(np.float32)
```
DinoBektesevic, line 445: What dtype does Rubin butler return for the mask? I have a feeling I assumed something here and then forgot about it. Mostly astropy.io.fits and the butler should already return float32s. I think Butler might be returning an int for mask, but in that case this cast would still be better in the standardizer I think.

DinoBektesevic (approved):
This will work and since it's my biggest dream ever to get rid of the damned layered and raw image I don't mind a bit of hot-fixes that are in place - just a few questions so that at least I understand what the problem is.

src/kbmod/standardizers/butler_standardizer.py
wilsonbb: Ah good point, I removed the call to `LayeredImage.set_obstime`

src/kbmod/standardizers/butler_standardizer.py
wilsonbb: I'm not quite sure if there is a good reason tbh. I saw that you added it in this ["Fix bug in Fits Standardizers." commit](https://github.com/dirac-institute/kbmod/pull/334/commits/7c44e788718ba7e34b1d38e72dbd0b593071c5b6) but not sure what it specifically was trying to fix

We also currently spoof that the mask should be `np.int32` [here](https://github.com/dirac-institute/kbmod/blob/43218e4e9580e3cf33dfa13dd6eadc2fe0de9fc2/tests/utils/mock_fits.py#L270)

FITS standardizer is fairly general so seems fine to keep the type conversion here to me, but I can look a bit deeper into it if you want

src/kbmod/standardizers/fits_standardizers/fits_standardizer.py
wilsonbb: (This comment seems to be about the butler standardizer but is on the FITS standardizer file? Sorry if I'm the one confused here, but answering about the FITS standardizers)

Yeah  I saw that you added it in this ["Fix bug in Fits Standardizers." commit](https://github.com/dirac-institute/kbmod/pull/334/commits/7c44e788718ba7e34b1d38e72dbd0b593071c5b6) but I am also not sure what it was you were specifically trying to fix. But overall the FITS standardizer seems fairly broad/general so it seems like it could be fine to do a default conversion of the dtype

Related, we also currently spoof that the mask should be `np.int32` [here](https://github.com/dirac-institute/kbmod/blob/43218e4e9580e3cf33dfa13dd6eadc2fe0de9fc2/tests/utils/mock_fits.py#L270) and then in turn explicitly cast the mask dtype to be boolean ion [KBMODV1._standardize_mask()](https://github.com/dirac-institute/kbmod/blob/43218e4e9580e3cf33dfa13dd6eadc2fe0de9fc2/src/kbmod/standardizers/fits_standardizers/kbmodv1.py#L193).

So I guess the overall the question maybe is just what do we want the mask dtype to be since we seem to be a bit inconsistent about it

src/kbmod/standardizers/butler_standardizer.py
wilsonbb: Though actually taking another look, in main we do still still get back to a float32, probably in the conversion to Eigen when we create the `RawImage`

I'm starting to feel that we should remove all the type conversions both in here and in

`standardizeMaskImage`

src/kbmod/standardizers/fits_standardizers/fits_standardizer.py
DinoBektesevic: The array is read into the memory by Astropy based on the header keywords `DTYPE`. It's the same for Rubin. Originally Eigen proposal was to have 3 different arrays (float, double and ints) but that was shut down. This induces a copy and a cast to float, since float is the only array people wanted. There's really no need to have more than int8 for the mask and it would save space on the GPU, but the work to get that done was deemed too much to do at the time. 

So this is true for both Rubin and any of the FITS standardizers. Then I floated the casts up to Python by banning the implicit conversion and casts in the Python bindings in C++:

https://github.com/dirac-institute/kbmod/blob/main/src/kbmod/search/raw_image.cpp#L465

But I seem to have a few unnecessary ones. Rubin seems to return int masks, other files seems to return whatever they want. This can be fixed and reduce memory and time requirements, but it's nothing that should be addressed in this PR so it's all ok. Just one more thing for the TODO list.

PR #529, opened 2024-03-21
add correct_parallax

maxwest-uw:
adds the parallax correction function as a utility and adds a test (values described [here](https://github.com/DinoBektesevic/region_search_example/blob/main/02_accounting_parallax.ipynb).


DinoBektesevic (approved):
Can you add 

```
References
----------

.. [1] `Jupyter Notebook <https://github.com/DinoBektesevic/region_search_example/blob/main/02_accounting_parallax.ipynb>`_
```

Make black ignore the line if it mangles it but it shouldn't.

Issue #356, opened 2023-09-28
Refactor StackSearch to do batching at search() function


jeremykubica:
The `search()` function in `StackSearch` runs the GPU-based search for the entire cross product of starting pixels and `search_list` velocities. Given N x M images and V different velocities, it runs N * M * V searches and stores 8 *  N * M * V  results in a `results` vector. To avoid overloading python, it provides a function to read the results back in batches.

We could reduce memory usage by doing the batching at the search function instead. The search function would run a subset of the searches (maybe a range of starting pixels) and return the results directly. It would not need to store the full result set in memory in either the GPU (during computation) or CPU (after computation). Individual batches could be filtered **before** the next set is run.

In order to do this batching efficiently, we need to do the precomputation (creating the `psi_images` and `phi_images`) once and copy the data to the GPU once. This would require new functions to prepare and clean up the GPU memory (both based off of `deviceSearchFilter`).

jeremykubica:
Closed by #530

PR #530, opened 2024-03-23
Batch search inside stack search

vlnistor:
This tackles #356 

The idea is that you'd replace the full in-memory search:

```
min_observations = ... # Some min observations
candidates = ... # Some trajectories
stack = ImageStack(im_list)
search = StackSearch(stack)
search.search(candidates, min_observations)
```

With a batched version: 

```
 with BatchSearchManager(StackSearch(stack), candidates, min_observations) as batch_search:
    batch_results = []
    for i in range(0, width, 5):
        batch_search.set_start_bounds_x(i, i + 5)
        for j in range(0, height, 5):
            batch_search.set_start_bounds_y(j, j + 5)
            batch_results.extend(batch_search.search_batch())

```

The steps are:
1. Use the `BatchSearchManager` context manager to obtain a managed `batch_search` object
2. Loop through the whole grid by breaking it into smaller subsets

### DIFF
--- a/src/kbmod/batch_search.py
+++ b/src/kbmod/batch_search.py
@@ -1,0 +1,40 @@
+class BatchSearchManager:
+    def __init__(self, stack_search, search_list, min_observations):
+        """
+        Initialize the context manager with an instance of StackSearch,
+        the list of trajectories to search, and the minimum number of observations.
+
+        Parameters:
+        - stack_search: Instance of the StackSearch class.
+        - search_list: List of trajectories to search.
+        - min_observations: Minimum number of observations for the search.
+        """
+        self.stack_search = stack_search
+        self.search_list = search_list
+        self.min_observations = min_observations
+
+    def __enter__(self):
+        """
+        This method is called when entering the context managed by the `with` statement.
+        It prepares the batch search by calling `prepare_batch_search` on the StackSearch instance.
+        """
+        # Initialize or prepare memory for the batch search.
+        self.stack_search.prepare_batch_search(self.search_list, self.min_observations)
+        # Return the object that should be used within the `with` block. Here, it's the StackSearch instance.
+        return self.stack_search
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """
+        This method is called when exiting the context.
+        It cleans up resources by calling `finish_search` on the StackSearch instance.
+
+        Parameters:
+        - exc_type: The exception type if an exception was raised in the `with` block.
+        - exc_value: The exception value if an exception was raised.
+        - traceback: The traceback object if an exception was raised.
+        """
+        # Clean up resources or delete initialized memory.
+        self.stack_search.finish_search()
+        # Returning False means any exception raised within the `with` block will be propagated.
+        # To suppress exceptions, return True.
+        return False
--- a/src/kbmod/search/stack_search.cpp
+++ b/src/kbmod/search/stack_search.cpp
@@ -9,7 +9,7 @@ extern "C" void evaluateTrajectory(PsiPhiArrayMeta psi_phi_meta, void* psi_phi_v
                                    SearchParameters params, Trajectory* candidate);
 #endif
 
-StackSearch::StackSearch(ImageStack& imstack) : stack(imstack), results(0) {
+StackSearch::StackSearch(ImageStack& imstack) : stack(imstack), results(0), gpu_search_list(0) {
     debug_info = false;
     psi_phi_generated = false;
 
@@ -149,6 +149,65 @@ Trajectory StackSearch::search_linear_trajectory(short x, short y, float vx, flo
     return result;
 }
 
+void StackSearch::finish_search(){
+    psi_phi_array.clear_from_gpu();
+    gpu_search_list.move_to_cpu();
+}
+
+void StackSearch::prepare_batch_search(std::vector<Trajectory>& search_list, int min_observations){
+    DebugTimer psi_phi_timer = DebugTimer("Creating psi/phi buffers", debug_info);
+    prepare_psi_phi();
+    psi_phi_array.move_to_gpu();
+    psi_phi_timer.stop();
+
+    
+    int num_to_search = search_list.size();
+    if (debug_info) std::cout << "Preparing to search " << num_to_search << " trajectories... \n" << std::flush;
+    gpu_search_list = TrajectoryList(search_list);
+    gpu_search_list.move_to_gpu();
+
+    params.min_observations = min_observations;
+}
+
+
+std::vector<Trajectory> StackSearch::search_batch(){
+    if(!psi_phi_array.gpu_array_allocated()){
+        throw std::runtime_error("PsiPhiArray array not allocated on GPU. Did you forget to call prepare_search?");
+    }
+
+    DebugTimer core_timer = DebugTimer("Running batch search", debug_info);
+    // Allocate a vector for the results and move it onto the GPU.
+    int search_width = params.x_start_max - params.x_start_min;
+    int search_height = params.y_start_max - params.y_start_min;
+    int num_search_pixels = search_width * search_height;
+    int max_results = num_search_pixels * RESULTS_PER_PIXEL;
+
+    if (debug_info) {
+        std::cout << "Searching X=[" << params.x_start_min << ", " << params.x_start_max << "]"
+                  << " Y=[" << params.y_start_min << ", " << params.y_start_max << "]\n";
+        std::cout << "Allocating space for " << max_results << " results.\n";
+    }
+    results = TrajectoryList(max_results);
+    results.move_to_gpu();
+
+        // Do the actual search on the GPU.
+    DebugTimer search_timer = DebugTimer("Running search", debug_info);
+#ifdef HAVE_CUDA
+    deviceSearchFilter(psi_phi_array, params, gpu_search_list, results);
+#else
+    throw std::runtime_error("Non-GPU search is not implemented.");
+#endif
+    search_timer.stop();
+
+    results.move_to_cpu();
+    DebugTimer sort_timer = DebugTimer("Sorting results", debug_info);
+    results.sort_by_likelihood();
+    sort_timer.stop();
+    core_timer.stop();
+
+    return results.get_batch(0, max_results);
+}
+
 void StackSearch::search(std::vector<Trajectory>& search_list, int min_observations) {
     DebugTimer core_timer = DebugTimer("Running core search", debug_info);
 
@@ -173,7 +232,7 @@ void StackSearch::search(std::vector<Trajectory>& search_list, int min_observati
     // Allocate space for the search list and move that to the GPU.
     int num_to_search = search_list.size();
     if (debug_info) std::cout << "Searching " << num_to_search << " trajectories... \n" << std::flush;
-    TrajectoryList gpu_search_list = TrajectoryList(search_list);
+    gpu_search_list = TrajectoryList(search_list);
     gpu_search_list.move_to_gpu();
 
     // Set the minimum number of observations.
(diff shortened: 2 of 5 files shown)


vlnistor:
I'd hold off until I fix the merge requests.

src/kbmod/batch_search.py
```diff
@@ -0,0 +1,40 @@
+class BatchSearchManager:
```
vlnistor, line 1: The main idea here was that I didn't want the user to have to call `prepare_batch_search` and `finish_search`

src/kbmod/search/stack_search.cpp
```diff
@@ -155,6 +155,26 @@ Trajectory StackSearch::search_linear_trajectory(short x, short y, float vx, flo
...
+void StackSearch::prepare_batch_search(std::vector<Trajectory>& search_list, int min_observations){
+    DebugTimer psi_phi_timer = DebugTimer("Creating psi/phi buffers", rs_logger);
+    prepare_psi_phi();
+    psi_phi_array.move_to_gpu();
+    psi_phi_timer.stop();
+
+    
+    int num_to_search = search_list.size();
+    if (debug_info) std::cout << "Preparing to search " << num_to_search << " trajectories... \n" << std::flush;
+    gpu_search_list = TrajectoryList(search_list);
```
vlnistor, line 172: This is the only reason why I implemented the move assignment operators. Note sure if there's a neater/better way to not have to do it.

src/kbmod/search/trajectory_list.cpp
```diff
@@ -22,6 +22,18 @@ TrajectoryList::TrajectoryList(int max_list_size) {
     gpu_array.resize(max_size);
 }
 
+// Move assignment operator.
+TrajectoryList& TrajectoryList::operator=(TrajectoryList&& other) noexcept {
+    if (this != &other) {
+        max_size = other.max_size;
+        data_on_gpu = other.data_on_gpu;
+        cpu_list = std::move(other.cpu_list);
+        gpu_array = std::move(other.gpu_array);
```
jeremykubica, line 31: I'm not sure this does that correct thing in terms of pointer deallocation. Shouldn't we need to deallocate the current gpu_array?

src/kbmod/search/stack_search.cpp
jeremykubica: Couldn't we just do `gpu_search_list.set_trajectories(search_list)` here? `set_trajectories()` should take care of resizing the memory allocations and copying over the new data.

src/kbmod/search/stack_search.h
```diff
@@ -82,6 +85,9 @@ class StackSearch {
 
     // Results from the grid search.
     TrajectoryList results;
+
+    // Trajectories that are being searched.
+    TrajectoryList gpu_search_list;
```
jeremykubica, line 90: We could expand this object variable to be used in other places by creating a helper function `set_search_list()` that sets `gpu_search_list` from a given vector. We could then use that function in both `prepare_batch_search()` and `search()` (instead of allocating a local `gpu_search_list` within the function). I think this would be cleaner from a naming perspective.

src/kbmod/search/stack_search.h
```diff
@@ -50,6 +50,9 @@ class StackSearch {
     // The primary search functions
     void evaluate_single_trajectory(Trajectory& trj);
     Trajectory search_linear_trajectory(short x, short y, float vx, float vy);
+    void prepare_batch_search(std::vector<Trajectory>& search_list, int min_observations);
+    void finish_search();
+    std::vector<Trajectory> search_batch();
```
jeremykubica, line 55: I think you can combine a bunch of duplicate code in `search()` and `search_batch()`.  We keep the common code in the function `search_batch()` by do not have it return the trajectories. We then use two outer functions:

`search_all()`:

1. Call `prepare_batch_search()` (renamed to something like `prepare_search()`)
2. Call `search_batch()`
3. Call `finish_search()`

`search_single_batch()`
1. Call `search_batch()`
2. Return `results.get_batch(0, max_results);`

src/kbmod/search/trajectory_list.h
```diff
@@ -29,6 +29,7 @@ class TrajectoryList {
     TrajectoryList(const TrajectoryList&) = delete;
     TrajectoryList& operator=(TrajectoryList&) = delete;
     TrajectoryList& operator=(const TrajectoryList&) = delete;
+    TrajectoryList& operator=(TrajectoryList&&) noexcept;
```
jeremykubica, line 32: As I noted in another comment, I think it would be cleaner to just use `set_trajectories` instead of the move operator.

tests/test_search.py
```diff
@@ -932,6 +933,73 @@ def test_coadd_filter_gpu(self):
...
+        psf_data[1:4, 1:4] = 0.1111111
+        p = PSF(psf_data)
+
+        # Create a stack with 10 20x20 images with random noise and times ranging from 0 to 1
+        count = 10
+        imlist = [make_fake_layered_image(width, height, 5.0, 25.0, n / count, p) for n in range(count)]
+        stack = ImageStack(imlist)
+        im_list = stack.get_images()
+        # Create a new list of LayeredImages with the added object.
+        new_im_list = []
```
jeremykubica, line 957: You don't need to create a new `ImageStack`. You should be able to insert the object directly into `stack`:

```
for i in range(count):
    im = stack.get_single_image(i)
    time = stack.get_zeroed_time(i)
    add_fake_object(im, 5.0 + (time * 8.0), 35.0 + (time * 0.0), 25000.0)
```

src/kbmod/batch_search.py
jeremykubica: Not for this PR, but something to consider as a future extension: In theory this could be extended further by making this a generator:

```
for batch_search in BatchSearchManager(stack_search, search_list, min_observations, x_size, y_size):
    batch_results.extend(batch_search.search_batch())
```

where the `BatchSearchManager` tracks the block sizes to search, current block locations, etc.

jeremykubica (commented):
A bunch of points for discussion.

src/kbmod/search/stack_search.h
vlnistor: I don't see that many occurrences of `StackSearch::search` across the project (a handful in some notebooks and the main one in `run_search.py`. Are you thinking we should rename all of them to `search_all` in this paradigm?

src/kbmod/search/stack_search.cpp
vlnistor: Yeah - that's cleaner and less error prone. Adapted.

src/kbmod/batch_search.py
```diff
@@ -0,0 +1,37 @@
...
+    def __init__(self, stack_search, search_list, min_observations):
+        """
+        Initialize the context manager with an instance of StackSearch,
+        the list of trajectories to search, and the minimum number of observations.
+
+        Parameters:
```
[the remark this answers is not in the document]
vlnistor: Changed

src/kbmod/search/trajectory_list.cpp
vlnistor: Yep. That would've been a memory leak. Using `set_trajectories()` should completely avoid this

src/kbmod/search/trajectory_list.h
vlnistor: Removed in favour of `set_trajectories`

tests/test_search.py
vlnistor: Changed

src/kbmod/search/stack_search.h
vlnistor: I think with the new paradigm of `search_all` this is less of a problem.  In terms of having a `set_search_list()` I worry that if it's not explicitly asked for in `prepare_search` and `search_all` you'll find yourself calling them before setting the search list. We can throw an error in that case, but I think it might be a little more intuitive for it to be explicitly asked for in the functions. I don't feel particularly strong though.

vlnistor: I think with the new paradigm of `search_all` this is less of a problem.  In terms of having a `set_search_list()` I worry that if it's not explicitly asked for in `prepare_search` and `search_all` you'll find yourself calling them before setting the search list. We can throw an error in that case, but I think it might be a little more intuitive for it to be explicitly asked for in the functions. I don't feel particularly strong though.

src/kbmod/search/stack_search.h
jeremykubica: I was thinking that, but perhaps this entire suggestion is over engineering.

src/kbmod/search/stack_search.h
```diff
@@ -29,6 +29,7 @@ using Point = indexing::Point;
 using Image = search::Image;
 
 class StackSearch {
+    int extract_max_results();
```
jeremykubica, line 32: Maybe rename this to `get_max_results` or `compute_max_results` to avoid any confusion about this somehow extracting the results from the cached list.

src/kbmod/search/stack_search.cpp
```diff
@@ -288,7 +306,10 @@ static void stack_search_bindings(py::module& m) {
             .def("prepare_psi_phi", &ks::prepare_psi_phi, pydocs::DOC_StackSearch_prepare_psi_phi)
             .def("clear_psi_phi", &ks::clear_psi_phi, pydocs::DOC_StackSearch_clear_psi_phi)
             .def("get_results", &ks::get_results, pydocs::DOC_StackSearch_get_results)
-            .def("set_results", &ks::set_results, pydocs::DOC_StackSearch_set_results);
+            .def("set_results", &ks::set_results, pydocs::DOC_StackSearch_set_results)
+            .def("search_single_batch", &ks::search_single_batch, pydocs::DOC_StackSearch_search_single_batch)
+            .def("prepare_search", &ks::prepare_search, pydocs::DOC_StackSearch_prepare_batch_search)
+            .def("finish_search", &ks::finish_search, pydocs::DOC_StackSearch_finish_search);
```
jeremykubica, line 312: I think it could be worthwhile to include the (renamed) `extract_max_results` in the python interface for use in testing.

src/kbmod/search/stack_search.cpp
```diff
@@ -155,63 +155,81 @@ Trajectory StackSearch::search_linear_trajectory(short x, short y, float vx, flo
...
-
-    TrajectoryList gpu_search_list(search_list);
-    gpu_search_list.move_to_gpu();
-
-    // Set the minimum number of observations.
-    params.min_observations = min_observations;
-
-    // Do the actual search on the GPU.
-    DebugTimer search_timer = DebugTimer("search execution", rs_logger);
+        // Do the actual search on the GPU.
```
jeremykubica, line 203: Indentation of this comment appears off.

src/kbmod/search/stack_search.h
jeremykubica: Sounds good. Let's keep as-is.

jeremykubica (commented):
This is looking really good. A few stylistic cleanups.

tests/test_search.py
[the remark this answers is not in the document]
vlnistor: Removed now

src/kbmod/search/stack_search.h
vlnistor: Good point - renamed

src/kbmod/search/stack_search.cpp
vlnistor: I exposed it but stopped short of writing a test for it because there seems that `RESULTS_PER_PIXEL` from `common.h` is not being exposed anywhere. I think it would be valuable to have some python binding that shows these settings

I see that the test I wrote for the batch search assumes 8 results per pixel but it would be nicer if it could extract it straight from a `Config` object that binds those settings from the internal cpp headers. I can do a follow-up PR if you think that makes sense.

src/kbmod/search/stack_search.cpp
vlnistor: Cleaned

jeremykubica (approved):

PR #532, opened 2024-04-03
Make the trajectory prediction logic more centralized

jeremykubica:
Predicting the trajectory's position is done manually throughout a lot of the code as `trj.x + dt * trj.vx + 0.5`. The 0.5 pixel offset is applied most of the time, but not everywhere (e.g. in computing the stamps). This PR centralizes the prediction logic into `common.h`. 

At the same time it makes the 0.5 the standard default throughout the code. This should lead to long term better performance, but requires some changes to tests (for stamps). This also leads to a *increase* in the number of results found by the diff test.


src/kbmod/search/common.h
```diff
@@ -53,9 +53,17 @@ struct Trajectory {
...
+    // the prediction starts from the center of the pixel (which it does in the search)
+    inline float get_x_pos(float time, bool centered = true) const {
+        return centered ? (x + time * vx + 0.5f) : (x + time * vx);
+    }
+    inline float get_y_pos(float time, bool centered = true) const {
+        return centered ? (y + time * vy + 0.5f) : (y + time * vy);
+    }
+
+    inline int get_x_index(float time) const { return (int)floor(get_x_pos(time, true)); }
+    inline int get_y_index(float time) const { return (int)floor(get_y_pos(time, true)); }
```
DinoBektesevic, line 66: I wonder if this deserves a `get_pos` that returns an `Index` or an `Point` and then snaps them to the respective grids. (For some reason this feels like it was done at one point but there are no traces of this in the code at all?)

On the GPU we have a separate `predict_index` anyhow and we should use our own inventions for consistency at least.

DinoBektesevic (approved):
Minor question regarding not using `Index` vs `Point` otherwise this is fine with me with or without the filtering or detection working differently.

Issue #535, opened 2024-04-05
Rewrite ResultList as a table


jeremykubica:
Rewrite the `ResultList` as a table so we work more dynamically with it and take advantage of vectorized operations.

jeremykubica:
Closed by #563

PR #536, opened 2024-04-05
Add some helpers for working with result tables

jeremykubica:
This is a temporary fix to enable more dynamic exploration until we address #535. It adds an index column to the table resulting from `ResultList.to_table()` which can be used to map back to the result's index in the original `ResultList`. It also adds a helper function to keep the two in sync (which is a bit of a hack and should be removed by a more proper approach like #535).

Adds examples of how to use the new functionality to the results notebook.


src/kbmod/result_list.py
```diff
@@ -890,6 +890,20 @@ def revert_filter(self, label=None):
...
+        appear in the table are removed from the list. The indices in the table
+        are updated to match the new ordering in the result list.
+
+        Parameters
+        ----------
+        table : `astropy.table.Table`
+            A table with the data as generated by to_table().
+        """
+        self.filter_results(table["index"], "Table filtered")
+        table["index"] = range(len(table))
```
wilsonbb, line 905: Just a comment for when we actually address https://github.com/dirac-institute/kbmod/issues/535 : I'm presuming we want to actually set an [astropy.table index](https://docs.astropy.org/en/stable/table/indexing.html#creating-an-index), but curious if there are any specific reasons why we were choosing not to do so (for example extra performance cost not worth it for hopefully small ResultLists)

wilsonbb (approved):
LGTM

PR #559, opened 2024-04-16
adds `invert_correct_parallax` function

### LINKED ISSUE #557: Make `correct_parallax` output a `SkyCoord`
right now it outputs as an "ICRS Coordinate", which has some but not all of the SkyCoord functionality. we should convert it to SkyCoord before we return it to the user.

(thanks to colin for calling this out).

### LINKED ISSUE #558: add `invert_correct_parallax` functionality
create a `reprojection_utils` function that can take in a parallax corrected coordinate and return the position in the original ICRS frame of reference.

maxwest-uw:
resolves #557 , resolves #558 

- creates the `invert_correct_parallax` function, which can take a coordinates in EBD space and returns the position in the original "uncorrected" icrs reference frame.
- also fixes an issue where `correct_parallax` would output as something other than a `SkyCoord`.
- also changes some of the naming around the different "distance" variables (now `heliocentric_distance` and `geocentric_distance`, respectively) to improve readability.

### DIFF
--- a/src/kbmod/reprojection_utils.py
+++ b/src/kbmod/reprojection_utils.py
@@ -82,6 +82,8 @@ def invert_correct_parallax(coord, obstime, point_on_earth, geocentric_distance,
         The location on Earth of the observation.
     geocentric_distance : `float`
         The distance from Earth to the object (generally a result from `correct_parallax`).
+    heliocentric_distance : `float`
+        The distance from the solar system barycenter to the object (generally an input for `correct_parallax`).
 
     Returns
     ----------
@@ -138,7 +140,8 @@ def fit_barycentric_wcs(
     Returns
     ----------
     An `astropy.wcs.WCS` representing the original image in "Explicity Barycentric Distance" (EBD)
-    space, i.e. where the points have been corrected for parallax.
+    space, i.e. where the points have been corrected for parallax, as well as the average best fit
+    geocentric distance of the object.
     """
     rng = np.random.default_rng(seed)
 
--- a/src/kbmod/reprojection_utils.py
+++ b/src/kbmod/reprojection_utils.py
@@ -6,7 +6,7 @@ from astropy.wcs.utils import fit_wcs_from_points
 from scipy.optimize import minimize
 
 
-def correct_parallax(coord, obstime, point_on_earth, guess_distance):
+def correct_parallax(coord, obstime, point_on_earth, heliocentric_distance):
     """Calculate the parallax corrected postions for a given object at a given time and distance from Earth.
 
     Attributes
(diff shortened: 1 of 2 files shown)


src/kbmod/reprojection_utils.py
```diff
@@ -38,28 +38,79 @@ def correct_parallax(coord, obstime, point_on_earth, guess_distance):
...
+    return answer, fit.x[0]
+
+
+def invert_correct_parallax(coord, obstime, point_on_earth, geocentric_distance, heliocentric_distance):
+    """Calculate the original ICRS coordinates of a point in EBD space, i.e. a result from `correct_parallax`.
+
+    Attributes
+    ----------
+    coord : `astropy.coordinate.SkyCoord`
+        The coordinate to be corrected for.
```
jeremykubica, line 78: I'm not really sure what "coordinate to be corrected for" means. Is that the input coordinate that is being modified? Or is that the pointing of the original image? Or is it something else?

src/kbmod/reprojection_utils.py
```diff
@@ -38,28 +38,79 @@ def correct_parallax(coord, obstime, point_on_earth, guess_distance):
...
+
+    References
+    ----------
+    .. [1] `Jupyter Notebook <https://github.com/maxwest-uw/notebooks/blob/main/uncorrecting_parallax.ipynb>`_
+    """
+    loc = (
+        point_on_earth.x.to(u.m).value,
+        point_on_earth.y.to(u.m).value,
+        point_on_earth.z.to(u.m).value,
+    ) * u.m
```
jeremykubica, line 100: Do you need to do the `.value` and `* u.m`? Since you are using `.to(u.m)` the values might already have the correct units appended.

src/kbmod/reprojection_utils.py
```diff
@@ -38,28 +38,79 @@ def correct_parallax(coord, obstime, point_on_earth, guess_distance):
...
-    return answer
+    return answer, fit.x[0]
+
+
+def invert_correct_parallax(coord, obstime, point_on_earth, geocentric_distance, heliocentric_distance):
+    """Calculate the original ICRS coordinates of a point in EBD space, i.e. a result from `correct_parallax`.
+
+    Attributes
+    ----------
+    coord : `astropy.coordinate.SkyCoord`
```
jeremykubica, line 77: Not for this PR, but something to consider for the future: I wonder how difficult it would be to vectorize this function.

tests/test_reprojection_utils.py
```diff
@@ -53,16 +53,58 @@ def test_parallax_equinox(self):
...
+            coord=self.sc2,
+            obstime=self.icrs_time2,
+            point_on_earth=self.eq_loc,
+            heliocentric_distance=50.0,
         )
 
         npt.assert_almost_equal(corrected_coord2.ra.value, expected_ra)
         npt.assert_almost_equal(corrected_coord2.dec.value, expected_dec)
 
+        assert type(corrected_coord1) == SkyCoord
```
jeremykubica, line 66: I think the Python style is generally `type(variable) is something` instead of `==`

jeremykubica (approved):
All small stylistic changes

src/kbmod/reprojection_utils.py
maxwest-uw: added some more clarification, it's an EBD coord that we're trying to find the original uncorrected position off

PR #561, opened 2024-04-18
Remove the BatchFilter abstract data type

jeremykubica:
The `BatchFilter` abstract data type is only ever used for clustering and is wrapped in the function `apply_clustering()`. To keep the code simple, this removes the `BatchFilter` abstract data type and `ResultList`'s `apply_batch_filter()` method and instead uses the `ResultList`'s `filter_results()` method directly.

This also simplifies some of the logic with using other result data types that I have been experimenting with in #549


wilsonbb (approved):
LGTM!

PR #562, opened 2024-04-22
make_trajectory doesn't fail when reading as floats

jeremykubica:
Sometimes floats can be passed into `make_trajectory` if it is ready data from a file, astropy Table, etc.


jeremykubica:
Reverting this based on comments from Dino in the another PR.

PR #563, opened 2024-04-22
Replace the ResultsList data structure with the Results

jeremykubica:
This is the second half of the change to switch from `ResultsList` to `Results`. The main part of this change is replacing `ResultList` in `run_search.py` and`trajectory_explorer.py`.

At the same time, using `Results` simplified the logic throughout the code base and lets us remove some now unneeded code, including:
- `ResultList` and `ResultRow` (all of `result_list.py`)
- The per-row filters including the base class, stats filters (now handled by masking), and stamp filters (never used as a row-filter)
- A bunch of individual results files used for testing and demos that are no longer needed.

The PR also updates the tests and notebooks (which is where the bulk of the diffs come from).


DinoBektesevic (approved):
The satisfaction scrolling through all these deleted files and locs...

PR #566, opened 2024-04-23
Simplifying dataId handling for portability, plus other fixes

ColinOrionChandler:
The previous dataframe was not portable (i.e., would not open without the Stack/Butler, repo all loaded). Changing the primary dataId column to be the simpler dicts, and a new dataId_str column that has a simplified string for the purpose of lookups. Adjusted code throughout the notebook to accommodate these changes.

Other fixes/improvements:
- Duplicate checking function now raises instead of asserting.
- New clean dataId function that takes various forms of a dataId (e.g., Butler object, tuple, dict) and uniformly returns a properly formatted dictionary.
- New dataId_to_dataIdStr() function for the new hashable dataframe column
- removed hard-coded "calexp" requirements, now either user-supplied or (default) uses the first of the desired_datasetTypes list.
- unified lookup tables (removed duplicate generation)
- other minor fixes

### DIFF
--- a/notebooks/region_search/Region Searching Workbook.ipynb
+++ b/notebooks/region_search/Region Searching Workbook.ipynb
@@ -61,6 +61,7 @@
     "from matplotlib import pyplot as plt\n",
     "from matplotlib.cm import get_cmap\n",
     "from matplotlib.colors import Normalize\n",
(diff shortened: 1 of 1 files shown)


wilsonbb (approved):
Thanks Colin!

PR #571, opened 2024-04-26
Provide an option to dynamically change the logging level

jeremykubica:
Allowing the C++ logger to dynamically change the logging level for debugging purposed (where we want to enable debug logging for only part of the run). This came up while using a notebook for debugging.

### DIFF
--- a/src/kbmod/search/logging.h
+++ b/src/kbmod/search/logging.h
@@ -77,6 +77,12 @@ public:
 
     virtual ~Logger() {}
 
+    // Change the threshold level for this Logger.
+    void set_level(LogLevel new_threshold) {
+        level_threshold = new_threshold;
+        config["level"] = LogLevelToString[new_threshold];
+    }
+
     std::string fmt_time() {
         std::time_t now = std::time(nullptr);
         std::tm timeinfo;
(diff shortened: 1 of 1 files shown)


DinoBektesevic:
Is this not handled by the Python's logging?
KBMOD Logger provides the dumbest default implementation for its own logging, but otherwise is a wrapper for the python objects accessible from C++. 

Setting 

```
logging.basicConfig(level=logging.DEBUG)
logging.getLogger("kbmod").setLevel(logging.DEBUG) # it wasn't a good idea to make logger called kbmod
```

and then resetting

```
logging.getLogger("kbmod").setLevel(logging.ERROR)
```
The point was to have a logger that integrates with python logging - so that tests can report logs at the end of the tests. Unittests don't but I've for a while wanted to just replace it with pytest for a prettier output (in a lot of ways it's the default really over unittest).
Anything that changes the default Python (global and shared I want to add) logging settings affects our logging settings. 

The default stdout loggers are there just to be able to keep that code compile-able without python present. I used to iterate on C++ code faster like that, which I think is an ability we might have lost along the way for some probably good reasons. 

This seems to have worked in the example notebook on baldur: `kbmod/jupyter/notebooks/hackday/simulated_data.ipynb` cell 15
Did this stop working?

There is an annoying problem in that I named the logger "kbmod" - and I should have just left it default. I didn't realize how config inheritance in the default Python logging works at the time:
https://github.com/dirac-institute/kbmod/blob/main/src/kbmod/__init__.py#L57

jeremykubica:
Apologies. I missed that line in the notebook (or rather saw it and thought it was only setting the Python portions of the code). So that does what I need it to do.

PR #587, opened 2024-05-09
Productionize filtering and improve its debugging

jeremykubica:
This PR is a series of changes to filtering that is mean to clean up the code, harden it against edge cases, and improve debugging through better logging. Changes include:
- More comprehensive logging throughout the filtering process.
- Skipping filtering when there are no results (in the sigma-g filtering and stamp filtering)
- Remove deprecated `ResultList` path for `append_all_stamps()` function
- Add basic statistics tracking to `Results` filtering to help with debugging. This is used even when track filtering is turned off.
- Simplify `Results.from_trajectories()` to just use the filter method instead of doing it manually.
- Remove `Results._append_filtered()` method which was only used in one place and merge logic into `Results.filter_rows()`
- Add filtering labels in `run_search.py`'s loading step so we can track them. Also breaks filtering into its two components for finer grained tracking.
- Add more testing

### DIFF
--- a/src/kbmod/filters/clustering_filters.py
+++ b/src/kbmod/filters/clustering_filters.py
@@ -261,6 +261,7 @@ def apply_clustering(result_data, cluster_params):
 
     # Skip clustering if there is nothing to cluster.
     if len(result_data) == 0:
+        logger.info("Clustering : skipping, no results.")
         return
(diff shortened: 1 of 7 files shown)


wilsonbb (approved):
Thanks, LGTM!

PR #588, opened 2024-05-09
Add more debugging info for KBMOD

jeremykubica:
Adds some functions for getting statistics about the input data that I found myself computing by hand in notebooks, including:
- The mean and standard deviation of values in an image
- The fraction of pixels that are masked within an image
As well as a function to dump this information (and more) to a logger.

This will probably only be used during manual debugging.


wilsonbb (approved):
Thanks!

PR #592, opened 2024-05-10
use reproject_adaptive as default algorithm

maxwest-uw:
+ changing some values in the test to reflect the new algo


wilsonbb (approved):
Awesome thanks!

PR #595, opened 2024-05-11
Helio reprojection work unit

maxwest-uw:


src/kbmod/work_unit.py
```diff
@@ -183,18 +227,22 @@ def from_fits(cls, filename):
 
             # Read the size and order information from the primary header.
             num_images = hdul[0].header["NUMIMG"]
-            if len(hdul) != 4 * num_images + 3:
+            n_constituents = hdul[0].header["NCON"]
+            if len(hdul) != (4 * num_images) + (2 * n_constituents) + 3:
                 raise ValueError(
                     f"WorkUnit wrong number of extensions. Expected "
-                    f"{4 * num_images + 3}. Found {len(hdul)}."
+                    f"{6 * num_images + 3}. Found {len(hdul)}."
```
jeremykubica, line 234: Should this be "4 * num_images + 2 * n_constituents + 3" to match the if statement.

src/kbmod/work_unit.py
```diff
@@ -183,18 +227,22 @@ def from_fits(cls, filename):
...
                 )
 
-            # Read in all the image files.
-            per_image_wcs = []
+            # Misc. reprojection metadata
+            reprojected = hdul[0].header["REPRJCTD"]
+            heliocentric_distance = hdul[0].header["HELIO"]
```
jeremykubica, line 239: Do we want all of these new fields to be required? I'm thinking yes for most of them, but maybe not for the HELIO centric distance or the GEO_{I} distances. If the image has not been reprojected, then I'm not sure it makes sense to read them out of the header.

There are two different ways we could handle this:
1) Make them required if they have been reprojected:
```
if reprojected:
    heliocentric_distance = hdul[0].header["HELIO"]
    ...
else:
    heliocentric_distance = something else
```

or 2) Make each one optional.
```
If "HELIO" in hdul[0].header:
    heliocentric_distance = hdul[0].header["HELIO"]
else:
   ....
```

src/kbmod/work_unit.py
```diff
@@ -205,8 +253,27 @@ def from_fits(cls, filename):
 
                 imgs.append(LayeredImage(sci, var, msk, p))
 
+            per_image_wcs = []
+            per_image_ebd_wcs = []
+            constituent_images = []
+            for i in range(n_constituents):
+                # Extract the per-image WCS if one exists.
+                per_image_wcs.append(extract_wcs_from_hdu_header(hdul[f"WCS_{i}"].header))
+                per_image_ebd_wcs.append(extract_wcs_from_hdu_header(hdul[f"EBD_{i}"].header))
```
jeremykubica, line 262: Same comment as the helio/geo distances in terms of EBD_{i}. Should it be optional for unreprojected images?

src/kbmod/work_unit.py
```diff
@@ -249,8 +316,14 @@ def from_dict(cls, workunit_dict):
         else:
             global_wcs = None
 
+        constituent_images = workunit_dict["constituent_images"]
+        heliocentric_distance = workunit_dict["heliocentric_distance"]
```
jeremykubica, line 320: Should these be optional for an unreprojected workunit (as above)?

src/kbmod/reprojection_utils.py
```diff
@@ -171,3 +171,49 @@ def fit_barycentric_wcs(
...
 
     return ebd_wcs, geocentric_distance
+
+
+def transform_wcses_to_ebd(
+    wcs_list, width, height, heliocentric_distance, obstimes, point_on_earth, npoints=10, seed=None
+):
+    """Transform a set of WCSes (for instance, a `WorkUnit.per_image_wcs`) into EBD space.
+
+    Attributes
```
jeremykubica, line 181: This should be "Parameters" to be consistent with the other doc strings.

src/kbmod/wcs_utils.py
```diff
@@ -195,7 +195,7 @@ def calc_actual_image_fov(wcs):
     Notes
     -----
     While this will work for any WCS, it is intended to be used with
-    tangent plane projections where the reference pixel is on the image.
+    tangent plane projections where the reference pixrrylyel is on the image.
```
jeremykubica, line 198: Is this a typo? I'm not sure what "pixrrylyel" means.

jeremykubica (commented):
A few small comments.

src/kbmod/wcs_utils.py
maxwest-uw: yes my cat must've jumped on my keyboard 😂

src/kbmod/work_unit.py
maxwest-uw: Technically the HELIO and GEO_{i} aren't "required" by the WorkUnit, I just had it output `None` or whatever the fits header equivalent is to keep everything consistent. We could be more explicit about them being optional in `from_fits` tho

src/kbmod/work_unit.py
maxwest-uw: I guess my general preference would be to just be explicit when we create a serialized representation of a `WorkUnit` to just have the attribute in the dict or fits file as a null value as I think it's easier to maintain long term as it requires less state checking about what is and isn't available.

src/kbmod/work_unit.py
jeremykubica: Keeping everything consistent makes sense. Let's go with that.

src/kbmod/work_unit.py
jeremykubica: Sounds good

src/kbmod/work_unit.py
```diff
@@ -37,41 +37,85 @@ class WorkUnit:
     wcs : `astropy.wcs.WCS`
         A global WCS for all images in the WorkUnit. Only exists
         if all images have been projected to same pixel space.
+    constituent_images: `list`
+        A list of strings with the original locations of images used
+        to construct the WorkUnit. This is necessary to maintain as metadata
+        because after reprojection we may stitch multiple images into one.
```
jeremykubica, line 43: High level question (that I should have realized and asked in the previous iteration): What happens to the image stack if multiple images are stitched together? Does the size of the image stack change?

Similarly can the size of the `per_image_wcs` and `per_image_ebd_wcs` lists be different?

From the code it looks like the answers are yes for the size of the image stack, but no for the size of the lists. Is this correct?

maxwest-uw: yes that's correct. Since we can stitch multiple images together and these are stored as one `LayeredImage` in the image stack, we keep the `per_image_wcs` and `per_image_ebd_wcs` around, which matches the amount of `constituent_images` in the `WorkUnit`, but not necessarily the image stack. Then when we need to transform coordinates from the reprojected `WorkUnit` back to the original coordinate space, we can

1. get the position in the reprojected image
2. get the index(es) of the constituent image(s) contained in that image
3. `invert_correct_parallax` on the  coordinate
4. determine the pixel positions in the original image from the `per_image_wcs`
5. in the case where a portion of the image stack has been tiled, we can check against all of the constituent wcses and determine which pixel position was actually in bounds

jeremykubica: How will the backward mapping in step 2 done? Will you need to add list of lists mapping the ebd transformed images back to the corresponding constituent images? Or can that already be done with the current listS?

maxwest-uw: oh right... let me think about this. Yeah we'll need some other way to store which `constituent_images` are in what layer of the `WorkUnit`

src/kbmod/work_unit.py
```diff
@@ -183,30 +236,60 @@ def from_fits(cls, filename):
...
+                sci_hdu = hdul[f"SCI_{i}"]
+                sci = hdu_to_raw_image(sci_hdu)
                 var = hdu_to_raw_image(hdul[f"VAR_{i}"])
                 msk = hdu_to_raw_image(hdul[f"MSK_{i}"])
 
+                # n_indices = sci_hdu.header["NIND"]
+                # sub_indices = []
+                # for j in range(n_indices):
+                #     sub_indices.append(sci_hdu.header[f"IND_{j}"])
+                # per_image_indices.append(sub_indices)
```
jeremykubica, line 264: Should this be uncommented?

jeremykubica (approved):

PR #597, opened 2024-05-13
Combine plotting utilities

### LINKED ISSUE #523: Plotting utilities
I was not aware of before (https://github.com/dirac-institute/kbmod/pull/520 at least) that we had a "ResultsVisualizer" class in analysis utils. I was recently asked to add some plotting utilities that I used in a different internal notebook in https://github.com/dirac-institute/kbmod/pull/514

This is a bit messy, I should have just fixed `ResultsVisualizer`, but I made a mistake. Mainly, it's a class with all of its methods static. This isn't really a Pythonic way of doing things, if there's no state to be tracked it doesn't belong in a class. Modules of functions are just fine when appropriately named, f.e. if we named the file as `result_visualizer` then the functions would in other code read like `result_visualizer.plot_single_stamp` for example. 

I also think that name is too long, just `plotting` is fine. So generally I propose:
* make a `plotting` module
* make static methods of `ResultVisualizer` just functions in that module
* order the keyword arguments to be the same between `ResultsVisualizer` and `plotting` utils from https://github.com/dirac-institute/kbmod/pull/514
* Simplify any methods we can so that at best we have one plotting command in it, then we can use kwargs to pass the many many matplotlib arguments onward and leave the user with the control. 
* try to minimize the number of functions that actually create their own figure and axes, it'd be best if they all could just take a single axis in and spit a modified axis out. That way plotting utils can be composable for the user in many different scenarios.

jeremykubica:
Closes #523 

Combine the two different approaches into a single plotting utilities library. Uses the API of Dino's `plotting.py` and removes the static `ResultsVisualizer` class. It also allows more object types to increase flexibility.

This PR also combines and expands two different visualization notebooks to provide a single location for more comprehensive examples.


wilsonbb (approved):
Nice, thanks!

PR #602, opened 2024-05-14
Add ability to generate multiple coadds.

jeremykubica:
Add the ability for KBMOD to generate multiple coadds (and do so without filtering). This is based on a request from Wes.

By using the `coadds` configuration parameter, the code can generate multiple coadds without filtering. These are put in columns named "coadd_{type}".

Also cleans up the code by removing redundant tests and cleans up some of the debug-level logging.


docs/source/user_manual/search_params.rst
```diff
@@ -39,6 +39,11 @@ This document serves to provide a quick overview of the existing parameters and
 |                        |                             | Must be one of ``all``, ``position``,  |
 |                        |                             | or ``mid_position``.                   |
 +------------------------+-----------------------------+----------------------------------------+
+| ``coadds``             | []                          | A list of additional coadds to create. |
+|                        |                             | These are not used in filtering, but   |
+|                        |                             | saved to columns for analysis. Can     |
+|                        |                             | include: "sum", "mean", and "median".  |
++------------------------+-----------------------------+----------------------------------------+
```
maxwest-uw, line 46: maybe specify that these are for stamp coadds specifically? my initial thought would be that this would be for a coadd somewhere in the actual kbmod search algorithm and I was slightly confused about it at first 😭

maxwest-uw (approved):
1 small document change but otherwise good

docs/source/user_manual/search_params.rst
jeremykubica: You can set the coadd used in the main algorithm using the parameter `stamp_type`. I expanded the comment to clarify this.

PR #612, opened 2024-05-17
Overhaul clustering

### LINKED ISSUE #539: Use a performance aware clustering algorithm
KBMOD currently uses DBSCAN to cluster similar trajectories. These trajectories can be clustered based on:

- Normalized initial position
- Normalized midpoint position
- Normalized initial position + normalized velocity (default).

and returns a single representative trajectory for each cluster. However the choice of representative trajectory does not factor in the "goodness" of the trajectory such as likelihood.

We should explore alternative approaches to clustering that take into account the trajectory's scores. This could be as simple as not clustering but rather:

1. Sort the trajectories by score
2. For each trajectory: Find all its neighbors within a given distance threshold. If there is a neighbor with a higher score, do not include the current trajectory. Otherwise include it.

jeremykubica:
Adds a few new options to KBMOD clustering and significantly updates the documentation. New options include:
- The ability *not* to scale the position clusterings (these are encoded as `_unscaled` clustering types)
- The ability to cluster on a combination of start and end positions

As a general cleanup, the various positional algorithms were combined into a single predicted position clustering algorithm that takes different times. For example the `ClusterPositionFilter` was replaced by using `ClusterPredictionFilter` with time=0 and `ClusterMidPosFilter` by `ClusterPredictionFilter` with the median time. This allows for a single (better tested) class and the easier ability to add new functions (like turning off scaling).

Performed a few additional small clean ups:
- More informative names for the clustering algorithm strings (for logging).
- The best point (according to LH) is chosen from each cluster. This was already happening because results were sorted by LH first, but this makes it explicit in case the ordering of results is changed somewhere else.

Closes #539

### DIFF
--- a/docs/source/user_manual/results_filtering.rst
+++ b/docs/source/user_manual/results_filtering.rst
@@ -41,24 +41,30 @@ Clustering
 
 Clustering is used to combine duplicates found during the initial search. Since each combination of starting pixels and velocity is considered separately, we might see multiple results corresponding to the same true object. For example, if we have an object starting at pixel (10, 15) we might see enough brightness in an adjacent pixel (10, 16) to register a trajectory starting in that location as well.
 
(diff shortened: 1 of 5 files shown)


wilsonbb (approved):
Neat!

PR #616, opened 2024-05-21
Initial demo of matching KBMOD results to known fakes

wilsonbb:
A simple demo notebook that reloads saved KBMOD results (both a results table and a reflex-corrected `WorkUnit`) and searches for results that are near a catalogue of known fakes (which have been similarly reflex-corrected).

In the future, we may want to standardize how we consume and represent catalogues of fakes, but this seemed helpful to share in its current state.

### DIFF
--- a/notebooks/kbmod_search_results_for_fakes.ipynb
+++ b/notebooks/kbmod_search_results_for_fakes.ipynb
@@ -8,7 +8,7 @@
     "  \n",
     "A basic notebook to demonstrate searching results against known fakes in the data.\n",
     "\n",
-    "Is intended to be run using shared data on baldur."
+    "Note that this notebook is intended to be run using shared data on baldur."
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


notebooks/kbmod_search_results_for_fakes.ipynb
```diff
@@ -0,0 +1,287 @@
...
+    "Then we can use astopy's `search_around_sky` to find which KBMOD results are near our known fakes, with a max separation limit of 1 arcsecond"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "result_coords = SkyCoord(\n",
```
jeremykubica, line 141: I think you can just use `results.table["ra_dec"]` here. So the whole thing could be:
```
results.table["ra_dec"] = wu.wcs.pixel_to_world(results["x"], results["y"])
... load the fakes ...
idx1, idx2, sep2dAngle, dist3d = search_around_sky(results.table["ra_dec"], fake_coords, 1 * u.arcsecond)
```

jeremykubica (approved):
Looks good in general. I added some optional places you could vectorize.

PR #618, opened 2024-05-22
Extend documentation

jeremykubica:
Add some more documentation (and bring some up to date) about filtering steps in KBMOD.

### DIFF
--- a/docs/source/user_manual/results_filtering.rst
+++ b/docs/source/user_manual/results_filtering.rst
@@ -9,6 +9,9 @@ Filtering
 
 KBMOD uses two stages of filtering to reduce the number of candidate trajectories. The first stage uses the candidate trajectory's light curve and the second uses the coadded stamp generated from the trajectory's predicted positions.
 
+Clipped SigmaG Filtering
+------------------------
+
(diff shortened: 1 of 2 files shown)


docs/source/user_manual/results_filtering.rst
```diff
@@ -19,9 +22,15 @@ Relevant light curve filtering parameters include:
...
+Stamp Filtering
+---------------
+
+The stamp filtering stage is only applied if the ``do_stamp_filter`` parameter is set to True. This stage by creating a single stamp representing the sum, mean, or median of pixel values for the stamps at each time step. The stamp type is defined by the ``stamp_type`` parameter and can take on values ``median``, ``mean``, or ``sum``. All of the stamp types drop masked pixels from their computations. The mean and median sums are computed over only the valid time steps from the light curve filtering phase (dropping stamps with outlier fluxes). The sum coadd uses all the time steps regardless of the first phase of filtering.
```
drewoldag, line 28: I think that the beginning of the second sentence is suppose to be "This stage creates ...". Not 100% sure.
```suggestion
The stamp filtering stage is only applied if the ``do_stamp_filter`` parameter is set to True. This stage by creates a single stamp representing the sum, mean, or median of pixel values for the stamps at each time step. The stamp type is defined by the ``stamp_type`` parameter and can take on values ``median``, ``mean``, or ``sum``. All of the stamp types drop masked pixels from their computations. The mean and median sums are computed over only the valid time steps from the light curve filtering phase (dropping stamps with outlier fluxes). The sum coadd uses all the time steps regardless of the first phase of filtering.
```

drewoldag (approved):
This seem very nice.

PR #621, opened 2024-05-23
Provide ability to configure the trajectory generator used

jeremykubica:
Provide ability to configure the trajectory generator used. The generator type and configuration parameters are read from a new configuration field `generator_config`, but the code falls back to current behavior if that is not provided (to prevent breaking any current workflows).

To make things easier also creates a mechanism for registering `TrajectoryGenerator` class names to objects, so they can be programmatically created and creates a new `KBMODV1SearchConfig` generator.

Expand the documentation to include all the new generators and their parameters.


wilsonbb (approved):
Looks good to me, just a small question I am confused about

docs/source/user_manual/search_space.rst
```diff
@@ -27,10 +27,61 @@ KBMOD provides the ability to expand or contract the range of starting pixels us
...
+* ``VelocityGridSearch`` - An evenly spaced grid of x and y velocities
+* ``KBMODV1SearchConfig`` - An evenly spaced grid of velocity magnitudes and angles (this was the only option in v1.1 and before)
+* ``RandomVelocitySearch`` - Randomly sampled x and y velocities
+Additional search strategies can be defined by overriding the ``TrajectoryGenerator`` class in trajectory_generator.py.
+
+The search is selected and configured by a single ``generator_config`` parameter that takes a dictionary. The dictionary must contain a ``name`` entry that matches one of the options above. For example to search a single trajectory with a given ``vx = 1.0`` and ``vy = -2.0``, we would use::
+
+    generator_config = { "name": "SingleVelocitySearch", "vx": 1.0, "vy": -2.0 }
+
+If no generator_config is provided, then KBMOD uses the ``KBMODV1SearchConfig`` search strategy and pulls the configuration parameters from the top level. In this case if ``average_angle`` is not specified, it is estimated from the image's WCS.
```
wilsonbb, line 44: "In this case if ``average_angle`` is not specified, it is estimated from the image's WCS."

Is this actually the case? From below it looks like we raise a ValueError instead. Sorry if I got confused here, but hoping this can be corrected clarified

PR #623, opened 2024-05-23
add filter bypass option to correct_parallax

### LINKED ISSUE #614: Need a way to bypass the fitter for correct parallax
A great utility would be to bypass the fitter by supplying the known "guess" distance. E.g., to work with fakes, where we have the "guess" for those chips cached.

maxwest-uw:
resolves #614 

- adds the `geocentric_distance` parameter to `kbmod.reprojection_utils.correct_parallax`, which when set will use the provided value instead of getting the geocentric distance from the minimizer.
- also makes the `loc` casting a little cleaner for both `correct_parallax` and `invert_correct_parallax`
- also adds a test skip for a test that has a GPU call.

### DIFF
--- a/src/kbmod/reprojection_utils.py
+++ b/src/kbmod/reprojection_utils.py
@@ -6,7 +6,7 @@ from astropy.wcs.utils import fit_wcs_from_points
 from scipy.optimize import minimize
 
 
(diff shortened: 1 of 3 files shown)


ColinOrionChandler (approved):
Looks great!

PR #625, opened 2024-05-24
Implementing parallelization in reprojection.py.

drewoldag:
For the time being I didn't want to completely remove the existing reprojection code. I have renamed the existing `reproject_work_unit` function to `_reproject_work_unit` so that we can use the original name with a default flag that will dispatch to new parallelized code, without having to make any modifications to any existing code. 

Note - To make use of concurrent.futures.ProcessPoolExecutors, all of the input and output variables need to be pickle-able. Unfortunately, many of the kbmod.search objects, such as WorkUnit are not. Thus a lot of the logic in this change involved unwrapping the images down to "simple" numpy arrays before passing them to astropys reprojection function, or re-wrapping numpy array representations of images back into WorkUnits.


drewoldag:
Looks like there are already some merge conflicts, so perhaps I'll either need to incorporate those changes into the parallelized code or abandon the use of the dispatcher function. I'll tackle that tomorrow :)

src/kbmod/reprojection.py
```diff
@@ -1,16 +1,19 @@
...
+import concurrent.futures
 import reproject
 from astropy.nddata import CCDData
 from astropy.wcs import WCS
 
 from kbmod.search import KB_NO_DATA, PSF, ImageStack, LayeredImage, RawImage
 from kbmod.work_unit import WorkUnit
 
+# The number of executors to use in the parallel reprojecting function.
+NUM_EXECUTORS = 8
```
wilsonbb, line 11: A quick drive-by comment, but would it be possible to make the number of executors an optional param to the function?

jeremykubica: Agreed that this would be better as an optional parameter.

src/kbmod/reprojection.py
```diff
@@ -31,7 +34,59 @@ def reproject_raw_image(image, original_wcs, common_wcs, obs_time):
         for footprint[i][j], it's 1 if there is a corresponding reprojected
         pixel and 0 if there is no data.
     """
-    image_data = CCDData(image.image, unit="adu")
+    return _reproject_image(image.image, original_wcs, common_wcs)
+
+
+def reproject_ndarray_image(image, original_wcs, common_wcs):
```
jeremykubica, line 40: Optional thought: These two functions could be collapsed into one by doing something like:
```
if type(image) is RawImage:
    image = image.image
```

src/kbmod/reprojection.py
```diff
@@ -132,3 +211,195 @@ def reproject_work_unit(work_unit, common_wcs):
...
+        # append timestamps and layeredImages to lists
+        time_stamps.append(time)
+        image_list.append(new_layered_image)
+
+    # sort the time_stamps, use the ordering to sort image_list
+    image_list = [image_list[i] for i in np.argsort(time_stamps)]
+    stack = ImageStack(image_list)
+
+    # Add the imageStack to a new WorkUnit and return it.
+    new_wunit = WorkUnit(im_stack=stack, config=work_unit.config, wcs=common_wcs)
```
jeremykubica, line 298: I defer to Max here, but I think we also want to keep around the list of old WCS information.

src/kbmod/reprojection.py
```diff
@@ -156,3 +197,231 @@ def reproject_work_unit(work_unit, common_wcs, frame="original"):
...
+
+    # Add the imageStack to a new WorkUnit and return it.
+    new_wunit = WorkUnit(
+        im_stack=stack,
+        config=work_unit.config,
+        wcs=common_wcs,
+        constituent_images=work_unit.constituent_images,
+        per_image_wcs=work_unit._per_image_wcs,
+        per_image_ebd_wcs=work_unit._per_image_ebd_wcs,
+        per_image_indices=unique_obstimes_indices,
```
maxwest-uw, line 303: You should also set `reprojected=True` as part of this new `WorkUnit`

maxwest-uw: (realizing this was something I forgot to do originally haha)

maxwest-uw (commented):
a few small things, overall looking good!!

src/kbmod/reprojection.py
```diff
@@ -41,7 +47,42 @@ def reproject_raw_image(image, original_wcs, common_wcs, obs_time):
...
+    work_unit : `kbmod.WorkUnit`
+        The WorkUnit to be reprojected.
+    common_wcs : `astropy.wcs.WCS`
+        The WCS to reproject all the images into.
+    frame : `str`
+        The WCS frame of reference to use when reprojecting.
+        Can either be 'original' or 'ebd' to specify whether to
+        use the WorkUnit._per_image_wcs or ._per_image_ebd_wcs
+        respectively.
+    parallelize : bool
```
jeremykubica, line 67:
```suggestion
    parallelize : `bool`
```

src/kbmod/reprojection.py
```diff
@@ -41,7 +47,42 @@ def reproject_raw_image(image, original_wcs, common_wcs, obs_time):
...
+        If True, use multiprocessing to reproject the images in parallel.
+        Default is True.
+    max_parallel_processes : int
```
jeremykubica, line 70:
```suggestion
    max_parallel_processes : `int`
```

src/kbmod/reprojection.py
```diff
@@ -156,3 +197,231 @@ def reproject_work_unit(work_unit, common_wcs, frame="original"):
...
+    work_unit : `kbmod.WorkUnit`
+        The WorkUnit to be reprojected.
+    common_wcs : `astropy.wcs.WCS`
+        The WCS to reproject all the images into.
+    frame : `str`
+        The WCS frame of reference to use when reprojecting.
+        Can either be 'original' or 'ebd' to specify whether to
+        use the WorkUnit._per_image_wcs or ._per_image_ebd_wcs
+        respectively.
+    max_parallel_processes : int
```
jeremykubica, line 220:
```suggestion
    max_parallel_processes : `int`
```

jeremykubica (commented):
Only very small stylistic changes.

src/kbmod/reprojection.py
drewoldag: Done, I can add this to the returned WorkUnit in `_reproject_work_unit` as well.

src/kbmod/reprojection.py
drewoldag: Done

src/kbmod/reprojection.py
drewoldag: Done

src/kbmod/reprojection.py
drewoldag: Done

src/kbmod/reprojection.py
```diff
@@ -156,3 +197,231 @@ def reproject_work_unit(work_unit, common_wcs, frame="original"):
...
+    # if the time isn't in the list of times, raise an error.
+    if time not in obstimes:
+        raise ValueError(f"Observation time {time} not found in work unit.")
+
+    images = work_unit.im_stack.get_images()
+    index = np.where(obstimes == time)[0][0]
+    return images[index].get_psf()
+
+
+def _reproject_images(science_images, variance_images, mask_images, obstimes, common_wcs, original_wcs):
```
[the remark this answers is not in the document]
drewoldag: Added docstring here.

maxwest-uw (approved):
lgtm!!!

PR #629, opened 2024-05-28
Support multiple algorithms for correct_parallax.

drewoldag:
This PR creates a new dispatcher function with the name `correct_parallax` that will utilize either the baseline (renamed to correct_parallax_with_minimizer) or the exact geometric algorithm. 

The baseline has been modified as well to allow the user to specify a particular scipy minimizer algorithm, as well as a bound on the total search space. 

I've updated tests and added a demo notebook is included as well that shows how the two algorithms compare.


ColinOrionChandler:
I did some initial testing and I'm seeing some unexpected behavior. 
image

ColinOrionChandler:
After more testing, the method seems to be the problem, not the bounds. If I just comment out the method line and leave the bounds in, I'm able to successfully reflex-correct orbits that failed previously, like (43500) Chandler, 282P, and (6478) Gault.

drewoldag:
Ok, good to know. Thanks for taking a look at this. Jeremy has shared an analytic solution for this with me as well. I'll try to get that in the PR tomorrow and perhaps we can spend some time running the code together. Or if you want to share the script that you used to produce the image above, that would be great too.

ColinOrionChandler (approved):
Looks good to me, like what we discussed. Thanks!

PR #631, opened 2024-05-30
Update the kernel memory functions to allow large images

jeremykubica:
Change the kernel memory functions to allow arrays with a large amount of data (needing indices of int64). Change `unsigned long` to `uint64_t` for consistency throughout the code.

Also adds a divide by zero check in the core search code.


wilsonbb (approved):
Thanks for the quick turnaround!

PR #633, opened 2024-05-31
Add ability to output GPU memory stats on crash

jeremykubica:
If the program fails during a GPU memory operation (allocation, freeing, etc.), it first dumps basic memory stats to stdout to help with debugging.


wilsonbb (approved):
Awesome, thanks!

PR #636, opened 2024-05-31
Fix compiler warning

jeremykubica:
Revert to normal ints so the compiler doesn't give a warning about downscoping the variables.


drewoldag (approved):
This seems reasonable. I'm assuming that data.get_num_times() won't be a value that overflows, right?

jeremykubica:
> This seems reasonable. I'm assuming that data.get_num_times() won't be a value that overflows, right?

Correct. We don't expect the number of images to come anywhere near an overflow.

PR #639, opened 2024-06-03
Remove a copy and add some checks in convolution

jeremykubica:
Remove a copy of the PFS (which requires removing some `const` qualifiers). Add some more checks on the dimensionality of the images for convolution.


drewoldag (approved):
This all looks pretty reasonable, my C++ skills are a bit rusty, but I don't see a lot of logic changing here, so I feel ok giving this the green check.

PR #642, opened 2024-06-07
Attempting to reduce the total, temporary memory usage of reprojection.

drewoldag:
There are several small changes here, but they all revolve around decreasing the datatype that is used for temporary arrays or eliminating the creation of temporary arrays. 

I'll leave commentary notes in the file for more details.

### DIFF
--- a/src/kbmod/reprojection.py
+++ b/src/kbmod/reprojection.py
@@ -40,11 +40,24 @@ def reproject_image(image, original_wcs, common_wcs):
     image_data = CCDData(image, unit="adu")
     image_data.wcs = original_wcs
 
-    new_image, footprint = reproject.reproject_adaptive(
-        image_data, common_wcs, shape_out=common_wcs.array_shape, bad_value_mode="ignore"
-    )
+    footprint = np.zeros(common_wcs.array_shape, dtype=np.ubyte)
 
-    return new_image, footprint
+    # if the input image is actually a stack of images, we need to duplicate the
+    # footprint to match the total number of images.
+    if type(image) is list:
+        footprint = np.repeat(footprint[np.newaxis, :, :], len(image), axis=0)
+
+    new_image, _ = reproject.reproject_adaptive(
+        image_data, common_wcs, shape_out=common_wcs.array_shape, bad_value_mode="ignore",
+        output_footprint=footprint
+        )
+
+    # if we passed in a stack of ndarrays (i.e. science, varianace, mask), we only
+    # need to return the first footprint, as they should all be the same.
+    if footprint.ndim == 3:
+        footprint = footprint[0]
+
+    return new_image.astype(np.float32), footprint
 
 
 def reproject_work_unit(
@@ -114,10 +127,10 @@ def _reproject_work_unit(work_unit, common_wcs, frame="original"):
         indices = list(np.where(obstimes == time)[0])
         per_image_indices.append(indices)
 
-        science_add = np.zeros(common_wcs.array_shape)
-        variance_add = np.zeros(common_wcs.array_shape)
-        mask_add = np.zeros(common_wcs.array_shape)
-        footprint_add = np.zeros(common_wcs.array_shape)
+        science_add = np.zeros(common_wcs.array_shape, dtype=np.float32)
+        variance_add = np.zeros(common_wcs.array_shape, dtype=np.float32)
+        mask_add = np.zeros(common_wcs.array_shape, dtype=np.float32)
+        footprint_add = np.zeros(common_wcs.array_shape, dtype=np.ubyte)
 
         for index in indices:
             image = images[index]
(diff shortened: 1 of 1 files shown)


src/kbmod/reprojection.py
```diff
@@ -40,11 +40,24 @@ def reproject_image(image, original_wcs, common_wcs):
     image_data = CCDData(image, unit="adu")
     image_data.wcs = original_wcs
 
-    new_image, footprint = reproject.reproject_adaptive(
-        image_data, common_wcs, shape_out=common_wcs.array_shape, bad_value_mode="ignore"
-    )
+    footprint = np.zeros(common_wcs.array_shape, dtype=np.ubyte)
```
drewoldag, line 43: The `footprint` output array can be pre-allocated. We do so here with a specific dtype=np.ubyte to save size. Previously the would produce a footprint with dtype=float64.

src/kbmod/reprojection.py
```diff
@@ -40,11 +40,24 @@ def reproject_image(image, original_wcs, common_wcs):
...
+        output_footprint=footprint
+        )
+
+    # if we passed in a stack of ndarrays (i.e. science, varianace, mask), we only
+    # need to return the first footprint, as they should all be the same.
+    if footprint.ndim == 3:
+        footprint = footprint[0]
 
-    return new_image, footprint
+    return new_image.astype(np.float32), footprint
```
drewoldag, line 60: Returning `new_image` as float32. This will create a copy in memory from the float64 to the float32 version, but ultimately the goal is to reduce the amount of time that a float64 array is sitting in memory.

src/kbmod/reprojection.py
```diff
@@ -114,10 +127,10 @@ def _reproject_work_unit(work_unit, common_wcs, frame="original"):
         indices = list(np.where(obstimes == time)[0])
         per_image_indices.append(indices)
 
-        science_add = np.zeros(common_wcs.array_shape)
-        variance_add = np.zeros(common_wcs.array_shape)
-        mask_add = np.zeros(common_wcs.array_shape)
-        footprint_add = np.zeros(common_wcs.array_shape)
+        science_add = np.zeros(common_wcs.array_shape, dtype=np.float32)
+        variance_add = np.zeros(common_wcs.array_shape, dtype=np.float32)
+        mask_add = np.zeros(common_wcs.array_shape, dtype=np.float32)
```
drewoldag, line 132: These three were using the default float64 dtype, and then being cast to float32, which would require another copy in memory. Defining them as float32 to start with saves space and the memory copy.

src/kbmod/reprojection.py
```diff
@@ -114,10 +127,10 @@ def _reproject_work_unit(work_unit, common_wcs, frame="original"):
...
+        footprint_add = np.zeros(common_wcs.array_shape, dtype=np.ubyte)
```
drewoldag, line 133: Since this is just a mask, the dtype can be much smaller than the default float64

src/kbmod/reprojection.py
```diff
@@ -159,17 +172,18 @@ def _reproject_work_unit(work_unit, common_wcs, frame="original"):
...
 
-        # transforms the mask back into a bitmask.
-        mask_add = np.where(np.isclose(mask_add, 0.0, atol=0.2), 0.0, 1.0)
+        # transforms the mask back into a bitmask. Note that we need to be explicit
+        # about the dtypes for 0.0 and 1.0, otherwise mask_add will be cast to float64.
+        mask_add = np.where(np.isclose(mask_add, 0.0, atol=0.2), np.float32(0.0), np.float32(1.0))
 
-        science_raw_image = RawImage(img=science_add.astype("float32"), obs_time=time)
-        variance_raw_image = RawImage(img=variance_add.astype("float32"), obs_time=time)
-        mask_raw_image = RawImage(img=mask_add.astype("float32"), obs_time=time)
```
drewoldag, line 172: These casts are no longer required. Saves a copy in memory.

src/kbmod/reprojection.py
```diff
@@ -159,17 +172,18 @@ def _reproject_work_unit(work_unit, common_wcs, frame="original"):
...
+        gaps = footprint_add == 0
         science_add[gaps] = KB_NO_DATA
         variance_add[gaps] = KB_NO_DATA
         mask_add[gaps] = 1
 
-        # transforms the mask back into a bitmask.
-        mask_add = np.where(np.isclose(mask_add, 0.0, atol=0.2), 0.0, 1.0)
+        # transforms the mask back into a bitmask. Note that we need to be explicit
+        # about the dtypes for 0.0 and 1.0, otherwise mask_add will be cast to float64.
+        mask_add = np.where(np.isclose(mask_add, 0.0, atol=0.2), np.float32(0.0), np.float32(1.0))
```
drewoldag, line 182: Without explicitly casting np.float32(0.0) and np.float32(1.0) numpy will recast mask_add to float64. (Another memory copy and extra space required)

maxwest-uw (approved):
lgtm after black formatting 😎 nice work

PR #643, opened 2024-06-07
creates image_positions_to_original_icrs

### LINKED ISSUE #575: add wcs reprojection to `WorkUnit`
We have all the pieces to run reprojection and reflex correction as part of our broader kbmod search pipeline, all that's left to do is add the infrastructure to take a `WorkUnit` and correct the wcs's for parallax in **Explicit Barycentric Distance** space (EBD).

## Requirements

- [ ] Take a `WorkUnit` and reproject all of the `per_image_wcs` elements into EBD space.
- [ ] Ensure that we maintain a copy of the original WCSs somewhere in the `WorkUnit` structure.
- [ ] Ensure that we can save the EBD space WCSs when we output with `to_fits` and read those WCSs back in again.
- [ ] Have the option to use the EBD corrected WCSs when we run `reproject_workunit` on a `WorkUnit` (if available).

maxwest-uw:
closes #575 

creates a function in `WorkUnit` that allows a user to transform coordinates in EBD space back into the "original" ICRS frame. Supports pixel positions and radec positions for input and output.

### DIFF
--- a/src/kbmod/reprojection_utils.py
+++ b/src/kbmod/reprojection_utils.py
@@ -106,7 +106,7 @@ def invert_correct_parallax(coord, obstime, point_on_earth, geocentric_distance,
     )
 
     original_icrs = gcrs_with_dist.transform_to(ICRS())
-    return original_icrs
+    return SkyCoord(ra=original_icrs.ra, dec=original_icrs.dec, unit="deg")
 
 
 def fit_barycentric_wcs(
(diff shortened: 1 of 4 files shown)


drewoldag (approved):
Not sure how urgent it is to get this into main, but I took a look over it. I haven't had a chance to go deeply into the test cases. The `image_positions_to_original_icrs` seems good. I would maybe argue that there's a lot of logic in one function, that could be broken apart, but that's personal preference.

PR #650, opened 2024-06-12
Remove copies for LayeredImage creation

jeremykubica:
When given a set of `RawImage`s the `LayeredImage` constructor does a copy. This PR adds a constructor that allows `LayeredImage` to take in the underlying Eigen data array (or numpy array from Python), claim ownership of the memory, and avoid the copy.

We also use `ImageStack`'s new `append_image()` function to add images to the stack one on one instead of allocating a full list and copying them in. This slightly reduces the peak memory needed. We include a `force_move` option that allows the `ImageStack` to take ownership of the image and avoid doing a copy when it inserts it into the vector.

### DIFF
--- a/src/kbmod/reprojection.py
+++ b/src/kbmod/reprojection.py
@@ -125,7 +125,7 @@ def _reproject_work_unit(work_unit, common_wcs, frame="original"):
     unique_obstimes = np.unique(obstimes)
     per_image_indices = []
 
-    stack = ImageStack([])
+    stack = ImageStack()
     for time in unique_obstimes:
         indices = list(np.where(obstimes == time)[0])
         per_image_indices.append(indices)
@@ -193,7 +193,7 @@ def _reproject_work_unit(work_unit, common_wcs, frame="original"):
             psf,
             time,
         )
-        stack.append_image(new_layered_image)
+        stack.append_image(new_layered_image, force_move=True)
 
     per_image_wcs = work_unit._per_image_wcs
     per_image_ebd_wcs = work_unit._per_image_ebd_wcs
--- a/src/kbmod/reprojection.py
+++ b/src/kbmod/reprojection.py
@@ -280,7 +280,6 @@ def _reproject_work_unit_in_parallel(
 
     # when all the multiprocessing has finished, convert the returned numpy arrays to RawImages.
     concurrent.futures.wait(future_reprojections, return_when=concurrent.futures.ALL_COMPLETED)
-    time_stamps = []
     stack = ImageStack([])
     for result in future_reprojections:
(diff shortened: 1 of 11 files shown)


src/kbmod/reprojection.py
```diff
@@ -287,29 +281,26 @@ def _reproject_work_unit_in_parallel(
     # when all the multiprocessing has finished, convert the returned numpy arrays to RawImages.
     concurrent.futures.wait(future_reprojections, return_when=concurrent.futures.ALL_COMPLETED)
     time_stamps = []
```
drewoldag, line 283: I think that we can remove this variable entirely, if we're able to use the `ImageStack.sort_by_time()` method.

src/kbmod/reprojection.py
```diff
@@ -287,29 +281,26 @@ def _reproject_work_unit_in_parallel(
...
-            mask_raw_image,
+            science_add,
+            variance_add,
+            mask_add,
             psf,
+            time,
         )
 
         # append timestamps and layeredImages to lists
         time_stamps.append(time)
```
drewoldag, line 299: And can remove this line for the same reason as above.

src/kbmod/search/layered_image.cpp
```diff
@@ -306,6 +316,9 @@ static void layered_image_bindings(py::module& m) {
 
     py::class_<li>(m, "LayeredImage", pydocs::DOC_LayeredImage)
             .def(py::init<const ri&, const ri&, const ri&, pf&>())
+            .def(py::init<search::Image&, search::Image&, search::Image&, pf&, double>(),
+                 py::arg("sci").noconvert(true), py::arg("var").noconvert(true),
+                 py::arg("msk").noconvert(true), py::arg("psf"), py::arg("obs_time"))
```
drewoldag, line 321: What do these `noconvert`s do? Just curious, it's unfamiliar.

drewoldag (approved):
This looks great, thanks! No major concerns, mostly just happy to see things getting cleaned up.

src/kbmod/reprojection.py
jeremykubica: Cleaned it up

src/kbmod/reprojection.py
jeremykubica: cleaned this up too.

src/kbmod/search/layered_image.cpp
jeremykubica: I don't know all the details. But they are needed for passing numpy arrays of float32s into Eigen matrices in C++. I'm not sure if it prevents it from converting to a float64 or into a list of lists or something else.

PR #653, opened 2024-06-13
Update core search to use shared memory for times

jeremykubica:
This reintroduces an optimization that Peter had in the original KBMOD. Bring the image timestamps into faster per-block shared memory. This reduces waiting on memory and provides a few percent (~2%) speedup to the on-GPU search.

### DIFF
--- a/src/kbmod/search/kernels/kernels.cu
+++ b/src/kbmod/search/kernels/kernels.cu
@@ -224,7 +224,7 @@ __global__ void searchFilterImages(PsiPhiArrayMeta psi_phi_meta, void *psi_phi_v
     // copy their time before progressing. We need to do this before pruning on
     // (x, y) in order to correctly handle blocks at the edge of the image.
     __shared__ double shared_times[MAX_NUM_IMAGES];
-    int time_idx = threadIdx.x + threadIdx.y * THREAD_DIM_X;
+    int time_idx = threadIdx.x + threadIdx.y * blockDim.x;
     if (time_idx < psi_phi_meta.num_times) {
         shared_times[time_idx] = image_times[time_idx];
     }
--- a/src/kbmod/search/kernels/kernels.cu
+++ b/src/kbmod/search/kernels/kernels.cu
@@ -220,6 +220,16 @@ __global__ void searchFilterImages(PsiPhiArrayMeta psi_phi_meta, void *psi_phi_v
     assert(psi_phi_vect != nullptr && image_times != nullptr && trajectories != nullptr &&
            results != nullptr);
 
+    // Copy the times to faster shared memory for the block. Make sure all threads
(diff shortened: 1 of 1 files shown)


src/kbmod/search/kernels/kernels.cu
```diff
@@ -220,6 +220,16 @@ __global__ void searchFilterImages(PsiPhiArrayMeta psi_phi_meta, void *psi_phi_v
...
 
+    // Copy the times to faster shared memory for the block. Make sure all threads
+    // copy their time before progressing. We need to do this before pruning on
+    // (x, y) in order to correctly handle blocks at the edge of the image.
+    __shared__ double shared_times[MAX_NUM_IMAGES];
+    int time_idx = threadIdx.x + threadIdx.y * THREAD_DIM_X;
+    if (time_idx < psi_phi_meta.num_times) {
+        shared_times[time_idx] = image_times[time_idx];
+    }
+    __syncthreads();  // Block until all are done loading.
```
drewoldag, line 231: Is my understanding correct, that we call `__syncthreads` here because we have to wait for `shared_times` to be completely populated by all threads?

src/kbmod/search/kernels/kernels.cu
```diff
@@ -220,6 +220,16 @@ __global__ void searchFilterImages(PsiPhiArrayMeta psi_phi_meta, void *psi_phi_v
     assert(psi_phi_vect != nullptr && image_times != nullptr && trajectories != nullptr &&
            results != nullptr);
 
+    // Copy the times to faster shared memory for the block. Make sure all threads
+    // copy their time before progressing. We need to do this before pruning on
+    // (x, y) in order to correctly handle blocks at the edge of the image.
+    __shared__ double shared_times[MAX_NUM_IMAGES];
+    int time_idx = threadIdx.x + threadIdx.y * THREAD_DIM_X;
```
drewoldag, line 227: I believe you that this is the correct calculation of `time_idx`, but I not familiar enough to say with certainty. Is it really `threadIdx.x + threadIdx.y * THREAD_DIM_X`?

drewoldag (approved):
This looks good. I left a comment on the `time_idx` calculation, just because I don't know enough to say if it's right or wrong. Mostly just a prompt to double check :)

src/kbmod/search/kernels/kernels.cu
jeremykubica: Yes. That is correct. Each thread in the block loads a single time stamp.

src/kbmod/search/kernels/kernels.cu
jeremykubica: For the current code, I think that will work.  But I changed it to:

int time_idx = threadIdx.x + threadIdx.y * blockDim.x;

which should be safer in case the code that calls the kernel function changes.

PR #660, opened 2024-06-17
Change numpy NAN

jeremykubica:
Numpy's `NAN` is deprecated in v2.0. This changes the instances to `nan`. It does not fix everything broken with numpy v2.0, but crosses one known item off the list.


drewoldag (approved):
Looks good to me.

PR #661, opened 2024-06-20
Add some additional filtering operations

jeremykubica:
Add functions to filter results that might be helpful in analysis:
- Filter individual observations by min and max likelihood
- Filter result rows by the number of days on which they occur (based on @wilsonbb notebook)
- Filter result rows that cover less than a given duration in time

Also creates a related helper function in `Results` and fixes `update_obs_valid()` to remove rows where nothing is valid.


src/kbmod/filters/basic_filters.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    logger.info(f"Threshold clipping {len(result_data)} results with " f"bounds=[{lower_bnd}, {upper_bnd}]")
+
+    lh = result_data.compute_likelihood_curves(filter_obs=True, mask_value=np.nan)
+    obs_valid = np.isfinite(lh) & (lh <= upper_bnd) & (lh >= lower_bnd)
+    result_data.update_obs_valid(obs_valid)
+
+
+def apply_time_range_filter(result_data, mjds, min_days=np.inf, colname=None):
+    """Filter any row that does not valid observations that cover at least
```
wilsonbb, line 35: NIt: s/not valid/not have valid

src/kbmod/filters/basic_filters.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # Compute the duration of each result.
+    delta = np.nanmax(time_mat, axis=1) - np.nanmin(time_mat, axis=1)
+    if colname is not None:
+        result_data.table[colname] = delta
+    result_data.filter_rows(delta >= min_days, f"duration>={min_days}")
+
+
+def apply_unique_day_filter(result_data, mjds, min_days, min_per_day=1, colname=None):
+    """Filter any row that does not valid observations that occur on at least
```
wilsonbb, line 67: Same nit as above

s/not valid/not have valid

wilsonbb (approved):
LGTM, just a few typos!

PR #663, opened 2024-07-12
Standardizers/speedup butlerstd

DinoBektesevic:
Review after Friday Jul 12th, have to meet with Pedro to ensure we extract all the necessary metadata. 

The region_search tests pass `DatasetRef`s as `dataIds` - this is hot-patched in `MockedButler` but is something that we should probably fix before merging. 

These changes to `ButlerStandardizer` improve its performance. Standardizes `A0a`, ~400 images, in ~94 seconds or so. Metadata is extracted via objects that do not load image, but just header metadata. This could additionally be cleaned up to save even more time per dataset reference, by constructing bounding box out of the corners of the image in `summaryStats` instead of calculating them via WCS. 

These changes make the old `ButlerStandardizer` image collection unusable anymore, but given the costs to recreate those in this PR, I hope this is not a big problem. Minor changes to other functionality, more specific tests for raised errors and documentation.

Future change should probably be to just unroll the bounding box from objects into columns for ease of use.

### DIFF
--- a/src/kbmod/standardizers/butler_standardizer.py
+++ b/src/kbmod/standardizers/butler_standardizer.py
@@ -144,7 +144,9 @@ class ButlerStandardizer(Standardizer):
 
         return False
 
-    def __init__(self, id, butler, config=None, **kwargs):
+    def __init__(self, dataId, butler, config=None, **kwargs):
+        deferred_import("lsst.daf.butler", "dafButler")
+
         # Somewhere around w_2024_ builds the datastore.root
         # was removed as an attribute of the datastore, not sure
         # it was ever replaced with anything back-compatible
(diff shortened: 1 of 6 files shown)


src/kbmod/standardizers/butler_standardizer.py
```diff
@@ -141,53 +144,167 @@ def resolveTarget(self, tgt):
...
+        meta["NAXIS2"] = self._naxis2
+        self._wcs = WCS(meta)
+
+        bbox = self._computeBBox(self._wcs, self._naxis1, self._naxis2)
+        # this will unroll the entire bbox into columns
+        self._metadata.update(bbox)
+
+        # this is so the current imagecollection doesn't break, the plan is
+        # to update all BBOXes to unravel as columns instead of object-dict
+        # but thats meant for a different PR, TODO: change bbox
```
wilsonbb, line 284: Can we put this in a github issue for tracking

tests/utils/mock_butler.py
```diff
@@ -181,11 +258,16 @@ def getDataset(self, datid):
     def get(self, ref, collections=None, dataId=None):
         orig_ref = ref
 
+        # this covers tests in region_search because they pass dataId (that are
+        # actuall dataRefs) with a component name as ref. TODO: fix this?
+        if dataId is not None and isinstance(ref, str):
+            orig_ref = dataId.makeComponentRef(ref.split(".")[-1])
```
wilsonbb, line 264: Yeah I don't remember how this happened based on my early butler testing, but I should fix this. Thanks!

wilsonbb (approved):
Thanks Dino, this looks good to me!

Approving with a few small comments

src/kbmod/standardizers/butler_standardizer.py
```diff
@@ -141,53 +144,167 @@ def resolveTarget(self, tgt):
...
+        self._metadata["psfArea"] = summary.psfArea
+        self._metadata["nPsfStar"] = summary.nPsfStar
+        self._metadata["zeroPoint"] = summary.zeroPoint
+        self._metadata["skyBg"] = summary.skyBg
+        self._metadata["skyNoise"] = summary.skyNoise
+        self._metadata["meanVar"] = summary.meanVar
+        self._metadata["astromOffsetMean"] = summary.astromOffsetMean
+        self._metadata["astromOffsetStd"] = summary.astromOffsetStd
+
+        # Will be nan because it's VR filter and task doesn't include it
```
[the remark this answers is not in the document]
DinoBektesevic: Not a silly question. It will be true for DEEP because DEEP was taken in the VR filter (no filter really) and the task that creates summary statistics only works with ugrizy filters. So the effective metrics don't exist because the task never found any data it could run on. So if it was DECam data in DEEP that was in a filter, that would likely have those values is my understanding.

PR #664, opened 2024-07-15
`WorkUnit` fits sharding and lazy loading

maxwest-uw:
The cleaned up version of my prototype for sharded WorkUnit file storage.

- these are the changes to just the `WorkUnit`, for reprojection we'll still need to rewrite the code to load in the data lazily per worker. However, a sharded `WorkUnit` that's eager loaded will work exactly the same as any other `WorkUnit` in memory.

### DIFF
--- a/src/kbmod/work_unit.py
+++ b/src/kbmod/work_unit.py
@@ -458,8 +458,6 @@ class WorkUnit:
         Uses the following extensions:
             0 - Primary header with overall metadata
             1 or "metadata" - The data provenance metadata
-
-
             2 or "kbmod_config" - The search parameters
             3+ - Image extensions for the science layer ("SCI_i"),
                 variance layer ("VAR_i"), mask layer ("MSK_i"), and
@@ -548,10 +546,10 @@ class WorkUnit:
 
         hdul.writeto(filename, overwrite=overwrite)
 
-    def to_fits_shard(self, filename, directory, overwrite=False):
+    def to_sharded_fits(self, filename, directory, overwrite=False):
         """Write the WorkUnit to a multiple FITS files.
         Will create:
-            - One "primary"file, containing the main WorkUnit metadata
+            - One "primary" file, containing the main WorkUnit metadata
             (see below) as well as the per_image_wcs information for
             the whole set. This will have the given filename.
             -One image fits file containing all of the image data for
@@ -581,7 +579,7 @@ class WorkUnit:
             Indicates whether to overwrite an existing file.
         """
         logger.info(
-            f"Writing WorkUnit shards with {self.im_stack.img_count()} images with main file {filename} in {directory }"
+            f"Writing WorkUnit shards with {self.im_stack.img_count()} images with main file {filename} in {directory}"
         )
         primary_file = os.path.join(directory, filename)
         if Path(primary_file).is_file() and not overwrite:
@@ -662,9 +660,9 @@ class WorkUnit:
         hdul.writeto(os.path.join(directory, filename), overwrite=overwrite)
 
     @classmethod
-    def from_fits_shard(cls, filename, directory, lazy=False):
+    def from_sharded_fits(cls, filename, directory, lazy=False):
         """Create a WorkUnit from multiple FITS files.
-        Pointed towards the result of WorkUnit.to_fits_shard.
+        Pointed towards the result of WorkUnit.to_sharded_fits.
 
         The FITS files will have the following extensions:
 
@@ -701,7 +699,6 @@ class WorkUnit:
 
         # open the main header
         with fits.open(os.path.join(directory, filename)) as primary:
-            print(primary[2].name)
             config = SearchConfiguration.from_hdu(primary["kbmod_config"])
 
             # Read in the global WCS from extension 0 if the information exists.
@@ -762,7 +759,7 @@ class WorkUnit:
                     sub_indices.append(sci_hdu.header[f"IND_{j}"])
                 per_image_indices.append(sub_indices)
 
-        file_paths = None if lazy else file_paths
+        file_paths = None if not lazy else file_paths
         result = WorkUnit(
             im_stack=im_stack,
             config=config,
--- a/src/kbmod/work_unit.py
+++ b/src/kbmod/work_unit.py
@@ -541,6 +541,205 @@ class WorkUnit:
 
         hdul.writeto(filename, overwrite=overwrite)
 
+    def to_fits_shard(self, filename, directory, overwrite=False):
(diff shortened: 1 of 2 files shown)


src/kbmod/work_unit.py
```diff
@@ -75,9 +76,13 @@ def __init__(
         geocentric_distances=None,
         reprojected=False,
         per_image_indices=None,
+        lazy=False,
+        file_paths=None,
```
jeremykubica, line 80: How does using file_paths overlap with what `ImageCollection` is doing when getting things? Is there anything there we could reuse?

src/kbmod/work_unit.py
```diff
@@ -541,6 +548,237 @@ def to_fits(self, filename, overwrite=False):
...
+            The base filename to which to write the data.
+        directory: `str`
+            The directory to place all of the FITS files.
+            Recommended that you have one directory per
+            sharded file to avoid confusion.
+        overwrite : bool
+            Indicates whether to overwrite an existing file.
+        """
+        logger.info(
+            f"Writing WorkUnit shards with {self.im_stack.img_count()} images with main file {filename} in {directory }"
```
jeremykubica, line 584: You can remove the space between the directory variable and the closing }.

src/kbmod/work_unit.py
```diff
@@ -541,6 +548,237 @@ def to_fits(self, filename, overwrite=False):
...
+                else:
+                    file_paths.append(shard_path)
+
+                n_indices = sci_hdu.header["NIND"]
+                sub_indices = []
+                for j in range(n_indices):
+                    sub_indices.append(sci_hdu.header[f"IND_{j}"])
+                per_image_indices.append(sub_indices)
+
+        file_paths = None if lazy else file_paths
```
jeremykubica, line 765: This looks like it is backwards to me. I think we would want to make the `file_paths` `None` if the setting is not lazy, because we need `file_paths` for the later load.

src/kbmod/work_unit.py
```diff
@@ -541,6 +548,237 @@ def to_fits(self, filename, overwrite=False):
 
         hdul.writeto(filename, overwrite=overwrite)
 
+    def to_fits_shard(self, filename, directory, overwrite=False):
```
jeremykubica, line 551: As a minor naming note, I would suggest `to_sharded_fits`. But I'm fine if you want to keep it as-is.

src/kbmod/work_unit.py
```diff
@@ -541,6 +548,237 @@ def to_fits(self, filename, overwrite=False):
...
+            im_ebd_wcs = self._per_image_ebd_wcs[i]
+            ebd_hdu = fits.TableHDU()
+            append_wcs_to_hdu_header(im_ebd_wcs, ebd_hdu.header)
+            ebd_hdu.name = f"EBD_{i}"
+            hdul.append(ebd_hdu)
+
+        hdul.writeto(os.path.join(directory, filename), overwrite=overwrite)
+
+    @classmethod
+    def from_fits_shard(cls, filename, directory, lazy=False):
```
jeremykubica, line 665: As a minor naming note, I would suggest `from_sharded_fits`. But I'm fine if you want to keep it as-is.

tests/test_work_unit.py
```diff
@@ -284,6 +284,72 @@ def test_save_and_load_fits(self):
             # We succeed if overwrite=True
             work.to_fits(file_path, overwrite=True)
 
+    def test_save_and_load_fits_shard(self):
```
jeremykubica, line 287: Can you also add a test for the lazy load?

tests/test_work_unit.py
```diff
@@ -284,6 +284,72 @@ def test_save_and_load_fits(self):
...
+                sci1 = li.get_science()
+                var1 = li.get_variance()
+                msk1 = li.get_mask()
+
+                li_org = self.im_stack.get_single_image(i)
+                sci2 = li_org.get_science()
+                var2 = li_org.get_variance()
+                msk2 = li_org.get_mask()
+
+                for y in range(self.height):
```
jeremykubica, line 322: You can use the `RawImage.l2_allclose()` function here. It will be a lot faster and handle NaNs.

```
self.assertTrue(sci1.l2_allclose(sci2, 1e-3))
```

tests/test_work_unit.py
```diff
@@ -284,6 +284,72 @@ def test_save_and_load_fits(self):
...
+                        self.assertAlmostEqual(sci1.get_pixel(y, x), sci2.get_pixel(y, x))
+                        self.assertAlmostEqual(var1.get_pixel(y, x), var2.get_pixel(y, x))
+                        self.assertAlmostEqual(msk1.get_pixel(y, x), msk2.get_pixel(y, x))
+
+                # Check the PSF layer matches.
+                p1 = self.p[i]
+                p2 = li.get_psf()
+                self.assertEqual(p1.get_dim(), p2.get_dim())
+
+                for y in range(p1.get_dim()):
```
jeremykubica, line 333: You could use the `PSF.is_close()` function here. It is much faster.

src/kbmod/work_unit.py
```diff
@@ -541,6 +548,237 @@ def to_fits(self, filename, overwrite=False):
...
+        for i in range(self.im_stack.img_count()):
+            layered = self.im_stack.get_single_image(i)
+            c_indices = self._per_image_indices[i]
+            n_indices = len(c_indices)
+            sub_hdul = fits.HDUList()
+
+            img_wcs = self.get_wcs(i)
+            sci_hdu = raw_image_to_hdu(layered.get_science(), img_wcs)
+            sci_hdu.name = f"SCI_{i}"
+            sci_hdu.header["NIND"] = n_indices
```
jeremykubica, line 626: Not for this PR, but a thought for the future: If we are going to use the `NIND` and `IND_{j}` header values a lot, it might be worth breaking these into a helper function. Something like "fits_header_read_image_indices()" and "fits_header_append_image_indices()"

jeremykubica (commented):
Generally this looks good. I made a few stylistic suggestions and have a few questions.

src/kbmod/work_unit.py
maxwest-uw: I think the `file_paths` for `WorkUnit` here specifically refer to files generated by a WorkUnit as part of sharding, and my assumption is that we would want to keep it that way as there is a one-to-many relationship between `ImageCollection` URIs and `WorkUnit`s, but there might be some wisdom in making `WorkUnit` loading even lazier (esp. when we're not doing reprojection) 😅

jeremykubica (approved):

Issue #548, opened 2024-04-08
Add sorting to ImageCollection


DinoBektesevic:
It'd be pretty cool if we could sort IC the same way we do a Table. That would enable groups to work to I think. Issue is ordering of the standardizers and the WCS. Right now they're kept as separate lists (for lazy loading) but really they can just be columns. 

Making them be columns lets us sort without worrying about ordering.

Issue #666, opened 2024-07-16
Change how WCS and BBOX work in ImageCollection


DinoBektesevic:
Instead of providing a mechanims to encapsulate per-row serializable objects unroll the WCS and BBOX into columns. 
I'm not sure WCS can be serialized into columns, since all the keys can not be known in-advance, but certainly BBOX could.

For WCS also calculate on-sky coordinates of all 4 corners of the detector and the central pixel value. 
The current mechanism can be changed from what it is to lazy loading of columns for BBOX and WCS (which can be left in a column as a json header string to be loaded at request time instead of in the `read`).

DinoBektesevic:
Fixed in #668 but wasn't automatically closed.

PR #668, opened 2024-07-17
Speedups and memory savings for ImageCollection and Standardizers

DinoBektesevic:
This PR is mostly intended for large ImageCollection optimizations really. 
The measured performance benefits are substantial in terms of loading times and memory footprint of large image collections:
```
before 
real	47m44.168s 

after
real	7m42.795s
```

per-standardizer row we save another 5K of data, or approximately 50% of memory we used to take. I wanted to test the impact on the standardization times, since those must not be significantly impacted, but someone is currently really thrashing the disks so I can't get reliable results. For 3 datasetrefs, should be on the order of a second, I got the following results:
```
old
Time to standardize A0a collection: 14.083693213993683
Time to standardize A0a collection: 12.05290256603621
Time to standardize A0a collection: 16.573551627923734

new
Time to standardize A0a collection: 16.573551627923734
Time to standardize A0a collection: 4.487202679971233
Time to standardize A0a collection: 13.363859862089157
```
so I'm going to have to repeat these tests to see what's the damage. I expect a slight increase in time because of the serialization of the WCS takes a bit more steps now, but I don't think it should be that bad.

Speedups include:
- not loading all BBOX and WCS as objects at IC init time
- adding the ability to disable lazy loading
- serializing BBOX as columns to save on data volume of IC
- serializing WCS as dict to save on size of IC
- serializing WCS and config with the smallest separators possible
- trimming padding and spaces from the serialized objects

Changes to BBox and WCS handling:
- WCS is now a mandatory part of the standardized data
- Location was removed as mandatory part of standardized data
- BBox is now unravelled as columns instead of dicts.
- BBox is now defined as on-sky coordinates of the 4 chip corners of the chip, clockwise, and center pixel
- WCS is not longer part of the standardized metadata WCS is now serialized by the IC.
- WCS is serialized as a dictionary
- WCS remains an attribute of the standardizers

Changes to IC:
- lazy loading can be turned of at instantiation time now. Lazy loading only happens if _standardized list exists.
- `standardizers`, `wcs` and `bbox` properties are now iterators
- `get_wcs` and `get_bbox` are now getters that do not support lazy loading
- `get_standardizer` continues to support lazy loading

### DIFF
--- a/src/kbmod/image_collection.py
+++ b/src/kbmod/image_collection.py
@@ -377,7 +377,7 @@ class ImageCollection:
             # the warnings that some keywords might be ignored are expected
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore")
-                yield WCS(json.loads(self.data["wcs"][i]), relax=True)
+                yield WCS(json.loads(self.data[i]["wcs"]), relax=True)
 
     def get_wcs(self, idxs):
         # select column before indices, because a copy of the data
@@ -411,12 +411,6 @@ class ImageCollection:
     def columns(self):
         """Return metadata columns."""
         return self.data[self._userColumns].columns
-
-    @property
-    def standardizers(self):
-        """Standardizer generator."""
-        for i in range(len(self.data)):
-            yield self.get_standardizer(i)
 
     def get_standardizer(self, index, **kwargs):
         """Get the standardizer and extension index for the selected row of the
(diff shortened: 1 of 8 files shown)


wilsonbb (approved):
This looks good to me. And the performance improvements here are awesome to see!

DinoBektesevic:
Resolves https://github.com/dirac-institute/kbmod/issues/548 and https://github.com/dirac-institute/kbmod/issues/666

PR #675, opened 2024-07-30
Lazy reprojection

### LINKED ISSUE #615: Reproject taking 300+ Gb of RAM for 200 (mosaics)
Too much RAM

maxwest-uw:
closes #615 

builds on the changes made in #664 

- adds in a `reproject_lazy_work_unit` function that can take in a lazy loaded `WorkUnit` and reproject it in parallel, only loading in the necessary data per task which significantly keeps the amount of memory used significantly lower from ~200 gb to ~2 gb.
- Also writes the reprojected `WorkUnit` into a sharded work unit.
- Slight refactor of the `WorkUnit` fits functionality, now have separate functions to write out the metadata.
- Minor change in the `_reproject_images` function to only pass a single `obstime` to the function.

### DIFF
--- a/src/kbmod/reprojection.py
+++ b/src/kbmod/reprojection.py
@@ -3,9 +3,13 @@ import concurrent.futures
 import reproject
 from astropy.nddata import CCDData
 from astropy.wcs import WCS
+from astropy.io import fits
+import os
+from copy import copy
 
 from kbmod.search import KB_NO_DATA, PSF, ImageStack, LayeredImage, RawImage
 from kbmod.work_unit import WorkUnit
+from kbmod.wcs_utils import append_wcs_to_hdu_header
 
 # The number of executors to use in the parallel reprojecting function.
 MAX_PROCESSES = 8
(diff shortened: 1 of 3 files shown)


src/kbmod/reprojection.py
```diff
@@ -264,7 +268,7 @@ def _reproject_work_unit_in_parallel(
             variance_images_at_obstime = [this_image.get_variance().image for this_image in images_at_obstime]
             mask_images_at_obstime = [this_image.get_mask().image for this_image in images_at_obstime]
 
-            obstimes = [all_obstimes[i] for i in indices]
+            obstime = all_obstimes[indices[0]]
```
drewoldag, line 271: Thanks for cleaning this up!

src/kbmod/reprojection.py
```diff
@@ -467,4 +620,31 @@ def _reproject_images(science_images, variance_images, mask_images, obstimes, co
...
-    return science_add, variance_add, mask_add, time
+    return science_add, variance_add, mask_add, obstime
+
+
+def image_add_to_hdu(add, name, obstime, wcs=None):
+    """Helper function that creates a HDU out of RawImage.
+
+    Parameters
+    ----------
+    img : `RawImage`
```
drewoldag, line 631: Minor point, some of the input params don't match the docstring.

src/kbmod/reprojection.py
```diff
@@ -391,7 +474,80 @@ def _get_first_psf_at_time(work_unit, time):
...
+    science_images = []
+    variance_images = []
+    mask_images = []
+    psfs = []
+
+    for file_path, index in zip(file_paths, indices):
+        with fits.open(file_path) as hdul:
+            science_images.append(hdul[f"SCI_{index}"].data.astype(np.single))
+            variance_images.append(hdul[f"VAR_{index}"].data.astype(np.single))
+            mask_images.append(hdul[f"MSK_{index}"].data.astype(np.single))
```
drewoldag, line 513: Does it make a difference that this is brought in as a `np.single` data type? Would `np.bool` work?

src/kbmod/work_unit.py
```diff
@@ -78,11 +78,13 @@ def __init__(
         per_image_indices=None,
         lazy=False,
         file_paths=None,
+        obstimes=None,
```
drewoldag, line 81: I would recommend updating the docstring too.  If you have good context for `lazy` and `file_paths` that would be nice too.

src/kbmod/work_unit.py
```diff
@@ -773,9 +710,83 @@ def from_sharded_fits(cls, filename, directory, lazy=False):
...
+        pri.header["HELIO"] = self.heliocentric_distance
+        for i in range(len(self.constituent_images)):
+            pri.header[f"GEO_{i}"] = self.geocentric_distances[i]
+
+        # If the global WCS exists, append the corresponding keys.
+        if self.wcs is not None:
+            append_wcs_to_hdu_header(self.wcs, pri.header)
+
+        n_constituents = len(self.constituent_images)
+        pri.header["NCON"] = n_constituents
```
drewoldag, line 747: Is there a reason this is two lines?

drewoldag: Oh, I think I see what's going on here. This was abstracted out of the `to_fits` and `to_sharded_fits` methods.

drewoldag (approved):
This all looks pretty good. I left a couple comments, but I don't see any major problems.

PR #676, opened 2024-08-01
unique obstime refactor

maxwest-uw:
change how we handle/get the unique obstimes and indices.


drewoldag (approved):
Cool, this looks like a great clean up!

PR #680, opened 2024-08-06
Caching doesn't cache

DinoBektesevic:
I used `row_index` instead of `std_idx` to index standardizer cache list making the cache miss everytime. 
Also packing of a single exposure packs the whole table which causes len(table) to be 0 instead of 1 in the unpack table function.

Thanks Drew!


drewoldag (approved):
This looks good, thank you!

PR #683, opened 2024-08-07
add sharded WorkUnit output for non-lazy input WorkUnits in reprojection

### LINKED ISSUE #678: add ability for non-lazy reprojection to write to a sharded fits
should support #622 

If we want to ensure that reprojection can be handled by one function call, we should add the functionality to have a parallel reprojection output the result as a sharded fits. Right now, this functionality is only available to an already sharded fits loaded with `lazy=True`, but for certain work flows we shouldn't have to write out an intermediary `WorkUnit` just to run reprojection. It is still worth it imo to write the final result to a sharded `WorkUnit` instead of pickling+unpickling it in serial after reprojection, as reading it from disk after reprojection is significantly faster.

maxwest-uw:
resolves #678 

allows us more flexibility in the reprojection workflow to choose between time and memory usage, i.e. instead of writing out an intermediary `WorkUnit` between the ic->wu generation and reprojection we can just pass in our already loaded `WorkUnit` but have the output be a sharded `WorkUnit`, using more memory but potentially saving time.

- new `write_output` flag in `reproject_work_unit` to create write the output of reprojection instead of re-materializing as a `WorkUnit` in memory
- also integrates the `reproject_lazy_work_unit` function into the main `reproject_work_unit` function.

### DIFF
--- a/src/kbmod/reprojection.py
+++ b/src/kbmod/reprojection.py
@@ -111,7 +111,8 @@ def reproject_work_unit(
 
     Returns
     ----------
-    A `kbmod.WorkUnit` reprojected with a common `astropy.wcs.WCS`.
+    A `kbmod.WorkUnit` reprojected with a common `astropy.wcs.WCS`, or `None` in the case
+    where `write_output` is set to True.
     """
     if (work_unit.lazy or write_output) and (directory is None or filename is None):
         raise ValueError("can't write output to sharded fits without directory and filename provided.")
@@ -165,7 +166,8 @@ def _reproject_work_unit(
         The base filename where output will be written if `write_output` is set to True.
     Returns
     ----------
-    A `kbmod.WorkUnit` reprojected with a common `astropy.wcs.WCS`.
+    A `kbmod.WorkUnit` reprojected with a common `astropy.wcs.WCS`, or `None` in the case
+    where `write_output` is set to True.
     """
     images = work_unit.im_stack.get_images()
     unique_obstimes, unique_obstime_indices = work_unit.get_unique_obstimes_and_indices()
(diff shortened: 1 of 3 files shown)


src/kbmod/reprojection.py
```diff
@@ -191,35 +233,64 @@ def _reproject_work_unit(work_unit, common_wcs, frame="original"):
...
-        wcs=common_wcs,
-        constituent_images=work_unit.constituent_images,
-        per_image_wcs=per_image_wcs,
-        per_image_ebd_wcs=per_image_ebd_wcs,
-        per_image_indices=unique_obstime_indices,
-        geocentric_distances=work_unit.geocentric_distances,
-        reprojected=True,
-    )
+        hdul = new_work_unit.metadata_to_primary_header()
+        hdul.writeto(os.path.join(directory, filename))
```
jeremykubica, line 266: It looks like you are no longer returning a workunit down this branch. Is that correct? If so, you should update the doc string to note that the function will return `None` if `write_output` is set to `True`.

src/kbmod/reprojection.py
```diff
@@ -191,35 +233,64 @@ def _reproject_work_unit(work_unit, common_wcs, frame="original"):
...
+    else:
+        per_image_wcs = work_unit._per_image_wcs
+        per_image_ebd_wcs = work_unit._per_image_ebd_wcs
```
jeremykubica, line 269: I don't think you need to set these here. I think you can just reference them directly in the constructor parameters below.

src/kbmod/reprojection.py
```diff
@@ -191,35 +233,64 @@ def _reproject_work_unit(work_unit, common_wcs, frame="original"):
...
+
+        new_wunit = WorkUnit(
+            im_stack=stack,
+            config=work_unit.config,
+            wcs=common_wcs,
+            constituent_images=work_unit.constituent_images,
+            per_image_wcs=per_image_wcs,
```
jeremykubica, line 276: I think this can just be:
```
per_image_wcs=work_unit._per_image_wcs
```

src/kbmod/reprojection.py
```diff
@@ -624,6 +790,56 @@ def _reproject_images(science_images, variance_images, mask_images, obstime, com
     return science_add, variance_add, mask_add, obstime
 
 
+def _write_images_to_shard(
+    science_add, variance_add, mask_add, psf, wcs, obstime, obstime_index, indices, directory, filename
+):
+    """
```
jeremykubica, line 796: Can you add a one sentence summary of the function? In this case I'm guessing is something along the lines of "Write the three image layers to a workunit shard"

src/kbmod/reprojection.py
```diff
@@ -289,38 +387,52 @@ def _reproject_work_unit_in_parallel(
...
+            constituent_images=work_unit.constituent_images,
+            per_image_wcs=work_unit._per_image_wcs,
+            per_image_ebd_wcs=work_unit._per_image_ebd_wcs,
+            per_image_indices=unique_obstimes_indices,
+            geocentric_distances=work_unit.geocentric_distances,
+            reprojected=True,
+        )
 
-    return new_wunit
+        return new_wunit
```
jeremykubica, line 435: It looks like you are no longer returning a workunit down this branch. Is that correct? If so, you should update the doc string to note that the function will return None if write_output is set to True.

jeremykubica (commented):
A few questions and stylistic comments

jeremykubica (approved):

PR #687, opened 2024-08-21
Make SkyWcs the default bbox WCS and calculate WCS errors.

DinoBektesevic:
Add wcs error column to the butler standardizer. 
Make SkyWcs the default way to calculate BBOXes so that we can confidently say that's correct. 
Rejigger the ordering of the points in the BBOX to match AstroPy convention.

### DIFF
--- a/src/kbmod/standardizers/butler_standardizer.py
+++ b/src/kbmod/standardizers/butler_standardizer.py
@@ -203,6 +203,58 @@ class ButlerStandardizer(Standardizer):
         self._naxis1 = None
         self._naxis2 = None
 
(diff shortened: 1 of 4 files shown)


wilsonbb (changes requested):
This looks fine to me, though there seem to be some unit test failures due to what at a glance looks like a mocking issue?

DinoBektesevic:
Yep, I do math with SkyWcs now so that wasn't mocked. I'm just going to wait for Colin to confirm the new collections actually work before merging cause it's possible I overlooked something. Thanks for the review.

wilsonbb (approved):

PR #689, opened 2024-08-23
add tqdm disabling option

### LINKED ISSUE #688: progress bars should be optional
should be a way to disable the `tqdm` progress bars

maxwest-uw:
resolves #688 

also adds the progress bar to the lazy work unit reprojection changes (which weren't merged at the time of the original pr 😅 )

### DIFF
--- a/src/kbmod/__init__.py
+++ b/src/kbmod/__init__.py
@@ -13,6 +13,8 @@ from logging import config as _config
 # Import the rest of the package
 from kbmod.search import Logging
 
+PROGRESS_BAR = bool(int(os.environ.get("KB_PROGRESS_BARS", 1)))
+
 # there are ways for this to go to a file, but is it worth it?
 # Then we have to roll a whole logging.config_from_shared_config thing
 _SHARED_LOGGING_CONFIG = {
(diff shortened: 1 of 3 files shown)


DinoBektesevic (changes requested):
Thanks for the quick turnaround. This all looks fine to me, but I do have one request. 

Can you make a global var in top-level init something like:
```
PROGRESS_BAR=True
```

and then make 

```
disable_progress=kbmod.PROGRESS_BAR
```

so that I can 

```
import kbmod
kbmod.PROGRESS_BARS = False
```

and turn them all off at once. 

Bonus points if you set the default value as:

```
PROGRESS_BARS = bool(os.environ.get("KB_PROGRESS_BARS", True)
```

so that I can also set the `KB_PROGRESS_BARS` env to a falsy value in the running scripts instead of having to repeat it in every task after import.

maxwest-uw:
part one of your request (being able to set `PROGRESS_BAR` at runtime) doesn't seem to work, as the downstream import in `WorkUnit` only ever seems to load what it was initially

DinoBektesevic:
Importing it will make a copy of the variable and then keep it in the module's global namespace. You have to get your values from where they're defined. I rejiggered the global variable access and removed some unused imports. This seems to work fine for me now, perhaps double check.

maxwest-uw:
looks good to me (can I say that on my on pr? 😭  )

DinoBektesevic (approved):

PR #690, opened 2024-08-29
Obsolete the old masking

DinoBektesevic:
I think I did this right, but I did have to trim out some masking config setting in the tests to get rid of the warning messages.


jeremykubica (approved):
Thank you for putting this together. I'm really excited to see this cleanup go in.

PR #691, opened 2024-08-29
add Visualizer and visualizer_utils

maxwest-uw:
current usage:
```python
import kbmod as kb
from kbmod.analysis.visualizer import Visualizer

wu_path = "/path/to/wu.fits" 
res_path = "/path/to/results/table.ecsv"

wu = kb.work_unit.WorkUnit.from_fits(wu_path)
stack = wu.im_stack
results = kb.results.Results.read_table(res_path))

vis = Visualizer(stack, results)

# maybe this should be done as part of the __init__,
# but these methods/columns might not be needed
# with other visualizations, curious what you think
vis.generate_all_stamps()
vis.count_num_days()

# plots result 0
vis.plot_daily_coadds(0)

# saves plot to png
vis.plot_daily_coadds(0, filename="result_0.png")

# plots all coadds with more than one days worth of day
idxs = np.where(results.table["num_days"] > 1)
for idx in idxs:
    vis.plot_daily_coadds(idx)
```

### DIFF
--- a/src/kbmod/analysis/visualization_utils.py
+++ b/src/kbmod/analysis/visualization_utils.py
@@ -15,4 +15,4 @@ def mjd_to_day(mjd):
     A `str` with a calendar date, in the format YYYY-MM-DD.
     e.g., mjd=60000 -> '2023-02-25'
     """
-    return str(Time(mjd, format="mjd").to_value("datetime")).split()[0]
+    return Time(mjd, format="mjd").strftime("%Y-%m-%d")
--- a/src/kbmod/analysis/visualization_utils.py
+++ b/src/kbmod/analysis/visualization_utils.py
@@ -1,0 +1,18 @@
+from astropy.time import Time
+
+
+def mjd_to_day(mjd):
+    """Takes an mjd and converts it into a day in
+    calendar date format.
+
+    Parameters
+    ----------
+    mjd : `float`
+        mjd format date.
+
+    Returns
+    ----------
+    A `str` with a calendar date, in the format YYYY-MM-DD.
+    e.g., mjd=60000 -> '2023-02-25'
+    """
+    return str(Time(mjd, format="mjd").to_value("datetime")).split()[0]
--- a/src/kbmod/analysis/visualizer.py
+++ b/src/kbmod/analysis/visualizer.py
@@ -38,7 +38,8 @@ class Visualizer:
 
     def count_num_days(self):
         """Counts the number of days that a given result
-        has valid data in.
+        has valid data in and adds a new column called
+        `num_days` to the results table.
         """
         num_days = []
         for idx in range(len(self.results)):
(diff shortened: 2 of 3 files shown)


src/kbmod/analysis/visualization_utils.py
```diff
@@ -0,0 +1,18 @@
...
+    ----------
+    mjd : `float`
+        mjd format date.
+
+    Returns
+    ----------
+    A `str` with a calendar date, in the format YYYY-MM-DD.
+    e.g., mjd=60000 -> '2023-02-25'
+    """
+    return str(Time(mjd, format="mjd").to_value("datetime")).split()[0]
```
DinoBektesevic, line 18: It should be possible to print a timestamp in whatever format you want, including custom C-style formatting, without making an entire module for it. 

https://docs.astropy.org/en/stable/api/astropy.time.Time.html

```
>>> from astropy.time import Time
>>> t = Time(56000, format="mjd")
>>> t.fits
'2012-03-14T00:00:00.000'
>>> t.yday
'2012:074:00:00:00.000'
>>> t.ymdhms
(2012, 3, 14, 0, 0, 0.)
>>> t.iso
'2012-03-14 00:00:00.000'
>>> t.jyear
2012.1984941820672
>>> t.iso
'2012-03-14 00:00:00.000'
>>> t.strftime("%Y-%M-%d")
'2012-00-14'
>>> t.to_datetime()
datetime.datetime(2012, 3, 14, 0, 0)
```

maxwest-uw: I'm planning on expanding the `visualization_utils` module with other functions, but for the first PR I just wanted to have everything set up to replicate the TNO notebook functionality

src/kbmod/analysis/visualizer.py
```diff
@@ -0,0 +1,113 @@
...
+        radius : `int`
+            radius of the stamp.
+        """
+        self.results.table["all_stamps"] = [
+            StampCreator.get_stamps(self.im_stack, trj, radius) for trj in self.trajectories
+        ]
+
+    def count_num_days(self):
+        """Counts the number of days that a given result
+        has valid data in.
```
jeremykubica, line 41: Can you add a sentence that it saves this internally to a new column in the results?

src/kbmod/analysis/visualization_utils.py
jeremykubica: I think it is okay to leave in for now given how it is used in the other file (to generate a key for a hash table). But if we don't end up adding more to `visualization_utils` we should clean it up.

jeremykubica (approved):

src/kbmod/analysis/visualization_utils.py
DinoBektesevic: Can we at least avoid casting to two different objects and making a copy of one with:

> t.strftime(fmt)

jeremykubica: Sounds good. Let's do that.

PR #693, opened 2024-09-04
Cleanup tests that rely on legacy file formats

jeremykubica:
Remove a bunch of tests that use legacy file formats (partial step toward all file reading happening through the standardizers):

- Remove the file reading and writing tests for `LayeredImage`
- Make the regression test create the `WorkUnit` in memory instead of a collection of legacy files.
- Remove `save_deccam_layered_image` which produces fits files that are not compatible with the standardizer.
- Removing the saving portion of the fake data creation notebook.


DinoBektesevic (approved):
Looks good to me

PR #694, opened 2024-09-04
Fix the wrongly oriented BBox in IC.

DinoBektesevic:
This, hopefully, finally correctly orients the BBox in the image collections. 
PR also:

1. Fixes the default astropy bbox calculation method
2. Flips the wrongly mirrored bottom left and bottom right corners
3. Flips the mirrored naxis keywords in between the SkyWcs and Astropy's WCS methods
4. Makes a dictionary out of the PropertyList object returned by the stack so that standardizer can be re-used without exhausting the generator values.
5. Adds more explicit documentation about what corner is what 

Sorry about the mess.


wilsonbb (approved):
Thanks!

PR #697, opened 2024-09-06
Get rid of the legacy demo data

jeremykubica:
Convert the notebooks and `test_end_to_end.py` to use a saved `WorkUnit` from the create_fake_data notebook instead of the directory of legacy files demo/. The data in that directory was out of data in terms of format (e.g. lacked "IMAGE" label on that layer). This will make it easier to keep the notebooks up to date, because we can directly generate the data from one of the notebooks.

This is a partial step toward removing the legacy readers in favor of using the standardizers.


data/demo_data.fits
DinoBektesevic, line 1: I'm not in favor of adding 15.6MB large file to this repo that will will then lug around in perpetuity (at least when it comes to the repo history and repo diff size). This is already a quarter of a GB large repo due to some binary files that were committed very early in its history.

Is there any other way?

jeremykubica: That's a good point. We can create WorkUnit dynamically and save it locally. It will add a little time to the test, but avoid storing the file.

I've made that change.

DinoBektesevic (approved):

PR #704, opened 2024-09-12
Clustering cleanup

jeremykubica:
This PR overhauls how the clustering options and configurations to make them less dependent on the input data. Changes include:
- The "all" clustering type now uses unnormalized position and velocity (x0, y0, vx, vy) instead of normalized position, velocity magnitude, and velocity angle. This means clustering eps does not change with image size.
- Drop the normalization option from all searches. The clustering should not depend on image size.
- Add an explicit "cluster_v_scale" parameter to control the tradeoff of position vs velocity instead of using the search parameter range. This is more flexible and stable over different searches.
- Rename `eps` to the more meaningful `cluster_eps`.
- Set `cluster_eps` default to 20.0 to account for the lack of normalization in the searches.
- Remove broken benchmark for clustering.


src/kbmod/filters/clustering_filters.py
```diff
@@ -145,49 +126,30 @@ def _build_clustering_data(self, result_data):
...
-
-        self.v_scale = (vel_lims[1] - vel_lims[0]) if vel_lims[1] != vel_lims[0] else 1.0
-        self.a_scale = (ang_lims[1] - ang_lims[0]) if ang_lims[1] != ang_lims[0] else 1.0
-        self.min_v = vel_lims[0]
-        self.min_a = ang_lims[0]
-
+        super().__init__(cluster_eps, **kwargs)
+        if cluster_v_scale <= 0.0:
+            # Avoid divide by zero.
+            cluster_v_scale = 1e-12
```
DinoBektesevic, line 151: Is this supposed to be == 0? But also if the scaling is 0, then it's equivalent to setting the velocities to 0 and defaulting to position-only matching, and that's shown in line 172 and 173 where they get multiplied together (velocities and v_scaling) so I'm a bit confused where does the division zero come to play?

DinoBektesevic (changes requested):
I think this is worth it just for the name changes, but I've a question regarding the velocity_scale == 0 situation.

src/kbmod/filters/clustering_filters.py
jeremykubica: That's an error. I was originally scaling by 1/cluster_v_scale (to stay consistent with the old scales), but figured a multiplicative scaling makes more sense. I forgot to remove that check.

I've fixed it now.

DinoBektesevic (approved):

PR #706, opened 2024-09-13
Cleanup trajectory generation configuration

jeremykubica:
Overhaul the configuration for trajectory generation:

- Removes the legacy parameters such as `ang_arr` and `average_angle`
- Creates a new `EclipticSearch` generator based on Dino's approach that uses the intuitive definition of the parameters.
- Makes the `EclipticSearch` generator the default throughout the code.
- Adds `compute_ecliptic_angle()` helper to the `WorkUnit`


src/kbmod/trajectory_generator.py
```diff
@@ -314,25 +312,171 @@ def __init__(self, v_arr, ang_arr, average_angle, *args, **kwargs):
...
+    def __init__(
+        self,
+        vel_steps=0,
+        min_vel=0.0,
+        max_vel=0.0,
+        ang_steps=0,
+        min_ang_offset=0.0,
+        max_ang_offset=0.0,
+        angle_units="radians",
+        force_ecliptic=None,
```
DinoBektesevic, line 371: I'm not a big fan of this for a few reasons. 

- we're flipping the convention for no reason. Everywhere in Python land slices are ubiquitous objects and they're always `(start, stop, stride)` not `(stride, start , stop)`
- I'd make them a slice, tuple, list, frankly anything except the separate kwarg:
    ```
    velocities: (start, stop, stride)
    angles: (start, strop, stride)
    unit: radian
    ecliptic_angle: None
    ```
    which helps with the naming - ang_offset from what is a bit weird.  Or search_angle, force_ecliptic_angle, force_search_angle here could work if we name the class better.
- Ecliptic trajectory is not an ecliptic trajectory, it;'s a `EclipticCenteredSearch`, `EclipticSearch`, or `EclipticBasedSearch` or something so that the change of the naming scheme doesn't make argument names ambiguous anymore  This is not more confusing than the current setup in use I think, for example:
    ```
    with EclipticCenteredSearch(
        velocities=(0, 100, 10),
        angles=(-15, 15, 30), 
        angle_units="degree",    # the hidden default is velocity_units = u.pixel/u.day
        wcs=wcs
    ) as gen:
        trj = gen.generate()
   ```
    or 
    ```
    with EclipticCenteredSearch(
        velocities=(0, 100, 10),
        angles=(-15, 15, 30), 
        ecliptic_angle = 10*u.degree  # or like I said above, force_ecliptic_angle, search_angle etc...
    ) as gen:
        trj = gen.generate()
   ```
- I'd name the units in the same names as the astropy uses them because then: you can:
   ```
    if not isinstance(angles, u.Quantity):
        angles = angles * u.Unit(angle_units)
    angles = angles.to(u.rad).values

    if isinstance(velocities, u.Quantity):
        velocities = angles.to(u.pixel/u.days).values
    ````
    The first hit here is `u.Unit(angle_unit)` - if the names match, degree/radian/arcsecond - then all you get back is the `u.degree`, `u.radian`, `u.arcsecond` object. Second hit here is that we can specify pixels/hour etc. and still force the velocities to per day under the hood (because that's what the mjd timestamps are). So those two are nice-to-haves. For example:
    ```
    with EclipticTrajectory(
        velocities=(0, 100, 10) * u.pixel/u.hour,
        angles=(-15, 15, 30) * u.degree, 
        wcs=wcs
    ) as gen:
        trj = gen.generate()
   ```
    should work. 

I get that `wcs` can't be passed into the trajectory gen because of how the code right now works - but this is really the place for the `calc_eclip_angle` to live as either a class method or just a method so I wanted to point that out. Ultimately perhaps they should all get a "work_unit" argument in the init or something and then ignore it if they don't use it or something, but otherwise derive the values they need.


    

I don't know how much of this you feel like fixing, I have to go ahead and merge all these changes and run a test so I can make sure everything works for tomorrow. We can live with the current sampler in that scenario if you want to accept some or all these changes.

DinoBektesevic (changes requested):
Some changes requested.

src/kbmod/trajectory_generator.py
jeremykubica: I made most of the changes you recommend. Specifically I renamed the generator `EclipticCenteredSearch` and made the velocity and angle information into vectors. 

The one change I did not make was using astropy units. This doesn't work naturally with how we save the configuration data into the fits header. But there is probably a cleaner way to do that than I tried.

DinoBektesevic: Even if we assume the default units in rad and pixel/day? All the casts happen in the init so I'm surprised by this.

src/kbmod/trajectory_generator.py
```diff
@@ -314,25 +312,148 @@ def __init__(self, v_arr, ang_arr, average_angle, *args, **kwargs):
...
+        The maximum search angle in the image (ecliptic_angle + max_ang_offset) in radians.
+    """
+
+    def __init__(
+        self,
+        velocities=[0.0, 0.0, 0],
+        angles=[0.0, 0.0, 0],
+        angle_units="radians",
+        given_ecliptic=None,
+        computed_ecliptic_angle=None,
```
DinoBektesevic, line 362: Is there a point to differentiating between these two? I get having a flag in search config, but doesn't really seems necessary here.

DinoBektesevic (approved):

src/kbmod/trajectory_generator.py
jeremykubica: I think I have a better way to do this, which I will implement tomorrow.

src/kbmod/trajectory_generator.py
jeremykubica: The init function takes a dictionary that is extracted from the base `SearchConfiguration` object. The difficulty is that this base configuration itself can be stored in different ways, including in the HDU header ([example](https://github.com/dirac-institute/kbmod/blob/e01f7752afa5bfde41b1f8a31509f38811629afc/src/kbmod/work_unit.py#L268)). We could change how we store the configuration in a fits file to better support this if we want more than radians and degrees.

src/kbmod/trajectory_generator.py
jeremykubica: I changed the `TrajectoryGenerator` to take an optional reference to the `WorkUnit` so it can extract any information that it needs.

src/kbmod/trajectory_generator.py
```diff
@@ -53,7 +49,12 @@ def create_trajectory_generator(config):
     logger = logging.getLogger(__name__)
     logger.info(f"Creating trajectory generator of type {name}")
 
-    return TrajectoryGenerator.generators[name](**config)
+    # Add any keyword arguments to the params, overriding current values.
+    params = copy.deepcopy(config)
+    params.update(kwargs)
+    logger.debug(str(params))
+
+    return TrajectoryGenerator.generators[name](**params, work_unit=work_unit)
```
DinoBektesevic, line 57: I had no expectation of this being a part of this PR, I was just thinking aloud of what is the most generic common interface. Nice!

src/kbmod/trajectory_generator.py
```diff
@@ -314,25 +315,153 @@ def __init__(self, v_arr, ang_arr, average_angle, *args, **kwargs):
...
+            elif angle_units[:3] == "rad":
+                ecliptic_angle = given_ecliptic
+            else:
+                raise ValueError(f"Unknown angular units {angle_units}")
+        elif work_unit is not None:
+            # compute_ecliptic_angle() always produces radians.
+            ecliptic_angle = work_unit.compute_ecliptic_angle()
+            print(f"Using WU = {ecliptic_angle}")
+        else:
+            logger = logging.getLogger(__name__)
```
DinoBektesevic, line 402: You should make 1 logger per module, so this should go to the top of the file.

src/kbmod/trajectory_generator.py
```diff
@@ -314,25 +315,153 @@ def __init__(self, v_arr, ang_arr, average_angle, *args, **kwargs):
...
+            can be used to derive parameters that depend on the input.
+        """
+        super().__init__(**kwargs)
+
+        if given_ecliptic is not None:
+            if angle_units[:3] == "deg":
+                ecliptic_angle = given_ecliptic * (math.pi / 180.0)
+            elif angle_units[:3] == "rad":
+                ecliptic_angle = given_ecliptic
+            else:
```
DinoBektesevic, line 395: I'm not sure I fully understand why we can't check for quantity and always cast to radians and pixel/day internally and then strip the unit? I'm not going to be a stickler about it, this is fine, but I am curious.

DinoBektesevic (approved):
Move the logger to the top of the file and merge the PR.

src/kbmod/trajectory_generator.py
DinoBektesevic: I'm still not clear tbh, I gave you an example above all it would require to store is a string and a float:

```python
>>> import astropy.units as u
>>> a = {"min":1, "max":2, "units":"degree"}
>>> a["min"] * u.Unit(a["units"])
<Quantity 1. deg>
>>> a = {"min":1, "max":2, "units":"radian"}
>>> a["min"] * u.Unit(a["units"])
<Quantity 1. rad>
>>> a = {"min":1, "max":2, "units":"pixel/h"}
>>> a["min"] * u.Unit(a["units"])
<Quantity 1. pix / h>
>>> a = {"min":1, "max":2, "units":"pixel/day"}
>>> a["min"] * u.Unit(a["units"])
<Quantity 1. pix / d>
```

Inside our class, because we can't pass quantities to the core search, strip the units and send forward a float. 
```python
>>> a = {"min":1, "max":2, "units":"pixel/h"}
>>> ((a["min"] * u.Unit(a["units"]))).to(u.pixel/u.day)
<Quantity 24. pix / d>
>>> ((a["min"] * u.Unit(a["units"]))).to(u.pixel/u.day).value
24.0
```

If we don't feel like making sure we serialize different units we don't even have to and we could just make it so we only ever write the units in their default units and just say so out loud and still let quantities in so that in interacive plotting use one can at least experiment with the values. regardless, writing the unit is just:

```python
>>> b = (a["min"] * u.Unit(a["units"]).to(u.pixel/u.day)
>>> str(b.unit)
'pix / d'
```

All we ever store are some strings and floats which doesn't seem incompatible with header units or dicts, or yamls etc...

jeremykubica: Got it. I misinterpreted your suggestion as passing the astropy units as the parameter (instead of parsing a string with astropy.units). Your approach is much cleaner and more flexible. I've made that change for the angular units.

We don't currently pass velocity units. Do you think it would be worthwhile? I'm happy to add that functionality if so.

DinoBektesevic: I feel like I'm the wrong person to ask this at this point. I've been in pixels/day so long that, I'm afraid to say, it started looking normal to me. I have no idea whether or not workunit is really helpful or not. Just when I wrote the ecliptic centered search to test the old one, it felt like it. Add the units, it's literally a 2 line if statement and a parameter, if it turns out not useful I don't think it's a big deal and a lot of effort wasted.

jeremykubica: Added.

DinoBektesevic: Let's merge this finally :)

PR #711, opened 2024-09-17
Add a few more checks and logging to EclitipicCenteredSearch

jeremykubica:


drewoldag (approved):
This all seems quiet reasonable. Thanks!

PR #712, opened 2024-09-18
Move Psi/Phi curve generation out of Python loop

jeremykubica:
Move the generation of the psi and phi curves from inside a Python loop to inside a C++ loop. This is a very small speedup, but also adds very little complexity.

This PR also:

-  breaks up some of the giant `test_search.py` by moving a few tests to `test_stack_search_results.py`
-  Applies clang formatting to the C++ files.


src/kbmod/search/stack_search.cpp
```diff
@@ -256,17 +256,33 @@ std::vector<float> StackSearch::extract_psi_or_phi_curve(Trajectory& trj, bool e
     return result;
 }
 
+std::vector<std::vector<float> > StackSearch::get_psi_curves(std::vector<Trajectory>& trajectories) {
```
gitosaurus, line 259: Any reason why the input isn't `const &`?

jeremykubica: No particularly good reason. I made that change (and also updated some of the other functions).

tests/test_stack_search_results.py
```diff
@@ -0,0 +1,147 @@
...
+        self.assertEqual(psi_curves.shape[1], self.num_times)
+        self.assertTrue(np.all(psi_curves > 0.0))
+
+        phi_curves = np.array(self.search.get_phi_curves(self.fake_trjs))
+        self.assertEqual(phi_curves.shape[0], self.num_objs)
+        self.assertEqual(phi_curves.shape[1], self.num_times)
+        self.assertTrue(np.all(phi_curves > 0.0))
+
+        # Check that the batch getters give the same results as the iterative ones.
+        for i in range(self.num_objs):
```
drewoldag, line 70: Nice, looks good.

drewoldag (approved):
This all looks pretty reasonable to me.

Issue #714, opened 2024-09-19
Make StampCreator into functions instead of a class of static functions.


jeremykubica:
The `StampCreator` class in search/stamp_creator.cpp doesn't do anything that requires it to be a class and just adds overhead when we are using it from Python (creating an unneeded object). We should remove the class and just call the functions directly.

Issue #713, opened 2024-09-19
Move a bunch of the stamp tests from `test_search.py` to `test_stamp_creator.py`


jeremykubica:
The file `test_search.py` ended up as a dumping ground for a lot of random tests. We should break out the stamp creation tests into the `test_stamp_creator.py` file.

PR #715, opened 2024-09-19
Add the ability to generate variance scaled stamps

jeremykubica:
Add the ability to generate variance scaled stamps using on Steven's formulation. Adds the ability to uses the weighted stamps for both filtering and to store as auxiliary stamps.

This PR made me realize that we need two followups:

- Move a bunch of the stamp tests from `test_search.py` to the new `test_stamp_creator.py` to have better organization. (#713)
- Make StampCreator into functions instead of a class of static functions. (#714)

I'm holding off from both those changes so as to avoid burying the logic change in this PR with clean ups.


src/kbmod/search/stamp_creator.cpp
```diff
@@ -68,7 +68,8 @@ std::vector<RawImage> StampCreator::get_coadded_stamps(ImageStack& stack, std::v
     rs_logger->info("Performing stamp filtering on " + std::to_string(t_array.size()) + " trajectories.");
     DebugTimer timer = DebugTimer("stamp filtering", rs_logger);
 
-    if (use_gpu) {
+    // We use the GPU if we have it for everything except STAMP_VAR_WEIGHTED which is CPU only.
```
maxwest-uw, line 71: is being CPU only a limit of the formula for the weighted var stamps or are we just holding off on the gpu implementation for now? I'm fine either way :)

maxwest-uw (approved):
lgtm !

src/kbmod/search/stamp_creator.cpp
jeremykubica: It's a "holding off for now". The GPU version is a bit more tricky because, unlike the other stamp functions, we will need to transfer the variance layers to the GPU as well. Instead of doing all that machinery now, I wanted to put something in place that we could test. If it is useful and a bottleneck, we can optimize on the GPU.

PR #716, opened 2024-09-20
Move obstime from RawImage to LayeredImage

jeremykubica:
A very small step toward cleaning up the C++ code and removing the `RawImage` class. Move the storage of obstime from the `RawImage` to the `LayeredImage`.


DinoBektesevic (approved):
I'll be honest at best I gave it a provisional look-over, but it looks ok.

PR #717, opened 2024-09-20
Upgrade to numpy 2.0

### LINKED ISSUE #658: Support numpy v2.0
Using numpy 2.0 results in an immediate segfault for KBMOD. Determine what is breaking, fix it, and unpin the numpy version.

jeremykubica:
Closes #658 

Upgrade to support numpy 2.0 and above. This requires pulling new versions of both eigen and pybind11 using `git submodule update --remote`. We add minimum versions for a bunch of packages, such as `reproject` and `astropy-healpix`.


jeremykubica:
Rolling this back because it is failing in some weird places

PR #719, opened 2024-09-20
Cleanup unused code

jeremykubica:
- Remove code for unused visualization that has not be used in a long time. This was loading from files that we have no outputted in a while such as a file of the individual psi and phi curves.
- Remove the`FileUtils` class. Move the functionality for saving a list of trajectories into `Results` (the only place it is used).


wilsonbb (approved):
Lots of red here, LGTM!

PR #728, opened 2024-10-29
Move tests from test_search to test_stamp_creator

### LINKED ISSUE #713: Move a bunch of the stamp tests from `test_search.py` to `test_stamp_creator.py`

jeremykubica:
Closes #713 

Move a bunch of tests from one file to another so they tests are better organized. This is almost completely cut and paste, with a few naming changes because the setups in the files used the same names.


wilsonbb (approved):
LGTM

PR #738, opened 2024-11-07
Pass the location information to the WorkUnit

jeremykubica:
This is a step in better propagation of provenance information. Pass information from the standardizer's `location` field to the `WorkUnit`. This might not provide perfect information about the data source. For example in memory fits files will just produce ":memory:".

### DIFF
--- a/src/kbmod/image_collection.py
+++ b/src/kbmod/image_collection.py
@@ -885,10 +885,18 @@ class ImageCollection:
 
         logger.info("Building WorkUnit from ImageCollection")
         layeredImages = []
+        locations = []
         for std in self.get_standardizers(**kwargs):
+            location = std["std"].location
             for img in std["std"].toLayeredImage():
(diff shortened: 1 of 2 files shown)


DinoBektesevic:
This is probably not the right way to do it. I would use `toBinTableHDU` and append it to the HDUList of a WorkUnit. If you want to add an `imagecollection` property to the WorkUnit there's a `fromBinTableHDU` on an imagecollection. As you note `location` is not a unique init key required by Standardizers.

This was blocked for a bit by messy metadata tracking within WorkUnit. Some of this was cleaned up so I guess we can do it. It can't hurt to add it, but I've seen WU with 700+ HDUs. It's probs more worth to clean that up in comparison, mostly because we don't keep WUs around during/after processing due to their size.

jeremykubica:
Closing since Dino indicates there is a better way to do this.

PR #739, opened 2024-11-07
Refactor the image-level meta data in WorkUnit

jeremykubica:
As a step toward adding more provenance data in the workflow (specifically traveling through the `WorkUnit`), I regrouped a bunch of the data about constituent images into a single astropy Table. I think this makes it a little easier to track and reason about, but I could also make a case that this adds unneeded complexity. So let me know if you think this is worth it.

### DIFF
--- a/src/kbmod/reprojection.py
+++ b/src/kbmod/reprojection.py
@@ -220,7 +220,7 @@ def _reproject_work_unit(
             variance = image.get_variance()
             mask = image.get_mask()
 
-            original_wcs = frame[index]
+            original_wcs = wcs_list[index]
             if original_wcs is None:
                 raise ValueError(f"No WCS provided for index {index}")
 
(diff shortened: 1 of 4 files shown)


DinoBektesevic:
Hah, funny, sorry to double dip but I believe these two PRs might be related: https://github.com/dirac-institute/kbmod/pull/738

If we're already collating into BinTables from Headers - de-duplicating things like time stamps, per-image WCS et. al. via imagecollection HDU could be helpful here.

jeremykubica:
> Hah, funny, sorry to double dip but I believe these two PRs might be related: #738
> 
> If we're already collating into BinTables from Headers - de-duplicating things like time stamps, per-image WCS et. al. via imagecollection HDU could be helpful here.

I agree that there is more cleanup we can do, especially with de-duplicating things. I was didn't want to make any changes to the serialized data in this PR though, since I don't want to introduce any incompatibilities with WorkUnits currently living on disk. I think we can do that in a future round.

I would need to talk through how to use the imagecollection HDU correctly.

maxwest-uw (approved):
much cleaner!

PR #740, opened 2024-11-12
Pass additional metadata to WorkUnit

jeremykubica:
Extract arbitrary metadata from the standardizer's `standardizeMetadata()` function and save it to the `WorkUnit`. If the meta data is not present for that standardizer, this will just fill in `None`. By default we extract "visit" and "filter".


jeremykubica:
Closing as this was largely replaced by #742

Issue #656, opened 2024-06-15
Need to preserve visit as part of the provenance.


ColinOrionChandler:
As I'm working on linking results to known objects it's clear that preserving the visit IDs would be extremely useful as an integer for lookup rather than trying to match datetimes to get back to the original pointing (RA, Dec) values.

PR #742, opened 2024-11-15
Save metadata

### LINKED ISSUE #410: Add provenance information
Incorporate provenance information into the `ImageCollection` and `WorkUnit` to track where the images came from and how they were processed.

jeremykubica:
Save the metadata table as its own HDU (instead of saving each individual item in fits headers). This should simplify the addition of new metadata or the removal of unused metadata. This PR does not remove the ability to read the previous fits headers as we want to preserve compatibility with any saved data.

This also incorporates the changes from #740 to pass metadata from an `ImageCollection` to a `WorkUnit`. It current passes the visit field (as requested in #656).

Closes #410

### DIFF
--- a/src/kbmod/image_collection.py
+++ b/src/kbmod/image_collection.py
@@ -888,7 +888,10 @@ class ImageCollection:
         for std in self.get_standardizers(**kwargs):
             for img in std["std"].toLayeredImage():
                 layeredImages.append(img)
+
         imgstack = ImageStack(layeredImages)
+        img_metadata = Table()
+
         if None not in self.wcs:
-            return WorkUnit(imgstack, search_config, per_image_wcs=list(self.wcs))
-        return WorkUnit(imgstack, search_config)
+            img_metadata["per_image_wcs"] = list(self.wcs)
+        return WorkUnit(imgstack, search_config, org_image_meta=img_metadata)
--- a/src/kbmod/image_collection.py
+++ b/src/kbmod/image_collection.py
@@ -884,14 +884,40 @@ class ImageCollection:
         from .work_unit import WorkUnit
 
         logger.info("Building WorkUnit from ImageCollection")
+
+        # Create a storage location for additional image metadata to save.
+        # Not all of this may be present in from the ImageCollection,
+        # in which case we will append None.
+        metadata_vals = {
+            "visit": [],
+        }
+
+        # Extract data from each standardizer and each LayeredImage within
+        # that standardizer.
         layeredImages = []
         for std in self.get_standardizers(**kwargs):
+            num_added = 0
             for img in std["std"].toLayeredImage():
                 layeredImages.append(img)
+                num_added += 1
 
-        imgstack = ImageStack(layeredImages)
-        img_metadata = Table()
+            # Get each meta data value from the standardizer so it can be
+            # passed to the WorkUnit. Use the same value for all images from
+            # this standardizer.
+            metadata = std["std"].standardizeMetadata()
+            for col in metadata_vals.keys():
+                value = metadata.get(col, None)
+                metadata_vals[col].extend([value] * num_added)
 
+        # Append the WCS information if we have it.
         if None not in self.wcs:
-            img_metadata["per_image_wcs"] = list(self.wcs)
-        return WorkUnit(imgstack, search_config, org_image_meta=img_metadata)
+            metadata_vals["per_image_wcs"] = list(self.wcs)
+
+        # Save the metadata as a table, but prune it if empty.
+        image_metadata = Table(metadata_vals) if len(metadata_vals.keys()) > 0 else None
+
+        # Create the basic WorkUnit from the ImageStack.
+        imgstack = ImageStack(layeredImages)
+        work = WorkUnit(imgstack, search_config, org_image_meta=image_metadata)
+
+        return work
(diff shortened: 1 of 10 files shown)


src/kbmod/image_collection.py
```diff
@@ -884,11 +884,40 @@ def toWorkUnit(self, search_config=None, **kwargs):
...
+            metadata_vals["per_image_wcs"] = list(self.wcs)
+
+        # Save the metadata as a table, but prune it if empty.
+        image_metadata = Table(metadata_vals) if len(metadata_vals.keys()) > 0 else None
+
+        # Create the basic WorkUnit from the ImageStack.
+        imgstack = ImageStack(layeredImages)
+        work = WorkUnit(imgstack, search_config, org_image_meta=image_metadata)
+
+        return work
```
DinoBektesevic, line 923: I'm confused as to why exporting the image collection as a table header unit doesn't work? 

Why doesn't
```
metadata = self.toBinTableHHDU()
return WorkUnit(....., imgcoll=metadata)
```

and then appending that HDU to WorkUnit in the constructor or something work?

jeremykubica: That's a good point. I missed that function (and thought you had meant to pull the data from the metadata functions). Using `toBinTableHDU` does what I need a lot cleaner. I changed the PR to use this.

I'm still using an astropy Table in the `WorkUnit` because we are adding other things to it (from the reprojection code and possibly other places). I know it could all be done as an HDU, but I don't see majors advantages to changing it (unless I would be able to avoid the WCS serialization hack).

DinoBektesevic: From my perspective this i over-complicated because you're completely reinterpreting all of the columns, adding "visit", wcs etc... [Except the required columns](https://github.com/dirac-institute/kbmod/blob/7f045b5fbda5441316dd32ec24b1f3764e11296f/src/kbmod/image_collection.py#L152) you don't know what might or might not exist.                                                                
Visit for DECam is what snapshot is for Rubin, and a basically unknown metadata for any other telescope.             
You don't know where the metadata comes from either, it isn't necessarily extracted from the image data (or image data only).               
Ultimately, all of this was was already once standardized to make the image collection.

I don't really understand what the confusion about using the HDU is? It's still just a table. 

```
>>> import kbmod as kb
>>> ic = kb.ImageCollection.read("20190402_A0a_062.collection")
>>> hdu = ic.toBinTableHDU()

>>> hdu.data["ra"]
array([214.97630148, 214.97149992, 214.9693644 , 214.97048054,
       214.97599117, 214.96944586])
>>> ic["ra"]
<Column name='ra' dtype='float64' length=6>
  214.976301477001
 214.9714999235584
214.96936440036148
214.97048053812478
214.97599116659285
 214.9694458550695
845576

>>> hdu.data["visit"]
array([845684, 845574, 845685, 845575, 845683, 845576])
>>> ic["visit"]
<Column name='visit' dtype='int64' length=6>
845684
845574
845685
845575
845683
845576

>>> hdu.data["obsid"]
chararray(['ct4m20190403t092726', 'ct4m20190403t045350',
           'ct4m20190403t093005', 'ct4m20190403t045619',
           'ct4m20190403t092457', 'ct4m20190403t045848'], dtype='<U19')
>>> ic["OBSID"]
<Column name='OBSID' dtype='str19' length=6>
ct4m20190403t092726
ct4m20190403t045350
ct4m20190403t093005
ct4m20190403t045619
ct4m20190403t092457
ct4m20190403t045848
```

```
>>> from astropy.table import Table
>>> tableagain = Table(hdu.data)
>>> tableagain
<Table length=6>
               dataId                ...
               str36                 ...
------------------------------------ ...
1f9b7276-c3e9-4656-aade-3e79b9a9adde ...
9751cec2-e84b-41f4-bf77-29e5d331f10b ...
6479fd75-414b-4b70-8a86-2c06d18d2b18 ...
c2e2c32d-073c-4731-a09b-b90e8a683dbf ...
b488c68a-d9c7-4156-8798-b996fdcef015 ...
182a9ec2-f341-45ab-ab8b-10a972d0eff8 ...
```
It already does everything you want it to do? If that's a no-go for some reason, cast it as a `Table` on `load`/`write` or, less preferable, just run `unpack` and then export the `.data` table instead of reinterpreting the values again. 

I kept mentioning the BinTableHDU because WorkUnit is *just* a FITS file I figured it'd make your life a bit easier when writing it to disk and/or sharding it, all you;'d have to do is `append` it to an `HDUList`. If you don't want it - ok, unpack and pass `.data` onward. When you go to write the WorkUnit to disk, you'll have to make it a `TableHDU` again anyhow.

jeremykubica: Are you saying that the current implementation is overcomplicating or the original implementation? The current implementation just calls `toBinTableHDU()`, casts that as a `Table`, and passes it along to the `WorkUnit`. I got rid of all the logic with custom extracting of the metadata. You are right that, I could leave it as an HDU, but the workunit already uses a Table to store the reprojection related metadata, so I went that direction. 

Part of my misunderstanding with `toBinTableHDU()` is I didn't realize that it would unpack so that each column had the same number of entries as the `ImageStack`.

DinoBektesevic: I didn't see the update, the initial code didn't work like that. The `unpack` will flatten the table so that it can be written down in any other format that doesn't support shared metadata-as-a-comment field format, f.e. TableHDUs or Parquet. 

The update looks find to me. I don't mind it being a `Table` or renaming of the `wcs` column in the WorkUnit.

PR #743, opened 2024-11-15
Add a pencil search TrajectoryGenerator

jeremykubica:
Add a `TrajectoryGenerator` that does a pencil search around a given velocity. This is used for analysis and is the first step in putting together a new of searches that Dino requested.


wilsonbb (approved):
LGTM!

PR #745, opened 2024-11-18
Add a neighborhood search to the TrajectoryExplorer

jeremykubica:
Add a search that will perform a hyper-localized search.  This search effectively uses a small neighborhood around a given trajectory (both in terms of starting pixel and velocities) and returns all results from this neighborhood. This localized set can be used to:
- refine trajectories by searching a finer parameter space around the best results found by the initial search, or
- collect a distribution of trajectories and their likelihoods around a single result.

As part of this change we remove a small optimization from the core search (which did not bother to sort trajectories with LH <= -1.0 since we knew they would be filtered eventually). This removes that optimization so we can get the full grid of nearby trajectories.

### DIFF
--- a/src/kbmod/run_search.py
+++ b/src/kbmod/run_search.py
@@ -19,6 +19,54 @@ from .work_unit import WorkUnit
 logger = kb.Logging.getLogger(__name__)
 
 
+def configure_kb_search_stack(search, config):
+    """Configure the kbmod SearchStack object from a search configuration.
+
+    Parameters
+    ----------
+    search : `kb.StackSearch`
+        The SearchStack object.
+    config : `SearchConfiguration`
+        The configuration parameters
+    """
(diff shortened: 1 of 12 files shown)


src/kbmod/run_search.py
```diff
@@ -20,6 +20,54 @@
 logger = kb.Logging.getLogger(__name__)
 
 
+def configure_kb_search_stack(search, config):
+    """Configure the kbmod SearchStack object from a search configuration.
```
wilsonbb, line 24: Here we're referring to the StackSearch object as a SearchStack

Presumably should also rename method to configure_kb_stack_search

wilsonbb (approved):
Looks good, just a small comment

DinoBektesevic:
Hold on with this please. I am using this branch.

PR #750, opened 2024-12-13
Implement SigmaG filtering in JAX

jeremykubica:
Change the numpy vectorized implementation of sigma G filtering to use JAX and vmap. This produced roughly a 2000x speedup on sigma-G filtering on timing tests on Baldur.

Also makes one behavior change. Previously sigma-G filtering with clipped_negative=True would filter out zero values of LH (but NOT all negative values). This does not make sense with the goal of sigma-G, so I have removed it.


wilsonbb (approved):
LGTM!

PR #752, opened 2024-12-16
Cleanup unused code in LayeredImage

jeremykubica:
Remove two unused functions from `LayeredImage`:
- `subtract_template()` is from when image differencing was done in KBMOD (before version 0.5)
- `compute_fraction_masked()` is a debugging function that was not very useful.


drewoldag (approved):
Looks good. Thanks for the clean up!

PR #754, opened 2024-12-16
Upgrade to numpy 2.0

jeremykubica:
Upgrade the eigen and pybind11 submodules (using `git submodule update --remote`) and upgrade to numpy >2.0. A lot of minimum versions need to change to ensure all the dependencies (direct and indirect) are compatible with numpy 2.0.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,16 +32,17 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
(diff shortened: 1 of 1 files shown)


drewoldag (approved):
This all seems reasonable. I checked the CI logs because I would assume that any trouble would show up there - everything looked fine there. Nice job.

PR #758, opened 2024-12-17
Remove verification step for standardizer test

jeremykubica:
The verification step is failing for this specific mocked data for the standardizer tests with astropy 7.0.


jeremykubica:
Full fix in in #760

PR #761, opened 2024-12-17
functionalize and remove the StampCreator class

### LINKED ISSUE #714: Make StampCreator into functions instead of a class of static functions.

maxwest-uw:
Removes the `StampCreator` and leaves all of its methods as functions

resolves #714 

(was going through the backlog looking for something to do and couldn't resist the urge to make something more functional)


jeremykubica (approved):
One blocking change to make (the benchmark) and a bunch of small stylistic fixes to make before merging.

PR #763, opened 2024-12-18
Breaking out helper functions

jeremykubica:
Break out some helper functions from the regression test so I can reuse them in some notebooks for testing speedups to filtering and clustering. 

Also adds a `copy()` method to `Results` to enable easier testing.

### DIFF
--- a/src/kbmod/trajectory_utils.py
+++ b/src/kbmod/trajectory_utils.py
@@ -15,6 +15,7 @@ import numpy as np
 
 from astropy.coordinates import SkyCoord
 from astropy.wcs import WCS
+from scipy.optimize import linear_sum_assignment
 
 from kbmod.search import Trajectory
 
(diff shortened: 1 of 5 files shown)


src/kbmod/trajectory_utils.py
```diff
@@ -263,3 +264,99 @@ def evaluate_trajectory_mse(trj, x_vals, y_vals, zeroed_times, centered=True):
     # Compute the errors.
     sq_err = (x_vals - pred_x) ** 2 + (y_vals - pred_y) ** 2
     return np.mean(sq_err)
+
+
+def ave_trajectory_distance(trjA, trjB, times=[0.0]):
```
wilsonbb, line 269: Optional nit:
I think avg_trajectory_distance is clearer than the old name

wilsonbb (approved):
Approving with an optional suggestion to rename one of the moved functions

PR #764, opened 2024-12-18
Vectorize match_trajectory_sets

jeremykubica:
The new `match_trajectory_sets()` function is running too slowly in even small clustering tests. This vectorizers it. It also breaks out the logic for extracting column information from a list of trajectories, makes that a module, and calls it from `Results`.


drewoldag (approved):
Looks great to me. It looks like `match_trajectory_sets` already had reasonable test coverage, prior to this change, so no major concerns about a regression here.

PR #766, opened 2024-12-20
Small numpy changes

jeremykubica:
Use `asarray` and `full` to remove some unneeded data copies. This doesn't really impact the performance, but cleans up the code a little.


drewoldag (approved):
Now I know why I should us `asarray`! This looks great, thank you!

PR #770, opened 2025-01-10
Have KBMOD append RA, dec columns if possible

jeremykubica:
If we provide KBMOD a `WorkUnit` with wcs values, append the EBD (RA, dec) and the original (RA, dec) where possible.

### DIFF
--- a/src/kbmod/run_search.py
+++ b/src/kbmod/run_search.py
@@ -346,6 +346,8 @@ def append_ra_dec_to_results(workunit, results):
 
     # Compute the predicted (RA, dec) positions for each trajectory in global space.
     if workunit.wcs is not None:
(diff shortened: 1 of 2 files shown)


maxwest-uw (approved):
this all looks good! the only issue that I can think of is that if the `WorkUnit` was reprojected with a common WCS that wasn't in EBD it could cause it to return some weirdness when using the `image_positions_to_original_icrs` function, but I don't think we need to cover that scope fully in this pr since that's not a typical use pattern afaik

jeremykubica:
Thanks @maxwest-uw . Long term, I think we will want to support non-EBD reprojections. I created Issue #772 and assigned it to you. I'm happy to discuss what is needed.

PR #776, opened 2025-01-17
Bring documentation up to date

jeremykubica:
This is the first PR in a batch of updates to the documentation. This brings it up to date with overall flow of KBMOD. More detail is needed for all of the stages, but will be included in later PRs so that we can iterate quickly.


maxwest-uw (approved):
two small grammar changes, but other than that great!!

PR #777, opened 2025-01-17
Remove unused legacy file format

jeremykubica:
Since we have switched to the `Results` table output, we do not use the legacy file formats anymore. This removes them to reduce code complexity.


drewoldag (approved):
This seems pretty reasonable to me. Good clean up!

Issue #774, opened 2025-01-14
num_obs check and dynamic setting


ColinOrionChandler:
Dynamically adjusting num_obs would be great, like a fraction, as the numbers will vary for Rubin especially.

Also, as a bare minimum, let's throw an exception if num_obs is > than number of images.

jeremykubica:
There are two ways we could set this dynamically:
1) We overload `min_obs` so that if it is (0.0, 1.0) it is treated as a fraction. The concern is that this could cause confusion with the user since the parameter means two different things.
2) We could add another parameter `fraction_obs` which specified `min_obs` as a fraction. The question is what to do if both are given? Take the smaller one? Give an error?

Do you have preferences?

PR #778, opened 2025-01-17
Add a check the num_obs is valid

jeremykubica:
Partially addresses #774 by checking that the minimum number of observations <= `min_obs`


drewoldag (approved):
Looks good.

PR #781, opened 2025-01-22
ButlerStandardizer should allow optional metadata headers to be missing

wilsonbb:
On the USDF embargo butler, certain optional metadata headers are missing, and by default we will fail if they are not provided (DTNSANAM, AIRMASS, DIMM2SEE, GAINA, GAINB). Since these are already considered "optional", here we only populate them in an ImageCollection/WorkUnit if present.

The `mock_butler` has been modified to allow for certain headers to be specified as "missing" from the metadata for testing purposes.

### DIFF
--- a/src/kbmod/standardizers/butler_standardizer.py
+++ b/src/kbmod/standardizers/butler_standardizer.py
@@ -377,14 +377,14 @@ class ButlerStandardizer(Standardizer):
             # is middleware so complicated! Best-effort attempt,
             # 90% cases?
             self._metadata["OBSID"] = meta["OBSID"]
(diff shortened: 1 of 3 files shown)


DinoBektesevic (approved):
This looks fine to me, optional keys should be optional. The one thing that I find scary about the optional(s) is what happens with set-like operations on collections - and I've encountered this in real life f.e. when stacking all extracted collections into a joined collection. 

In any case, this isn't the problem for this PR and is solvable just requires more logic than I feel not lazy to do. We should support optionals and this PR looks fine to me.

PR #782, opened 2025-01-23
Allow debug logging to be set by the config

jeremykubica:
Adds a path to turn on debug logging when the config's "debug" parameter is set to True.


wilsonbb (approved):
Nice!

PR #784, opened 2025-01-23
Fix encoding crash when an image is empty

jeremykubica:
The code was throwing a RuntimeError if any of the images in the ImageStack were empty (all NO_DATA). This might happen due to a combination of masking and reprojection. This PR fixes the code so it logs a warning that one of the images is empty, but otherwise proceeds.


wilsonbb (approved):
This seems a useful option, though hopefully we don't have to make use of it that often

PR #789, opened 2025-01-23
fix heliocentric_distance dropping from WorkUnit

maxwest-uw:


wilsonbb (approved):
I would add a small modification to the tests for this, but otherwise looks good to me

PR #790, opened 2025-01-24
Fix trajectory_predict_skypos mutating input array

wilsonbb:
`trajectory_predict_skypos` is meant to predict the positions of a KBMOD trajectory along a list or array of observation times, however it mutates the observation times if they are provided as a numpy array rather than a list.

It calls `dt = np.asarray(times)` which creates a new numpy array if `times` is a list or is a no-op if it is an array. We then subtract out the starting timestamp so all times are offset from 0, and we use the in-place operation of `dt -= dt[0]` to preserve memory. However if `times` is an array, the original object passed to the array has been mutated to now be offset by 0.

As quickly illustrated in a python repl:

```
>>> import numpy as np
>>> def modify_me(n):
...     new_n = np.asarray(n)
...     new_n -= new_n[0]
... 
>>> a = list(range(10, 20))
>>> modify_me(a)
>>> a
[10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
>>> b = np.asarray(a)
>>> modify_me(b)
>>> b
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]) 
```
This PR fixes this by choosing not to do the in-place subtraction.

To add better testing for this, we a) use tests with non-zero observation times to actually test the subtraction and b) test both lists and numpy arrays as inputs.

Note that this was showing up as an issue with the `KnownObjMatcher` having its numpy array of observation times mutated by the call to `trajectory_predict_skypos`. Then when it tries to compare the observation times against times from an ephemeris, a match cannot be made due to the temporal subtraction.

### DIFF
--- a/src/kbmod/trajectory_utils.py
+++ b/src/kbmod/trajectory_utils.py
@@ -124,7 +124,9 @@ def trajectory_predict_skypos(trj, wcs, times):
         A SkyCoord with the transformed locations.
     """
     dt = np.asarray(times)
-    dt -= dt[0]
+    # Note that we do a reassignment to avoid modifying the input, which may
+    # happen if `times` is already an array and `np.asarray` is a no-op.
+    dt = dt - dt[0]
 
     # Predict locations in pixel space.
     x_vals = trj.x + trj.vx * dt
--- a/src/kbmod/trajectory_utils.py
+++ b/src/kbmod/trajectory_utils.py
@@ -126,11 +126,11 @@ def trajectory_predict_skypos(trj, wcs, times):
     dt = np.asarray(times)
     # Note that we do a reassignment to avoid modifying the input, which may
     # happen if `times` is already an array and `np.asarray` is a no-op.
-    dt = dt - dt[0]
+    zeroed_dt = dt - dt[0]
 
     # Predict locations in pixel space.
-    x_vals = trj.x + trj.vx * dt
-    y_vals = trj.y + trj.vy * dt
+    x_vals = trj.x + trj.vx * zeroed_dt
+    y_vals = trj.y + trj.vy * zeroed_dt
 
     result = wcs.pixel_to_world(x_vals, y_vals)
     return result
--- a/tests/test_known_object_filters.py
+++ b/tests/test_known_object_filters.py
@@ -26,7 +26,10 @@ class TestKnownObjMatcher(unittest.TestCase):
 
         # Create a fake dataset with 15 x 10 images and 25 obstimes.
         num_images = 25
-        self.obstimes = np.array(create_fake_times(num_images))
+        start_time = 58922.1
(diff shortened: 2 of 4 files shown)


src/kbmod/trajectory_utils.py
```diff
@@ -124,7 +124,9 @@ def trajectory_predict_skypos(trj, wcs, times):
         A SkyCoord with the transformed locations.
     """
     dt = np.asarray(times)
-    dt -= dt[0]
+    # Note that we do a reassignment to avoid modifying the input, which may
+    # happen if `times` is already an array and `np.asarray` is a no-op.
+    dt = dt - dt[0]
```
jeremykubica, line 129: Nice fix. It might be slightly clearer if we also rename it to something like `zeroed_dt`, but I leave that up to you.

tests/test_known_object_filters.py
```diff
@@ -107,6 +110,9 @@ def setUp(self):
             time_offset=time_offset_mjd_close,
         )
 
+        # check that the obstimes have been unmodified and are still not zero-offset
+        self.assertTrue(np.all(self.obstimes >= start_time))
```
jeremykubica, line 114: To make sure they are completely unmodified, I think it is also worth adding `times_copy = np.copy(self.obstimes)` around line 30 and `self.assertTrue(np.all(self.obstimes == times_copy))` or something similar.

tests/test_trajectory_utils.py
```diff
@@ -62,12 +62,19 @@ def test_predict_skypos(self):
...
-        self.assertAlmostEqual(my_sky.dec[0].deg, -15.0)
-        self.assertAlmostEqual(my_sky.ra[1].deg, 45.2, delta=0.01)
-        self.assertAlmostEqual(my_sky.dec[1].deg, -15.5, delta=0.01)
+        # Predict locations at times 57921.0 and 57922.0, both as a list and as a numpy array.
+        obstimes = [57921.0, 57922.0]
+        for curr_obstimes in [obstimes, np.array(obstimes)]:
+            my_sky = trajectory_predict_skypos(trj, my_wcs, curr_obstimes)
+            # Verify that the obstimes were not mutated
+            self.assertEqual(curr_obstimes[0], 57921.0)
+            self.assertEqual(curr_obstimes[1], 57922.0)
```
jeremykubica, line 71: Nice test!

jeremykubica (approved):
Looks good. I added some optional comments (one stylistic and one extending the test).

tests/test_known_object_filters.py
wilsonbb: Ah good catch thanks!

src/kbmod/trajectory_utils.py
wilsonbb: Good, call

PR #799, opened 2025-02-07
Add CUDA_ARCHITECTURES flags

jeremykubica:
Build KBMOD for all GPU architectures. This will mean a larger binary, but should also mean that it works without needing the user to manually specify the GPU type.


drewoldag (approved):
Seems reasonable to me.

PR #801, opened 2025-02-07
Expand the documentation and comments

jeremykubica:
Update the documentation and comments to mostly add units. Also removes some unneeded documentation and an unneeded helper function for PSF.


maxwest-uw (approved):
nice

PR #802, opened 2025-02-08
Add tensorflow dependency

jeremykubica:
Daily build is failing because of a missing tensorflow depenency


maxwest-uw (approved):
`tensorflow` is kind of a big import, for the PR tests I just added `pip install -e ".[analysis]"` to include it, I think we could do that here but I'm also fine with just making it a required package as well

PR #824, opened 2025-03-06
Clean up scipy dependencies

jeremykubica:
Move a few of the scipy imports until where they are used so we don't import them if that path is not used.

Also remove koffi from pyproject.toml since we don't call it from within core KBMOD anymore.


drewoldag (approved):
Seems reasonable to me!

PR #832, opened 2025-03-11
WIP: Region Search Patch Generation

wilsonbb:


src/kbmod/region_search.py
```diff
@@ -1,424 +1,659 @@
...
+        containing the center coordinates, width, height, and image size as well as an ID corresponding
+        to its index in list `self.patches`.
 
-        # Now that we have the initial VDR data ids, we can also fetch the associated URIs
-        vdr_dict["uri"] = self.get_uris(vdr_dict["data_id"])
+        Parameters
+        ----------
+        arcminutes : float
+            The size of the patches in arcminutes.
+        overlap_percentage : float
```
drewoldag, line 278: Might want to clarify here that the float value should be between 0-100, and not 0-1. Simpletons like me are easily confused :)

src/kbmod/region_search.py
```diff
@@ -1,424 +1,659 @@
...
-
-    def _get_uris_serial(
-        self, data_ids, dataset_types=None, collections=None, butler=None, trim_uri_func=_trim_uri
-    ):
-        """Fetch URIs for a list of dataIds in serial fashion.
+        Patch
+            The Patch object with the given ID.
+        """
+        if not self.patches:
+            raise ValueError("No patches have been generated yet.")
```
drewoldag, line 356: Not sure if it's worth expanding the error message here.
```suggestion
            raise ValueError("No patches have been generated yet. Call `RegionSearch.generate_patches()`")
```

src/kbmod/region_search.py
```diff
@@ -1,424 +1,659 @@
...
-        return uris
-
-    def get_uris(self, data_ids, dataset_types=None, collections=None, trim_uri_func=_trim_uri):
-        """
-        Get the URIs for the given dataIds.
+        set of int
+            The indices of the patches that contain the ephemeris entries.
+        """
+        if guess_dist is not None and guess_dist != 0.0 and guess_dist not in self.guess_dists:
+            raise ValueError(f"Guess distance {guess_dist} not specified for RegionSearch")
```
drewoldag, line 378:
```suggestion
            raise ValueError(f"Guess distance {guess_dist} not specified for RegionSearch. Searchable guess distances are: {self.guess_dists}")
```

src/kbmod/region_search.py
```diff
@@ -1,424 +1,659 @@
...
+            [patch.ra for patch in self.patches],
+            [patch.dec for patch in self.patches],
+            unit=(u.deg, u.deg),
+            frame="icrs",
+        )
+
+        # Use search_around_sky to find the indices of the patches that may contain the ephemeris entries
+        # As our search radius from the patch center, we use the distance between the center of the patch
+        # and the corners of the patch, which can simply be calculated using the pythagorean theorem.
+        patch_size_deg = np.sqrt((self.patches[0].width / 2) ** 2 + (self.patches[0].height / 2) ** 2) * u.deg
```
drewoldag, line 408: Should this be an attribute of the Patch class so that you don't have to calculate it here?

src/kbmod/region_search.py
```diff
@@ -1,424 +1,659 @@
...
+
+        # Now we need to check if the ephemeris entry is actually in the patch
+        res_patch_indices = set([])
+        for ephem_idx, patch_idx in zip(ephems_idx, patch_idx):
+            if patch_idx not in res_patch_indices:
+                # Check if the ephemeris entry is in the patch
+                curr_ra, curr_dec = ephems_coords[ephem_idx].ra.deg, ephems_coords[ephem_idx].dec.deg
+                if self.patches[patch_idx].contains(curr_ra, curr_dec):
+                    res_patch_indices.add(patch_idx)
+        return res_patch_indices
```
drewoldag, line 422: Just to make sure I'm understanding correctly - the user passes in an ephemeris with many ra/dec values, and this function identifies all the patch ids that contain at least one of the ra/dec values from the ephemeris.

And if the user passes in an ephemeris with ra/decs that evenly cover the sky, we would likely return all the patch ids?

If that understanding is correct, is the expected use case that the user will generally pass in only a single ra/dec for a point on the sky to search around?

src/kbmod/region_search.py
```diff
@@ -1,424 +1,659 @@
...
+            new_ic.data["global_wcs_pixel_shape_0"] = patch_wcs.pixel_shape[0]
+            new_ic.data["global_wcs_pixel_shape_1"] = patch_wcs.pixel_shape[1]
+
+        # Reset standardizer-related metadata in our ImageCollection.
+        new_ic.meta["n_stds"] = len(new_ic)
+        new_ic.data["std_idx"] = range(len(new_ic))
+
+        return new_ic
+
+    def get_image_collection_from_patch(self, patch, guess_dist=0.0, min_overlap=0, max_images=None):
```
drewoldag, line 472: Just for readability, perhaps this method should go before `export_image_collection`? I'm assuming that the user would call this before they would call `export_image_collection`.

src/kbmod/region_search.py
```diff
@@ -1,424 +1,659 @@
...
+        patch : Patch or int
+            The patch to filter by.
+        guess_dist : float, optional
+            The guess distance to use for reflex correction. If 0.0, the original coordinates are used.
+        min_overlap : float, optional
+            The minimum overlap area in square degrees to include an image in the filtered ImageCollection.
+        max_images : int, optional
+            The maximum number of images to return. If None, all images are returned.
+        """
+        if guess_dist is not None and guess_dist != 0.0 and guess_dist not in self.guess_dists:
```
drewoldag, line 492: This is used a few times, would it be worthwhile moving it to a private function? 
```
def _validate_guess_dist(self, guess_dist):
    # the logic
```

src/kbmod/region_search.py
```diff
@@ -1,424 +1,659 @@
...
+            raise ValueError(f"Guess distance {guess_dist} not specified for Region Search")
+
+        if not isinstance(patch, Patch):
+            if not (isinstance(patch, int) or isinstance(patch, np.integer)):
```
drewoldag, line 496: I would suggest using this same conditional logic on line 455, 456. I think you can even do this:
```
if not isinstance(patch, [int, np.integer]):
```

And on line 455, 456, `if not isinstnace(patch, [Patch, int]):`

src/kbmod/region_search.py
```diff
@@ -1,424 +1,659 @@
...
+        new_ic.data["overlap_deg"] = overlap_deg
+        new_ic.data = new_ic.data[overlap_deg > min_overlap]
+
+        if max_images is not None and len(new_ic.data) > max_images:
+            # Limit the number of images to the maximum number of images requested,
+            # prioritizing the images with the highest overlap by sorting
+            new_ic.data.sort(["overlap_deg"], reverse=True)
+            new_ic.data = new_ic.data[:max_images]
+
+        return self.export_image_collection(ic_to_export=new_ic, guess_dist=guess_dist, patch=patch)
```
drewoldag, line 518: Ah, is `self.export_image_collection` a private method?

drewoldag (approved):
There are a buncha comments, but I don't think that there are any that should be considered blocking. Just questions for my own understanding and suggestions (that you're welcome to ignore). 

I would advocate for merging and moving ahead :)

src/kbmod/image_collection.py
```diff
@@ -461,6 +477,128 @@ def wcs(self):
...
+            The guess distance in parsecs.
+
+        Returns
+        -------
+        col_name : `str`
+            The name of the reflex-corrected column.
+        """
+        if col_name not in self.data.columns:
+            raise ValueError(f"Column {col_name} not in ImageCollection")
+        if not isinstance(guess_dist, float):
```
ColinOrionChandler, line 545: can we just convert ints, float64s, etc. to floats?

src/kbmod/image_collection.py
```diff
@@ -823,7 +998,7 @@ def vstack(self, ics):
 
         for meta, offset, ic in zip(old_metas, old_offsets, ics):
             ic.data["std_idx"] -= offset
-            ic.meta = meta
+            ic._meta = meta
```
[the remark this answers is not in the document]
ColinOrionChandler: Is this the re-indexing fix?

ColinOrionChandler (approved):
Made a couple of tiny comments, otherwise looks good.

PR #837, opened 2025-03-14
Add information and tests for compressing WorkUnit

### LINKED ISSUE #836: Look into compression of WorkUnits, etc.
46 Tb became 27 Tb with gzip

jeremykubica:
Closes #836

The WorkUnit functions already support compression given file names with the correct suffix. This PR adds documentation and tests.

### DIFF
--- a/tests/test_work_unit.py
+++ b/tests/test_work_unit.py
@@ -229,25 +229,13 @@ class test_work_unit(unittest.TestCase):
             self.assertIsNone(work2.wcs)
             for i in range(self.num_images):
                 li = work2.im_stack.get_single_image(i)
-                self.assertEqual(li.get_width(), self.width)
-                self.assertEqual(li.get_height(), self.height)
                 self.assertEqual(li.get_obstime(), 59000.0 + (2 * i + 1))
 
                 # Check the three image layers match.
-                sci1 = li.get_science()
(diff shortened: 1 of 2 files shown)


tests/test_work_unit.py
```diff
@@ -343,6 +373,41 @@ def test_save_and_load_fits_shard(self):
...
+            file_path = os.path.join(dir_name, "test_workunit.bz2")
+
+            # Write out the existing WorkUnit with a different per-image wcs for all the entries.
+            # work = WorkUnit(self.im_stack, self.config, None, self.diff_wcs)
+            work = WorkUnit(im_stack=self.im_stack, config=self.config, wcs=None, per_image_wcs=self.diff_wcs)
+            work.to_sharded_fits("test_workunit.bz2", dir_name)
+            self.assertTrue(Path(file_path).is_file())
+
+            # Read in the file and check that the values agree.
+            work2 = WorkUnit.from_sharded_fits(filename="test_workunit.bz2", directory=dir_name)
```
maxwest-uw, line 387: for some reason my intuition would be that this wouldn't work with a sharded work unit, nice to know that it does !

maxwest-uw (approved):
thanks for this! would be interesting to benchmark the i/o of reading from compressed work units, maybe we can do that next week

PR #845, opened 2025-03-20
Add another missing dependency

jeremykubica:
Another PR to fix the failing canary test. I believe this should be installed transitively (via astropy), but it is not being installed for the canary.


drewoldag (commented):
Looks good. How are these dependencies being added without updating the pyptoject.toml?

drewoldag (approved):

drewoldag:
I made the comment and in my head I approved this yesterday. Sorry for the delay!

PR #848, opened 2025-03-20
Add some utilities for analyzing fakes

jeremykubica:
Adds some helper functions that I ended up implementing as I was exploring one of the DEEP data sets.


src/kbmod/analysis/analyze_fakes.py
```diff
@@ -0,0 +1,225 @@
...
+
+        # Determine in which images the fake appears.
+        self.image_inds = get_matched_obstimes(obstimes, self.times, threshold=0.002)
+        if np.any(self.image_inds == -1):
+            raise ValueError(f"Unable to match one or more of the times:\n{obstimes}\n{self.times}")
+
+        # Compute the (x, y) pixel locations of the fakes in each image.
+        x_pos, y_pos = wu.get_pixel_coordinates(self.ra, self.dec, self.times)
+        self.x_pos_fakes = x_pos
+        self.y_pos_fakes = y_pos
```
wilsonbb, line 103: This all seems a generally fine path, but I believe alternatively you could just treat the fake info as an 'ephemeris' and simply call `make_stamps_from_emphems` instead. You tucked that function away in `precovery_utils` so easy to overlook

https://github.com/dirac-institute/kbmod/blob/6be0aad45e5333140bbe73dbacad87edb6dd1256/src/kbmod/analysis/precovery_utils.py#L10

src/kbmod/analysis/analyze_fakes.py
```diff
@@ -0,0 +1,225 @@
...
+            raise ValueError("compute_fit_mse can only be called after join_with_workunit.")
+
+        return evaluate_trajectory_mse(
+            self.trj,
+            self.x_pos_fakes,
+            self.y_pos_fakes,
+            self.times - self.times[0],
+        )
+
+    def compare_stamps(self, inds=None):
```
wilsonbb, line 150: This is nifty!

wilsonbb (approved):
Thanks! Made a note that there is a pre-existing function that may simplify some of this, but might be good reasons to not use that.

src/kbmod/analysis/analyze_fakes.py
jeremykubica: The real answer is that I forgot about that function.

But looking it over, I think I want to keep them separate for now simply because we need to extract the (x, y) pixel positions as well. So it is easier to do that together here instead of extracting (x, y) then re-extracting them for the stamps. I'm not sure this is a good reason though.

PR #854, opened 2025-03-24
Remove unused analysis files

jeremykubica:
Remove analysis files that are unused and have been replaced by other methods.


drewoldag (approved):
Seems reasonable enough to me!

PR #855, opened 2025-03-24
Add times attributes to Results

jeremykubica:
Add an optional times attribute to the `Results` object. This makes analysis easier by being able to refer to `results.times`. It can add some overhead in cases where the results object was already storing the times in the "mjd_mid" meta data column. We need to preserve that for compatibility with Dino's analysis scripts. But using the "time" meta data column will be clearer in the future.

### DIFF
--- a/src/kbmod/results.py
+++ b/src/kbmod/results.py
@@ -32,6 +32,9 @@ class Results:
     wcs : `astropy.wcs.WCS`
         A global WCS for all the results. This is optional and primarily used when saving
         the results to a file so as to preserve the WCS for future analysis.
+    times : `np.ndarray`
+        An array of the times for each observation. This is optional and primarily used
+        when saving the results to a file so as to preserve the times for future analysis.
     track_filtered : `bool`
         Whether to track (save) the filtered trajectories. This will use
         more memory and is recommended only for analysis.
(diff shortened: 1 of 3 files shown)


src/kbmod/run_search.py
```diff
@@ -297,6 +297,10 @@ def run_search(
         meta_to_save["dims"] = stack.get_width(), stack.get_height()
         meta_to_save["mjd_mid"] = [stack.get_obstime(i) for i in range(num_img)]
 
+        # We also save the times to the attribute. This technical writes them twice
```
wilsonbb, line 300: Nit: s/technical/technically

src/kbmod/results.py
```diff
@@ -32,6 +32,9 @@ class Results:
     wcs : `astropy.wcs.WCS`
         A global WCS for all the results. This is optional and primarily used when saving
         the results to a file so as to preserve the WCS for future analysis.
+    times : `np.ndarray`
+        An array of the times for each observation. This is optional and primarily used
+        when saving the results to a file so as to preserve the times for future analysis.
```
wilsonbb, line 37: Can we specify in the comment that these are also MJD_MID

jeremykubica: I think your overall point about it being less confusing to call this mjd_mid, so I renamed it. Can you take another look?

src/kbmod/results.py
```diff
@@ -68,9 +71,12 @@ def __init__(self, data=None, track_filtered=False, wcs=None):
             Whether to track (save) the filtered trajectories. This will use
             more memory and is recommended only for analysis.
         wcs : `astropy.wcs.WCS`, optional
-            A gloabl WCS for the results.
+            A global WCS for the results.
+        times : `np.ndarray`, optional
```
wilsonbb, line 75: s/times/mjd_mid in the doc string

wilsonbb (approved):
LGTM, just one doc string fix

PR #856, opened 2025-03-24
Add the velocity scatter plots.

jeremykubica:
Add scatter plots of the search space (black dots), the fakes (red dots), and KBMOD's results (blue dots).


drewoldag (approved):
nice

PR #858, opened 2025-03-24
fix extra memory usage in ImageCollection

maxwest-uw:
uses the `force_move` option for `ImageStack.append_image` to get rid of memory duplication.

## Before

icmem1

## After

icmem2


jeremykubica (approved):
Nice fix!

PR #860, opened 2025-03-25
Reduce JAX memory usage

jeremykubica:
By default JAX pre-allocated a huge amount of the GPU memory (75% of the memory) regardless of how much it needs. This PR turns that off and only allocates what is needed. The result is slower, but uses a tiny fraction of the GPU memory.


jeremykubica:
Yes. This should help a lot.

PR #862, opened 2025-03-26
Add a flag to turn of sigma G filtering

### LINKED ISSUE #861: Add flag to disable sigma-g (GPU, CPU)
Testing speed-up by disabling sigma-g, and diagnosing filtering problems.

jeremykubica:
Closes #861 

The bulk of this PR is adding a new test for for `load_and_filter_results()` which was previously untested. In the process, I found a small bug with `Results._update_likelihood()` where it was producing a 2-d array instead of a 1-d array.  This fixes that and adds a test.


drewoldag (approved):
Looks reasonable to me

PR #863, opened 2025-03-26
Switch from JAX to PyTorch

jeremykubica:
Switch SigmaG filtering from JAX to PyTorch. The main advantage is better usability and control of PyTorch. An additional advantage is the PyTorch implementation is faster (1.7ms for 10,000 rows instead of 2.6ms on Baldur).

### DIFF
--- a/src/kbmod/filters/sigma_g_filter.py
+++ b/src/kbmod/filters/sigma_g_filter.py
@@ -1,16 +1,15 @@
 """Functions to help with the SigmaG clipping.
 
 For more details see:
-Sifting Through the Static: Moving Objectg Detection in Difference Images
+Sifting Through the Static: Moving Object Detection in Difference Images
 by Smotherman et. al. 2021
 """
 
-from functools import partial
-from jax import jit, vmap
-import jax.numpy as jnp
 import logging
 import numpy as np
 import os
+import torch
+
 from scipy.special import erfinv
 
 from kbmod.results import Results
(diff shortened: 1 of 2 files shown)


src/kbmod/filters/sigma_g_filter.py
```diff
@@ -183,16 +130,45 @@ def compute_clipped_sigma_g_matrix(self, lh):
...
             or has been filtered (False).
         """
-        # We need to
-        os.environ["XLA_PYTHON_CLIENT_PREALLOCATE"] = "false"
-        os.environ["XLA_PYTHON_CLIENT_ALLOCATOR"] = "platform"
+        # Use a GPU if one is available.
+        if torch.cuda.is_available():
+            device = torch.device("cuda")
+        else:
+            device = torch.device("cpu")
```
maxwest-uw, line 137: I'm curious if there's a difference in speed between using `jax` and `pytorch` on cpu

src/kbmod/filters/sigma_g_filter.py
```diff
@@ -183,16 +130,45 @@ def compute_clipped_sigma_g_matrix(self, lh):
...
-        jnp_lh = jnp.array(lh)
-        jpn_inds_valid = self.sigma_g_jax_fn(jnp_lh).block_until_ready()
+        # Check whether the values fall within the bounds.
+        index_valid = torch.isfinite(torch_lh) & (torch_lh < upper_bnd) & (torch_lh > lower_bnd)
 
-        # Make sure the inds_valid is on CPU as a numpy array.
-        inds_valid = np.array(jpn_inds_valid)
-        return inds_valid
+        # Return as a numpy array on the CPU.
+        return index_valid.cpu().numpy()
```
maxwest-uw, line 171: I assume that `pytorch` will handle this gracefully when running on a machine without a gpu in the first place?

maxwest-uw (approved):

src/kbmod/filters/sigma_g_filter.py
jeremykubica: Yes. If the tensor is already on the cpu, then it is a no-op.

src/kbmod/filters/sigma_g_filter.py
jeremykubica: Yes. On my laptop the pytorch version takes about 7ms and the JAX version takes about 100ms.

PR #877, opened 2025-04-06
Make UTC standard throughout KBMOD

### LINKED ISSUE #865: Layup needs UTC, KBMOD results should output both UTC and TAI or add the unit to the column name
From the Layup sync today, the decision was made for Layup to use UTC. The Butler is in TAI. We would ideally output both TAI and UTC, but we could just output one as long as the unit label is in the column name. (This has already caused problems even before Layup.)

jeremykubica:
Closes #865

Make UTC the standard throughout KBMOD. This changes the Butler standardizer (which returns times in TAI) to convert to UTC (requires a small change to the tests). It updates a bunch of comments throughout to help clarify. The FITS standardizers were already in UTC, so those just have comments added.

### DIFF
--- a/src/kbmod/standardizers/butler_standardizer.py
+++ b/src/kbmod/standardizers/butler_standardizer.py
@@ -304,11 +304,11 @@ class ButlerStandardizer(Standardizer):
         half_way = mjd_start + (expt / 2) * u.s + 0.5 * u.s
         self._metadata["exposureTime"] = expt
 
-        # Note the timescales for MJD
-        # Name mjd into mjd_mid - make it obvious it's mdidle of exposure
-        # and add time scale like mjd_mid_utc
-        self._metadata["mjd_start"] = mjd_start.mjd
-        self._metadata["mjd_mid"] = half_way.mjd
+        # Note the timescales for MJD. The Butler uses TAI, but we convert
+        # time stamps to UTC for consistency.
+        # Name mjd into mjd_mid - make it obvious it's middle of exposure.
+        self._metadata["mjd_start"] = mjd_start.utc.mjd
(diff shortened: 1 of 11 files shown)


src/kbmod/standardizers/butler_standardizer.py
```diff
@@ -304,11 +304,11 @@ def _fetch_meta(self):
         half_way = mjd_start + (expt / 2) * u.s + 0.5 * u.s
```
wilsonbb, line 304: Since we're making a big time change, it seems worth investigating if the 0.5 constant should always be added here, and if so, at least comment why we're doing it. It's non-obvious to me at least.

wilsonbb (approved):
LGTM, just added a question about the line where we calculate the midpoint of the exposure since if we want any changes there should probably make them now.

src/kbmod/standardizers/butler_standardizer.py
wilsonbb: Hmm @ColinOrionChandler's comment on this line

"Recording the MJD at the halfway point during the exposure is different than trying to figure it out after the fact. The 0.5s is probably to account for a 1s shutter lag (or something similar) that occurs after mjd_start. This is not only instrument-specific, but can be specific to the mode of the instrument, potentially. So that's why it's not a good idea :slightly_smiling_face:"

wilsonbb: But fixing that may be out of scope for this PR

jeremykubica: Yeah. Let's hold off on changing that until we can talk to Dino and figure out if there is a better approach in general.

PR #879, opened 2025-04-07
Replace C++ PSF Object with Eigen Arrays

jeremykubica:
This PR mainly replaces the C++ PSF object with an Eigen matrix (allowing direct setting with numpy arrays).  It creates a new python based PSF class, but we might not need to use that much in the code (keeping it for now so we keep functionality). 

The PR maintains the current C++ convolution (since it is still needed while working with the C++ `RawImage`), but also adds a Pytorch based convolution that we can use in the future.


src/kbmod/search/image_stack.cpp
```diff
@@ -92,11 +92,6 @@ void ImageStack::sort_by_time() {
               [](const LayeredImage& a, const LayeredImage& b) { return a.get_obstime() < b.get_obstime(); });
 }
 
-void ImageStack::convolve_psf() {
-    if (data_on_gpu) throw std::runtime_error("Cannot modify images while on GPU");
-    for (auto& i : images) i.convolve_psf();
-}
-
```
maxwest-uw, line 99: just out of curiousity, did you benchmark the pytorch cpu convolve vs our previous C++ impl?

src/kbmod/search/kernels/image_kernels.cu
```diff
@@ -66,33 +65,44 @@ __global__ void convolve_psf(int width, int height, float *source_img, float *re
...
     int psf_dim = 2 * psf_radius + 1;
+    int psf_size = psf_dim * psf_dim;
+    
+    // Compute the PSF sum.
+    float psf_sum = 0.0;
+    for (int r = 0; r < psf_dim; ++r) {
+        for (int c = 0; c < psf_dim; ++c) {
+            psf_sum += psf_kernel[r * psf_dim + c];
+        }
+    }
```
maxwest-uw, line 84: I think `eigen` has a builtin `.sum()` method for their arrays, but I also see the kernel is being passed in as a `float *`, so is that an issue with passing between C++ and cuda?

maxwest-uw (approved):
nothing blocking, thanks for this!!

src/kbmod/search/image_stack.cpp
jeremykubica: Yes. Here is what I got from my timings:
C++ GPU: 7.3 ms
C++ CPU: 1.55 s
PyTorch GPU: 5.8 ms
PyTorch CPU: 189 ms
SciPy's convolve2d: 1.43 s

jeremykubica: The PyTorch CPU is a lot slower than the PyTorch GPU, but MUCH faster than the C++ CPU.

src/kbmod/search/kernels/image_kernels.cu
jeremykubica: Yes. CUDA code and eigen don't play well together. So I need to pass as a raw array of floats and manually do the sum.

Issue #875, opened 2025-04-04
Add compression for items writen to disk


ColinOrionChandler:
At least two potential benefits from compressing items stored on disk:

- roughly 50% on-disk savings (e.g., reprojected WorkUnits, results ECSV files)
- less read time

For FITS image HDUs, we could consider lossless packing. Gzip may offer the best balance between speed, savings, and Python convenience, but other options may exist.

jeremykubica:
The code is all there to compress WorkUnits in KBMOD. We would just need to change the names in the workflow configs.

jeremykubica:
Completed with the combination of #888 and #889

PR #889, opened 2025-04-09
Test Compression of Results

jeremykubica:
This PR is the other half of addressing #875 

The Results data structure supports writes in ecsv, hdf5, and parquet. Using takes a fraction of the space (~30% in tests) of ecsv. We need to change the file suffixes in the workflow to enable this switch.


wilsonbb (approved):
Yay for parquet!

PR #902, opened 2025-04-17
Move C++ convolution out of RawImage

jeremykubica:
This is a very small step to making the code more modular and allowing us to remove parts of the C++ code (such as RawImage). The ultimate goal is to simplify the build system and streamline the interface.


maxwest-uw (approved):
lgtm!

PR #906, opened 2025-04-18
Allow direct access of LayeredImage's Image data

jeremykubica:
Add a helper function to access pixel matrices from the `LayeredImage` directly.


drewoldag (approved):
Looks reasonable to me.

PR #907, opened 2025-04-18
Enable the Python stamp functions using an ImageStack

jeremykubica:
This PR expands the interface of the Python stamp generation functions and adds helper functions so that we can extract stacks of stamps directly from the C++ `ImageStack` (the last part is temporary code).  The most notable changes to the stamp generation API are:
- Allow the functions to either take a list of indices to use or a Boolean mask (from previously just a Boolean mask)
- Allow the functions to take a list of numpy arrays in addition to a 3-d numpy array.

### DIFF
--- a/src/kbmod/core/stamp_utils.py
+++ b/src/kbmod/core/stamp_utils.py
@@ -3,42 +3,12 @@
 import numpy as np
 import torch
 
-from numba import jit
+from numba import jit, typed
 
 
 # -----------------------------------------------------------------------------
 # --- Data Extraction Functions -----------------------------------------------
 # -----------------------------------------------------------------------------
-
-
-def extract_single_stamp(data, time_idx, x_val, y_val, radius):
-    """Generate a single stamp as an numpy array from a given time index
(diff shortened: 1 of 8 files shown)


src/kbmod/core/stamp_utils.py
```diff
@@ -331,22 +317,58 @@ def _extract_stamp_stack(imgs, x_vals, y_vals, radius, mask=None):
...
+    # Fill in each unmasked time step.
+    current = 0
+    for idx in range(num_times):
+        if mask is None or mask[idx]:
+            stamp_stack[current] = extract_stamp(imgs[idx], x_vals[idx], y_vals[idx], radius)
+            current += 1
+    return stamp_stack
+
+
+# @jit(nopython=True)
```
maxwest-uw, line 336: should this be commented out?

jeremykubica: No. That was a mistake.  I've removed it.

maxwest-uw (approved):

PR #911, opened 2025-04-21
Add psi and phi generation in Python

jeremykubica:
This PR adds more code to operation on images in Python (specifically `generate_psi_phi_images`). In the process I noticed a bug with the previous python convolve code. So this PR also adds some tests to check that the C++ and Python operations are giving the same results for non-toy data.

Also removes the Python PSF class's `make_square()` function because it isn't going to be used.


drewoldag (approved):
This all looks pretty reasonable to me.

PR #913, opened 2025-04-21
Add another parity test

jeremykubica:
Test that if we build a `PsiPhiArray` through either the C++ `ImageStack` or the Python processed data, we get the exact same information to send to the GPU, including compression.


drewoldag (approved):
This looks fine. I don't necessarily understand what all this new test is checking compared to the test that has been renamed. I would have guessed that the renamed test was implicitly making all these comparisons.

PR #916, opened 2025-04-22
Allow for CPU trajectory evaluation

jeremykubica:
This PR makes two changes:
1) It allows a single trajectory to be evaluated without KBMOD being compiled for GPU. This duplicates code, but should allow post-run analysis on non-GPU systems.
2) It removes the unused `BatchSearchManager` to simplify the C++ search code.


wilsonbb (commented):
It looks good to me, I guess I'm a bit worried about force_kernel being default False.

It seems like a big enough change to make the user be explicit about the behavior change for now, but maybe I'm being overly cautious. What are your thoughts?

jeremykubica:
That's a fair concern. I changed it to not have a default.

wilsonbb (approved):
LGTM

PR #919, opened 2025-04-23
Clean up result loading to reduce memory usage

### LINKED ISSUE #917: Enable truly disabling sigma G filtering, including light curves
Per discussion with KBMOD group, we need the ability to turn off sigma G filtering to, e.g., save a lot of memory.

jeremykubica:
Closes #917 

This PR introduces a bunch of improvements to data loading to reduce the total memory usage:
- Perform an initial round of filtering on the number of observations and likelihood in the C++ `TrajectoryList`. The data is already filtered with these same parameters in the GPU search, so we know the trajectories that do not meet these thresholds will not be valid. So we filter them from memory as soon as possible.
- Remove the iterative loop for LH checks during loading (since these checks are already done in the C++ code).
- Add a configuration parameter `generate_psi_phi` that can be set to `False` to disable computing the psi and phi curves (saving both running time and memory).
- Retrieve a reference to all the result trajectories instead of pulling down batches (and making copies).
- Build the results table (`keep`) once instead of in bits and appending.

It also adds some clean ups:
- Set the search's minimum number of observations via a setter instead of a function parameter to be consistent with the other parameters.
- Remove the unused `max_lh` parameter. We always set this so high that it wasn't triggered. So it was adding


maxwest-uw (approved):
lgtm! when usdf is back up tomorrow I can run my memory profile task to see the improvements from prefiltering out trajectories :)

PR #923, opened 2025-04-23
Make some PsiPhiArray functions const

jeremykubica:
Make PsiPhiArray getter and read functions constant in preparation for future support of possible multi-threading in KBMOD.


drewoldag (approved):
Looks good to me

PR #935, opened 2025-04-25
create provenance directory during search runtime and write out search configuration

### LINKED ISSUE #867: Need to write out provenance information, including configuration, with each run
Separate from the WorkUnit, e.g., geocentric distances, guess distance, other provenance.

maxwest-uw:
resolves #867 

if we are outputting a result filename, like "demo_res.ecsv", we will have the following structure in the result directory:

```
demo_res.ecsv
demo_res_provenance/
    demo_res_config.yaml
    ... any other config files we want to write out along with parsl
```


wilsonbb (approved):
Thanks LGTM!

PR #940, opened 2025-04-28
Add near duplicate pre-filtering

jeremykubica:
Use a hash table-based approximate filtering to remove near duplicates before doing any computational intensive filtering, such as sigma-G or full clustering.

### DIFF
--- a/docs/source/user_manual/results_filtering.rst
+++ b/docs/source/user_manual/results_filtering.rst
@@ -1,9 +1,28 @@
 Results Filtering
 =================
 
-The output files contain the set of all trajectories discovered by KBMOD. Many of these trajectories are false positive detections, some area already known objects and, because of the way KBMOD performs the search, some are duplicates. In the following sections we describe the various steps that remove unwanted trajectories from the set of results. These steps are applied by KBMOD in the order listed below.
+The output files contain the set of all trajectories discovered by KBMOD. Many of these trajectories are false positive detections, some are already known objects and, because of the way KBMOD performs the search, some are duplicates. In the following sections we describe the various steps that remove unwanted trajectories from the set of results. These steps are applied by KBMOD in the order listed below.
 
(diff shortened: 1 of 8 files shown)


docs/source/user_manual/results_filtering.rst
```diff
@@ -1,11 +1,30 @@
...
+The first step after the core search is to filter trajectories by their likelihoods and number of observations.  The relevant parameters are:
+
+ * ``lh_level`` - The minimum likelihood for a candidate trajectory to be kept.
+ * ``num_obs`` - The minimum number of non-masked observations for a candidate trajectory to be kept.
+
+
+Near Duplicate Pre-filtering
+----------------------------
+
+After the trajectories with too few observations or too small a likelihood are removed, the code then removes any near-duplicate results. Near duplicate detection uses an approximate hash table-based approach (the same as described in the "Grid Filtering" section below).  The pixel space is broken up into a grid at both the first and last time and each trajectory's indices in those grids are computed. If multiple candidates share the same starting **and** ending grid box, only the one with the highest likelihood is kept.
```
maxwest-uw, line 21: Is there a ballpark for about how much this filter reduces the results? my understanding was that a lot of "repeat" trajectories tend to converge towards the middle of the trajectory space and not the ends, but that might be something better served by dbscan and it's better to have the low hanging fruit like this filtered out first

maxwest-uw (approved):
lgtm

docs/source/user_manual/results_filtering.rst
jeremykubica: It depends on the grid size. But I have been finding anywhere from 25% to 75% of the results filtered in the test runs.

Unfortunately our use of DBScan isn't doing any sort of centroid based clustering. It clusters the points and then picks the single one with the highest likelihood. So it might give something on the edge of the cluster.

PR #947, opened 2025-05-01
Speed up NNSweepFilter

jeremykubica:
Speed up the core NNSweepFilter algorithm with a combination of multi-threading the kd-tree lookup and vectorizing some of the following processing.  The code processes results in batches to avoid computing an n^2 match matrix (in the case of very dense results) and using a lot of memory. The progressive querying also means we can avoid querying points that we already know are not the best in the region.

On the simple benchmarks this reduces the running time to about < 1% of previous (17.3s down to 0.136s for 90k results).

### DIFF
--- a/src/kbmod/filters/clustering_filters.py
+++ b/src/kbmod/filters/clustering_filters.py
@@ -194,15 +194,19 @@ class NNSweepFilter:
     """Filter any points that have neighboring trajectory with
     a higher likleihood within the threshold.
 
-    Parameters
+    Attributes
     ----------
     thresh : `float`
         The filtering threshold to use (in pixels).
     times : list-like
         The times at which to evaluate the trajectories (in days).
+    batch_size : `int`
+        The size of batching to use for kd-tree lookups.  A batch size of 1
+        turns off multi-threading and runs everything in series.
+        Default: 1000
     """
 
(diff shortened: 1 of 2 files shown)


src/kbmod/filters/clustering_filters.py
```diff
@@ -270,27 +270,30 @@ def keep_indices(self, result_data):
...
         cart_data = self._build_clustering_data(result_data)
         kd_tree = KDTree(cart_data)
+        build_data_timer.stop()
 
         num_pts = len(result_data)
         lh_data = result_data["likelihood"]
 
         # For each point, search for all neighbors within the threshold and
         # only keep the point if it has the highest likelihood in that range.
+        all_matches = kd_tree.query_ball_point(cart_data, self.thresh, workers=-1)
```
wilsonbb, line 283: It would be nice to add a flag or config param for this in case the number of available reported processors is inaccurate or let us disable the multithreading.

wilsonbb (approved):

src/kbmod/filters/clustering_filters.py
jeremykubica: I added a batch size parameter and the code will use a single worker if the batch size is one.

PR #951, opened 2025-05-05
Add multiple checks for runs with insufficient images

jeremykubica:
Count the number of images with fewer than 90% of the pixels masked and: a) fail if there are no valid images, or b) set `num_obs` to be at most the number of valid images.

Also gracefully handle cases where no valid results are returned from the core search.


wilsonbb (approved):
This is a good change!

General comment: 
I was thinking last week that we should be checking something like this earlier in the pipeline.

For ComCam processing, taking almost all WorkUnits seems useful, but for full Rubin we likely want to use this function to decide which WorkUnits are actually worth reflex-correcting. Ideally if WorkUnit construction was efficient enough, we could probably incorporate checks like this in region search itself.

Filed https://github.com/dirac-institute/kbmod/issues/953 for this

PR #954, opened 2025-05-05
Remove C++ objects from Standardizers and ImageCollection

jeremykubica:
Remove the dependency of both the Standardizers and ImageCollection on the C++ objects.  Within the standardizers use `LayeredImagePy` instead of `LayeredImage` since all we need is a data class to hold the various pieces of information.

Have `ImageCollection` build an `ImageStackPy` instead of a C++ `ImageStack`. This requires a slight change to `WorkUnit` to convert between the two types (moving the image references), so that we do not have to change everything that uses the `WorkUnit` or its `ImageStack` yet. Later changes will continue to push the boundary of the C++ code back.

### DIFF
--- a/src/kbmod/core/image_stack_py.py
+++ b/src/kbmod/core/image_stack_py.py
@@ -88,26 +88,36 @@ class ImageStackPy:
         The length T array of zeroed times.
     """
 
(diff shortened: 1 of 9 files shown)


jeremykubica:
Closing as this depends on an outstanding PR

PR #957, opened 2025-05-05
Remove image_allclose

jeremykubica:
Since we moved RawImages to Eigen a while back, `image_allclose` is effectively `np.allclose`. This removes the unneeded helper function and just uses np.allclose in tests.


drewoldag (approved):
This looks fine to me.

Issue #921, opened 2025-04-23
Split GPU stage from subsequent CPU-only work


ColinOrionChandler:
While we are GPU-limited we should have an option to split the GPU-dependent work from subsequent CPU-only work. This would free up the GPU much earlier for additional GPU-based searches. There would be a disk cost, and I/O cost, but probably worth pursuing.

jeremykubica:
CPU stages, such as: sigma-G filtering and generating stamps, can no be performed in isolation.

PR #959, opened 2025-05-06
Add commandline script for filtering

jeremykubica:
Partially addresses #921

Creates command line scripts for filtering and producing basic statistics from a results file. This allows the filtering to be run after the core search on a CPU-only machine.


maxwest-uw (approved):
will we eventually have the ability to run all search components from the CLI as well? nice change : )

PR #965, opened 2025-05-12
Cleanup unused functions

jeremykubica:
This PR removes a bunch of unused imports, functions, and tests.  Specifically:
- Unused imports are removed in a bunch of places.
- All basic filters are removed. These were never used and never even hooked up to the configuration so they could be used. Most of these can easily be done with masks and the `Results` table.
- Remove `Trajectory.is_close` as it is used in a single test and can be replaced by a few lines of logic.
- Remove test_search_filter.py as this test is covered by multiple other tests.


drewoldag (approved):
Looks like a great clean up to me.

PR #971, opened 2025-05-13
Fix a bug with compressing NaNs in FITs

jeremykubica:
For some types of compression, we may see problems writing and then reading NaNs to fits. I don't think we have seen this in any of our runs, because we always apply the mask (and thus set pixels to nan) after reading the `WorkUnit` and before running the search. But we could run into problems in the future if we try to save a `WorkUnit` whose mask has already been applied.

This PR fixes the issue by: 
1) When writing a WorkUnit:
    a) moving all NaN pixels to bits in the mask layer
    b) replace the NaNs in the science and variance layers with 0.0
2) When writing the WorkUnit: apply masking immediately to the science and variance layers.

### DIFF
--- a/src/kbmod/work_unit.py
+++ b/src/kbmod/work_unit.py
@@ -1026,7 +1026,8 @@ def add_image_data_to_hdul(
     wcs=None,
 ):
     """Add the image data for a single time step to a fits file's HDUL as individual
-    layers for science, variance, etc.
+    layers for science, variance, etc.  Masked pixels in the science and variance
+    layers are added to the masked bits.
 
     Parameters
     ----------
@@ -1095,6 +1096,7 @@ def add_image_data_to_hdul(
 
 def read_image_data_from_hdul(hdul, idx):
     """Read the image data for a single time step to a fits file's HDUL.
+    The mask is auto-applied to the science and variance layers.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


src/kbmod/work_unit.py
```diff
@@ -1116,9 +1130,12 @@ def read_image_data_from_hdul(hdul, idx):
     # Get the variance layer.
     var = hdul[f"VAR_{idx}"].data.astype(np.single)
 
-    # Allow the mask to be optional. Use an empty mask if none is present.
+    # Allow the mask to be optional. Apply the mask if it is present
+    # and use an empty mask if there is no mask layer.
     if f"MSK_{idx}" in hdul:
         mask = hdul[f"MSK_{idx}"].data.astype(np.single)
+        sci[mask > 0] = np.nan
+        var[mask > 0] = np.nan
```
maxwest-uw, line 1138: is it ok that all of the masked pixels in science and variance are converted to `np.nan`? if there are pixels masked out for reasons other than their being NaNs we'd lose that data. doesn't matter for search but could cause problems for debugging/provenance

jeremykubica: I think it is okay, but would like you to check my logic:
- We apply the masking flags in the standardizer, so we only propagate those bits we want to capture.
- We are already binarizing the mask when saving it (the write function saves "mask > 0" instead of the raw integers.
- We always apply OxFFFFFFF for the mask in the code.

maxwest-uw (approved):

PR #972, opened 2025-05-14
Allow fuzzing/disordering of WorkUnit obstimes

wilsonbb:
Applies random fuzzing to the obstimes of the WorkUnit for the purposes of generating unreasonable KBMOD search results for the purposes of ML training (AKA known negative results).

This should work on a WorkUnit whether or not it has been reflex-corrected. If there are multiple images with the same obstime, they will be mapped to the same new obstime (preserving mosaicking when later reprojected). The ImageStack is also sorted afterwards to preserve the sorted WorkUnit ordering caused by reflex-correcting/resampling.

### DIFF
--- a/src/kbmod/work_unit.py
+++ b/src/kbmod/work_unit.py
@@ -430,7 +430,8 @@ class WorkUnit:
         # been sorted so we do this to preserve that expectation after reordering.
         self.im_stack.sort_by_time()
 
-        # Reset the cached obstimes to use what was sorted in the ImageStack.
+        # Clear metadata and reset the cached obstimes to use what was sorted in the ImageStack.
+        self.clear_metadata()
         self._obstimes = None
 
     @classmethod
@@ -1037,6 +1038,11 @@ class WorkUnit:
         config_filename = f"{base_filename}_config.yaml"
         config_path = provenance_dir_path.joinpath(config_filename)
         self.config.to_file(config_path, overwrite)
+
+    def clear_metadata(self):
+        """Clear all WorkUnit metadata."""
+        self.org_img_meta = Table()
+        self._per_image_indices = [[i] for i in range(self.n_constituents)]
 
 
 def load_layered_image_from_shard(file_path):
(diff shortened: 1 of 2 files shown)


src/kbmod/work_unit.py
```diff
@@ -380,6 +380,59 @@ def get_unique_obstimes_and_indices(self):
         unique_indices = [list(np.where(all_obstimes == time)[0]) for time in unique_obstimes]
         return unique_obstimes, unique_indices
 
+    def disorder_obstimes(self):
+        """Reorders the timestamps in the WorkUnit to be random. Random offsets
+        are chosen for each unique obstime and added to the original obstime.
+        The maximum offset is the number of time in the ImageStack or
```
jeremykubica, line 386: "time" -> "times" ?

jeremykubica (commented):
One high level question: How should we handle all of the meta-data? Things like `per_image_indices` and `org_img_meta` will no longer align with the image order, correct?  My thinking is that we should just clear out all metadata so we do not accidentally use it.  We could create a function to do this and then have it if we add anything else that modifies order.

ColinOrionChandler:
I'm mulling over the logic of doing this via the WorkUnit instead of ImageCollection. If we do this by WorkUnit, we'll have plenty that were generated from the best datasets that span small timespans, including single-night. I think this is fine, and maybe we have a large enough volume of images that this will suffice, but if we were able to shuffle the large ImageCollections before we split them up, we'd have a different level of variety. I'm fine to try this WorkUnit approach first, though, and just make a mental note that we have another option available.

src/kbmod/work_unit.py
wilsonbb: Done

wilsonbb:
> One high level question: How should we handle all of the meta-data? Things like `per_image_indices` and `org_img_meta` will no longer align with the image order, correct? My thinking is that we should just clear out all metadata so we do not accidentally use it. We could create a function to do this and then have it if we add anything else that modifies order.

This makes sense, and I added a `clear_metadata` function to cover this. Our metadata has gotten kind of complicated so there might be some additional cases that I should be clearing as well, but I think this covers it.

jeremykubica (approved):

PR #974, opened 2025-05-16
Small clean ups

jeremykubica:
A few really small cleanups:
- Move some of the old unused functions to a directory called "unused" to make it clear that we are not using them.
- Remove the unused `estimate_gpu_memory()` from `WorkUnit`. This was replaced by a more comprehensive memory estimation function in run search.


drewoldag (approved):
Looks good to me.

PR #976, opened 2025-05-19
Clean up ImageStackPy masking helpers

jeremykubica:
This PR has a bunch of small clean ups, including:
- Add a function to `ImageStackPy` that creates a mask image.
- Removing duplicate code between counting number and fraction of masked pixels.
- Change `load_deccam_layered_image` to produce a `LayeredImagePy`
- Clean up some comments that talk about C++ data structures to make them either talk about the python version or just generic "image stack".

### DIFF
--- a/src/kbmod/analysis/plotting.py
+++ b/src/kbmod/analysis/plotting.py
@@ -7,14 +7,13 @@ import astropy.units as u
 from astropy.wcs import WCS
 from astropy.nddata.utils import Cutout2D
 from astropy.utils.exceptions import AstropyWarning
-from astropy.visualization import simple_norm, ZScaleInterval, AsinhStretch, ImageNormalize
+from astropy.visualization import ZScaleInterval, AsinhStretch, ImageNormalize
 
 import matplotlib.pyplot as plt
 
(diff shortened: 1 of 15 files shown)


src/kbmod/analysis/plotting.py
```diff
@@ -394,7 +393,9 @@ def plot_image(img, ax=None, figure=None, norm=True, title=None, show_counts=Tru
         pass
 
     # Check the image's type and convert to an numpy array.
-    if type(img) is LayeredImage:
+    if isinstance(img, LayeredImage):
+        img = img.sci
+    elif isinstance(img, LayeredImagePy):
         img = img.sci
```
wilsonbb, line 399: Why do we need the if/elif for if we are doing `img = img.sci` in both cases?

wilsonbb (commented):
Looks good overall but confused about one part

src/kbmod/analysis/plotting.py
jeremykubica: We don't need them as separate. We do need the if in case someone passes a numpy array (in which case we don't want to try to access the sci attribute). I fixed so the check uses an or

wilsonbb (approved):

PR #977, opened 2025-05-19
Remove unused notebooks

jeremykubica:
Remove `Kbmod_Reference.ipynb` since it is no longer needed (and is very out of date) with the code re-organization.  Move two other unused notebooks into the reference folder.


drewoldag (approved):
approved

PR #979, opened 2025-05-19
Remove the C++ ImageStack and LayeredImage

jeremykubica:
This PR completes the transition of core data loading and storage to Python. The `ImageStackPy` replaces the C++ `ImageStack` in the `WorkUnit`.


maxwest-uw (approved):
wooo!!

PR #980, opened 2025-05-21
Reduce WorkUnit memory use

jeremykubica:
PR #971 increased the memory usage when a `WorkUnit` wrote out to a file by creating copies of all three layers in order to overwrite the NaNs. However we don't need to actually overwrite the NaNs when writing to a file. We just need to make sure to update the mask layer and overwrite the NaNs on load (since sometimes the compression on some machines was moving NaN->non NaN... I'm still not sure how that was happening, regardless we need to make sure the NaNs are in the correct place after load). The PR removes copies of the science and variance layers (instead writing them as they are).

A new copy of the map layer is created, so it will use more memory there. However this is stored as an 8-bit integer (because the fits writer does not like Boolean arrays), so the memory footprint is 1/4th that of the other layers.

Other changes:
- A mislabeled warning was fixed in image_utils.py
- Dynamic quantization is used to better trade off size and accuracy of compressed files. Instead of quantizing to a given number of steps (`quantize_level=500.0`) we now quantize to at least 0.01 between fluxes (by specifying a negative `quantize_level`).
- Larger test images are used in the unittests.
- An object is added to the images in the unittest to increase the range of science pixels.
- A "large" unittest added that uses larger images with an even bigger range to confirm the quantization settings are not breaking anything.

### DIFF
--- a/src/kbmod/image_utils.py
+++ b/src/kbmod/image_utils.py
@@ -324,13 +324,13 @@ def validate_image_stack(
             is_valid = False
         if np.nanmin(var) < min_var:
             _im_stack_validation_error(
-                f"Image {idx} has invalid flux values: {np.nanmin(var)} < {min_var}",
+                f"Image {idx} has invalid variance values: {np.nanmin(var)} < {min_var}",
                 warn_only,
             )
             is_valid = False
         if np.nanmax(var) > max_var:
             _im_stack_validation_error(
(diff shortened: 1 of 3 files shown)


wilsonbb (approved):
This LGTM! 

At first I was unsure if the per-image index copy could be the memory culprit, but after revisiting the reproject code I agree since this method is being called in parallel

PR #989, opened 2025-05-27
Rename kernel_testing_helpers

jeremykubica:
This is a quick prefactor before adding more code to the kernel_testing_helpers / kernel_helpers file that moves the file beyond testing helpers.


src/kbmod/search/kernel_helpers.h
delucchi-cmu, line 1: This file is freaky. 
- It was already defined as `#define KERNEL_HELPERS_H_`? 
- Should the updated file-level comment go in the `.h`? 
- Do any file-level comments get pulled in as module-level documentation in the API docs?

delucchi-cmu (approved):

PR #990, opened 2025-05-27
Fix a noisy test warning

jeremykubica:
Fix a warning during tests that the code has detected a "dubious year" by using more realistic MJD time stamps.


drewoldag (approved):
A much more modern date, very nice.

PR #995, opened 2025-05-28
Provenance tracking

jeremykubica:
Propagate provenance tracking from the initial data to the results file. We have a field "data_loc" in the `WorkUnit` meta data that is meant for provenance tracking, but is not being filled (except from the "ILOC" header of legacy work unit files, but we don't write out the "ILOC" header anymore).

Use the standardizer's "location" meta data for the `WorkUnit` "data_loc" field. Also propagate the "dataId" field from a Butler created `ImageCollection`.

### DIFF
--- a/src/kbmod/run_search.py
+++ b/src/kbmod/run_search.py
@@ -490,7 +490,7 @@ class SearchRunner:
 
         # Extract extra metadata. We do not use the full org_image_meta table from the WorkUnit
         # because this can be very large and varies with the source. Instead we only save a
-        # few pre-defined fields to the results data.  This these columns are not present in the
+        # few pre-defined fields to the results data.  If these columns are not present in the
         # WorkUnit, they are skipped in the meta data.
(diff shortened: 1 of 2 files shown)


wilsonbb (approved):
This looks good to me

I went and looked through the various WorkUnit construction calls in the reprojection module and I believe this should all propagate https://github.com/dirac-institute/kbmod/blob/f953e121f06f684295a08ff66c33681f9933135e/src/kbmod/reprojection.py#L127

However we have been burned by this in the past, and if we wanted to increase the test coverage we could
a) add a test the org_image_meta is preserved by copy() (what most of the pathways in the reprojection module seems to assume?)
b) add explicit tests that provenance is propagated for reproject_work_unit. We do test this for one of the pathways and "geocentric_distance" http://github.com/dirac-institute/kbmod/blob/f953e121f06f684295a08ff66c33681f9933135e/tests/test_reprojection.py#L82

But I think this is a nice-to-have and doesn't block merging

PR #1000, opened 2025-05-29
Remove unused function in Results

jeremykubica:
Remove the unused `mask_based_on_invalid_obs` function from `Results`.  This was implemented for an old filtering approach that did not work out.

Also clean up an extra copy in `PSF.copy()`.


drewoldag:
I get to review the 1000th PR?!?!?! 🎉

PR #1007, opened 2025-06-06
Switch CNN Filter to `pytorch` and remove `tensorflow`

maxwest-uw:
- removes the `tensorflow` dependency and the `keras` testing model
- uses `pytorch` to carry out inference (also tested on actually trained kbmod-ml weights)
- uses the default `torchvision.resnet18` to run the unit tests with a manually set seed to get consistent results, which reduces the amount of test data the `kbmod` repo needs to carry around


src/kbmod/filters/stamp_filters.py
```diff
@@ -181,7 +230,9 @@ def _normalize_stamps(stamps, stamp_dimm):
     return np.array(normed_stamps)
 
 
-def filter_stamps_by_cnn(result_data, model_path, coadd_type="mean", stamp_radius=10, verbose=False):
+def filter_stamps_by_cnn(
+    result_data, model_path, model_type="resnet18", coadd_type="mean", stamp_radius=10, verbose=False
+):
```
jeremykubica, line 235: Can you add `model_type` to the docstring?

src/kbmod/filters/stamp_filters.py
```diff
@@ -201,26 +252,37 @@ def filter_stamps_by_cnn(result_data, model_path, coadd_type="mean", stamp_radiu
...
+        weights=model_path,
+        shape=stamp_shape,
+    )
+    # we'll need to check if we have a GPU available.
+    if model_path:
+        if torch.cuda.is_available():
+            cnn.load_state_dict(torch.load(model_path))
+        else:
+            cnn.load_state_dict(torch.load(model_path, map_location=torch.device("cpu")))
+    cnn.eval()
```
jeremykubica, line 278: Can you add a comment here that this just sets the model to evaluation model? I was originally confused, because I thought it was doing the computation here.

jeremykubica (approved):

PR #1009, opened 2025-06-10
Commandline Tool for Ingesting Butler Collections as KBMOD ImageCollections

wilsonbb:
Provides a simple commandline tool for ingesting Butler collections as ImageCollections.

The user must provide a butler repository path, and a butler datasetType, such as `preliminary_visit_image` or `difference_image`, to ingest. Then the user can either provide a list of collections to process or a regex across the repository's collection names.

Then for each specified collection, a KBMOD ImageCollection is created and serialized to a specified output directory.


jeremykubica (approved):
Looks good. Thanks!

PR #1038, opened 2025-07-14
return values for peak_offset_filter updated

SamSandwich07:
- updated return values from peak_offset_filter to only return results
- updated run_search.py to update results when peak_offset_filter is called
- updated unit tests for peak_offset_filter to reflect new return value


SamSandwich07:
I found another issue that has to do with weird Results vs. AstroPy table return values so please don't merge yet

PR #1044, opened 2025-07-30
separate `image_positions_to_original_icrs` from `WorkUnit`

### LINKED ISSUE #1041: move `image_positions_to_original_icrs` to seperate function
we no longer persist `WorkUnit`s as long as when we first started this WCS and reprojection work so we should move it out of the `WorkUnit` (but keep it maintained within the normal `run_search` pipeline)

maxwest-uw:
resolves #1041

### DIFF
--- a/src/kbmod/work_unit.py
+++ b/src/kbmod/work_unit.py
@@ -18,7 +18,7 @@ from tqdm import tqdm
 from kbmod import is_interactive
 from kbmod.configuration import SearchConfiguration
 from kbmod.core.image_stack_py import ImageStackPy, LayeredImagePy
-from kbmod.reprojection_utils import invert_correct_parallax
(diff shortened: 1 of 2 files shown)


src/kbmod/work_unit.py
```diff
@@ -878,89 +878,24 @@ def image_positions_to_original_icrs(
...
-                        f"ambiguous image origin for coordinate {i}, including all potential constituent images.",
-                        Warning,
-                    )
-            else:
-                positions.append(pos[0])
-        return positions
+
+        original_wcses = [w for w in self.org_img_meta["per_image_wcs"]]
+
+        return image_positions_to_original_icrs(
```
wilsonbb, line 884: It seems better to remove this wrapper and just have run_search.py call image_positions_to_original_icrs from reprojection_utils, but am fine with this more conservative approach if you prefer

PR #1055, opened 2025-08-28
Fix a memory leak

jeremykubica:
We should be clearing the data when it is on GPU, not when it isn't on GPU.


DinoBektesevic:
Just checked it on Klone, `nvidia-smi` says we jump up to 5.5GB then drop down to 500MB, then jump up to 5.5GB, as expected for this job I believe. I think the jumps in memory usage are the search loop is loading up the psi-phi arrays for the new search and then unloading them for each result, since expected footprint of the results is ~120MB or so. So this looks good to me is the conclusion.

DinoBektesevic (approved):

PR #1056, opened 2025-08-29
Add commandline tool for matching KBMOD result(s) files to ephemeris

wilsonbb:
A simple command line tool for taking ephemeris files and performing results matching against one or more results files.

This was used to generate a "backfill" of matching against known objects for a group of results files on USDF, and can also be used to match against an updated ephemeris (newly submitted objects, injected fakes, etc)

### DIFF
--- a/src/kbmod_cmdline/kbmod_result_matcher.py
+++ b/src/kbmod_cmdline/kbmod_result_matcher.py
@@ -1,0 +1,445 @@
+"""A program to match one or more KBMOD results files against an ephemeris of known objects.
+
+This script is designed to be run from the commandline and will process results files, matching them against known objects in a SkyTable database.
+It will output a tbale of which results mathched with which known objects, parameters used, and any exceptions encountered during processing.
+
+Usage:
+    python kbmod_result_matcher.py --results_glob <results_file> --ephem <ephemeris_file> [--output <output_dir>]
+
+Example:
+    python kbmod_result_matcher.py --results /output/resuls*.parquet --ephem skytable.parquet --output .
+
+This will produce a CSV file, `matching_results.csv` in the output directory with the matched results, and an `exceptions.csv` file with any files that could not be processed.
+"""
+
+import argparse
+import glob
+import os
+import pandas as pd
+
+from astropy.table import Table
+
+import csv
+
+csv.field_size_limit(131072 * 2)  # Increase field size limit for reading large CSV files.
+
+from kbmod import ImageCollection
+from kbmod.filters.known_object_filters import KnownObjsMatcher
+from kbmod.results import Results
+
+
+def reflex_correct_ephem_table(ephem_table, barycentric_dist, obs_site="Rubin"):
+    """Apply reflex correction to the ephemeris table if barycentric distance is provided.
+
+    Assumes that the observatory is Rubin Observatory.
+
+    Produces columns of the form 'ra_<barycentric_dist>' and 'dec_<barycentric_dist>' in the returned table.
+
+    Parameters
+    ----------
+    ephem_table : astropy.table.Table
+        The ephemeris table containing known objects, assumes 'RA', 'Dec', and 'mjd_mid' columns.
+    barycentric_dist : float
+        The barycentric distance in AU. If 0.0, no correction is applied.
+    obs_site : str
+        The observatory site to use for reflex correction. Default is "Rubin".
+
+    Returns
+    -------
+    astropy.table.Table
+        The corrected ephemeris table.
+    """
+
+    if barycentric_dist != 0.0:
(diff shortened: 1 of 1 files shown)


src/kbmod_cmdline/kbmod_result_matcher.py
```diff
@@ -0,0 +1,445 @@
...
+        corrected_skycoord, _ = correct_parallax_geometrically_vectorized(
+            ephem_table["RA"],
+            ephem_table["Dec"],
+            ephem_table["mjd_mid"],
+            barycentric_distance=barycentric_dist,
+            point_on_earth=EarthLocation.of_site(obs_site),
+        )
+        ephem_table[f"ra_{barycentric_dist}"] = corrected_skycoord.ra.deg
+        ephem_table[f"dec_{barycentric_dist}"] = corrected_skycoord.dec.deg
+    if barycentric_dist == 0.0:
```
jeremykubica, line 67: This could just be an else. Practically it doesn't matter though.

src/kbmod_cmdline/kbmod_result_matcher.py
```diff
@@ -0,0 +1,445 @@
...
+    This works for results files outputted from our KBMOD parsl workflow because
+    1. We copy the ImageCollection to the same directory as the results file.
+    2. The workflow outputs files with appended file extensions from the paths of previous steps.
+       e.g. `patch_33.collection.[other_steps].results.parqute` for a results file and `patch_33.collection` for the ImageCollection.
+
+    Parameters
+    ----------
+    res_filepath : str
+        The path to the results file.
+    Returns
```
jeremykubica, line 86: Add a blank line above "Returns"

src/kbmod_cmdline/kbmod_result_matcher.py
```diff
@@ -0,0 +1,445 @@
...
+        res = Results.read_table(results_file)
+    except Exception as e:
+        print(f"Error reading in results file {results_file}: {e}")
+    if verbose:
+        print(f"Processing {len(res)} results from file: {results_file}")
+    if max_results is not None:
+        if max_results <= 0:
+            raise ValueError("max_results must be a positive integer.")
+        if max_results < len(res):
+            print(f"Limiting processing to the first {max_results} results.")
```
jeremykubica, line 148: Should this also be an "if verbose"

src/kbmod_cmdline/kbmod_result_matcher.py
```diff
@@ -0,0 +1,445 @@
...
+                num_matched_obs = sum(obs_mask)
+                matched_obs.append(num_matched_obs)
+                obs_ratios.append(num_matched_obs / len(obs_mask))
+    matching_df = pd.DataFrame(
+        {
+            "results_file": [results_file] * len(uuids),
+            "barycentric_dist": [barycentric_dist] * len(uuids),
+            "sep_thresh": [sep_thresh] * len(uuids),
+            "time_thresh_s": [time_thresh_s] * len(uuids),
+            "min_obs": [min_obs] * len(uuids),
```
jeremykubica, line 200: Since a bunch of this is the same for every line, do we need it in table format? Pandas allows you to set table-wise meta data that we could use instead. Of course it really depends on how you are going to use it.

Actually, I see below you are outputting it as a CSV. So this makes sense.

src/kbmod_cmdline/kbmod_result_matcher.py
```diff
@@ -0,0 +1,445 @@
...
+                results_processed.to_csv(matched_results_file, mode="a", header=True, index=False)
+                first_write = False
+            else:
+                # The outputted file file is a serialized CSV file of a pandas DataFrame.
+                # We output in append mode if it already exists.
+                results_processed.to_csv(matched_results_file, mode="a", header=False, index=False)
+        except Exception as e:
+            print(f"Exception occurred: {e}")
+            exceptions["result_file"].append(results_file)
+            exceptions["error"].append(str(e))
```
jeremykubica, line 335: I like how you are capturing the exceptions here.

jeremykubica (approved):
Looks good. A bunch of optional recommendations.

src/kbmod_cmdline/kbmod_result_matcher.py
wilsonbb: Lol thanks for pointing this out. Probably left over from a previous iteration of logic.

src/kbmod_cmdline/kbmod_result_matcher.py
wilsonbb: Done.

src/kbmod_cmdline/kbmod_result_matcher.py
wilsonbb: Thanks!

PR #1058, opened 2025-09-15
Validate Image Stack Mask Shape Based on Current Mask Shape

wilsonbb:
We allow a KBMOD image stack to have images of different shapes, and when we add a new image to the stack we want to compare the dimensions of that mask to its corresponding science image.

### DIFF
--- a/tests/test_image_stack_py.py
+++ b/tests/test_image_stack_py.py
@@ -1,7 +1,6 @@
 import logging
 import numpy as np
 import unittest
-import pytest
 
 from kbmod.core.image_stack_py import ImageStackPy, LayeredImagePy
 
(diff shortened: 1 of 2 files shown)


jeremykubica (approved):

tests/test_image_stack_py.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import numpy as np
 import unittest
+import pytest
```
jeremykubica, line 4: Pytest isn't included as part of KBMOD. You can use: 

```
self.assertRaises(ValueError, function_name, params)
```

instead. See test_clustering_filters.py for an example

PR #1083, opened 2025-10-22
Issue884

jeremykubica:
Add the ability to scale the images by a magnitude value that depends on their filter. The magnitude value for each filter used is passed in as a dictionary (filter_name -> magnitude scale). The scaling is applied to the image stack before the run, so multiple scaling can be tried using a single workunit on disk.


ColinOrionChandler:
Looks great! Slacked Jeremy earlier.

PR #1086, opened 2025-10-22
Add masking and WCS approximation configuration to kbmod_build_ic commandline tool

### LINKED ISSUE #1061: Mask handling configurability needed
Growth should be optional
Amount of growth should be adjustable
Flags should be choosable

wilsonbb:
Allow for updating various butler standardizer config parameters from the kbmod_build_ic commandline utility.

Includes support for configuring:
* WCS approximation parameters (number of random points sampled and SIPS term)
* specifying different mask flags
* disabling growing mask
* Changing the kernel growth shape
* whether building an ImageCollection will fail if there are any images that can't be standardized.

This allows us to preserve the default standardizer flags that have been used by DEEP while updating for Rubin.

Fixes #1061

### DIFF
--- a/src/kbmod/filters/clustering_grid.py
+++ b/src/kbmod/filters/clustering_grid.py
@@ -39,12 +39,12 @@ class TrajectoryClusterGrid:
         bin_width : `int`, optional
             The width of each of the spatial bins (in pixels).
         """
-        if bin_width < 1:
-            raise ValueError("Bin width must be at least 1.")
+        if bin_width < 1 or not np.isfinite(bin_width):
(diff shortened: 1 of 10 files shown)


src/kbmod_cmdline/kbmod_build_ic.py
jeremykubica: Add a doc comment for `butler_standardizer_config` argument.

wilsonbb: Done.

jeremykubica (approved):

PR #1119, opened 2026-01-22
Fix a bunch of linting errors

jeremykubica:
The new version of black makes some small changes. I reran black to pack them up so the tests would stop failing.


maxwest-uw (approved):
thanks haha I was confused last night why the linter was failing

PR #1124, opened 2026-01-28
Support Chunked Reading of Large Result Files and Improve Image Column Support

wilsonbb:
A bump for astropy >= 5.3  ([released May 2023](https://pypi.org/project/astropy/5.3/)) and a few intertwined issues aimed to help us deal with large result files on USDF:

* For large KBMOD results files (say 20GB or more) we seem to run into segfaults or memory constraints. We introduce `Results.read_table_chunks` to support yielding these chunks one at a time. This currently ignores auxiliary files since this is usually reading large files for analysis and we don't want to load that many stamps into memory anyway.

* When serializing KBMOD results with image columns specified as separate auxiliary columns, we were mistakenly writing out image columns as parquets rather than .fits files. This was because when astropy serialized the image columns to parquet, multidimensional numpy arrays get flattened. These flattened arrays are then no longer considered image columns when loaded back into memory. I considered solving this through converting to byte arrays or pickling (or even  nested pandas over astropy tables), but instead we now store metadata of array shape for image columns in our serialized metadata. This is used to automatically reshape the image column arrays on load. There was an issue for roundtrip serialization in the CI environment that testing uncovered that was resolved by bumping our astropy version to >= 5.3

* We provide a commandline tool `kbmod-migrate-results` to glob over a directory's parquet Result files and splitting out any image columns into separate .fits files. This lets us take advantage of better compression to save on storage costs.

Example usage:
```
      # Finds parquet files with flattened coadd image columns and writes the images as auxilary columns
      kbmod-migrate-results --input=/path/to/results.parquet --image-columns '*coadd*' --stamp-dim 101
```

On an example test, this has let us change a 20 GB parquet file into 11 GB of a parquet file and associated rice_1 compressed fits files.

### DIFF
--- a/src/kbmod_cmdline/kbmod_migrate_results.py
+++ b/src/kbmod_cmdline/kbmod_migrate_results.py
@@ -101,6 +101,11 @@ def find_parquet_files(input_path, glob_pattern="**/*.search.parquet"):
         if ".search.parquet" in glob_pattern:
             parq_pattern = glob_pattern.replace(".search.parquet", ".search.parq")
             files.extend(input_path.glob(parq_pattern))
+        elif ".search.parq" in glob_pattern:
+            # On the other hand, if pattern uses .parq, also check for .parquet
+            # Note since .parq is a substring of .parquet, we check this second.
+            parquet_pattern = glob_pattern.replace(".search.parq", ".search.parquet")
+            files.extend(input_path.glob(parquet_pattern))
         return sorted(set(files))
 
     raise FileNotFoundError(f"Input path not found: {input_path}")
@@ -207,7 +212,7 @@ def count_parquet_rows_and_columns(filepath):
     return num_rows, len(colnames), colnames
 
 
-def validate_migration(original_path, new_path, expected_aux_files, expected_image_shapes=None):
+def validate_migration(original_path, new_path, expected_aux_files, expected_image_shapes):
     """Validate that migration was successful.
 
     Parameters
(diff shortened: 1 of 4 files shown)


src/kbmod_cmdline/kbmod_migrate_results.py
```diff
@@ -0,0 +1,914 @@
...
+        name_lower = input_path.name.lower()
+        if name_lower.endswith(".search.parquet") or name_lower.endswith(".search.parq"):
+            return [input_path]
+        else:
+            raise ValueError(f"Input file must be a .search.parquet file: {input_path}")
+
+    if input_path.is_dir():
+        files = list(input_path.glob(glob_pattern))
+        # Also check for .parq extension if pattern uses .parquet
+        if ".search.parquet" in glob_pattern:
```
jeremykubica: Would you ever want to go in the other direction ".search.parq" -> ".search.parquet"? Or does global already pick that up?

src/kbmod_cmdline/kbmod_migrate_results.py
```diff
@@ -0,0 +1,914 @@
...
+        # from legacy format, auxiliary files for image columns should not exist yet.
+        for chunk in Results.read_table_chunks(str(file_path), chunk_size=chunk_size):
+            # Reshape image columns in this chunk
+            reshape_image_columns_inplace(chunk, matched_columns, stamp_dim)
+            num_rows += len(chunk)
+
+            if accumulated_results is None:
+                accumulated_results = chunk
+            else:
+                accumulated_results.extend(chunk)
```
jeremykubica: Doesn't this end up reading the entire thing into memory since it is extending `accumulated_results` with each chunk?

jeremykubica (commented):
This looks really good. Only a few changes requested for some edge cases. I also put in some musings on how some things could be done more vectorized, but you don't need to make those changes.

PR #1128, opened 2026-01-30
Fix the occasional crashing bug

jeremykubica:
The tests were crashing in some cases due to invalid memory access during the CPU-only search. This only happened when the number of candidate trajectories was smaller than the results per pixel. The PR fixes it with improved bounds checking.

During debugging, I spent a while adding more safety checks throughout the core search. The PR includes those extra checks.


wilsonbb (approved):
Thanks for this!

PR #1148, opened 2026-06-11
Add option to deduplicate search candidates

jeremykubica:
To prevent searching many trajectories that end up on the same starting and ending pixels (e.g. all angle steps with v=0 or sufficiently small v), we pre-filter the candidates. This should improve the runtime of the core search and reduce the memory overhead.


wilsonbb (approved):
LGTM!

PR #1151, opened 2026-08-20
Fix wrong PSF assigned to every image in parallel reprojection

wilsonbb:
In `_reproject_work_unit_in_parallel`, the PSF lookup read `obstime` -- the leaked final value of the earlier for loop -- rather than the local variable of `time`, the per-result value already unpacked on the preceding line and used for `append_image`.

Because the carried over `obstime` is constant across the results loop, every reprojected image received the PSF associated with final obstime within the `WorkUnit`

Since the reprojection test used `shifted_wcs_diff_dimms_tiled.fits`, stores four byte-identical 3x3 PSFs, so selecting the wrong one returns an identical array. This error did not appear in production since we were using a uniform default PSF kernel rather than one derived from Butler, and once we fix that, this error would have surfaced. 

The new regression test therefore supplies its own deliberately distinct per-time kernels, differing in peak position and asymmetry rather than only in scale. Images sharing an obstime are given the same kernel so the test isolates the wrong-time defect from the separate question of which of several same-time constituent PSFs should be chosen; that remains unaddressed here.

Verified:
- New test failed on main (image 0 at t=60414 received the t=60416 kernel) and passes after the fix, on both serial and parallel paths. The serial path was already correct.
- Reprojected science, variance, mask, times, and per-image indices are bit-identical before and after the change on the existing test, whose identical PSFs make the corrected selection a no-op there.

### DIFF
--- a/src/kbmod/reprojection.py
+++ b/src/kbmod/reprojection.py
@@ -438,7 +438,10 @@ def _reproject_work_unit_in_parallel(
         stack = ImageStackPy()
         for result in future_reprojections:
             science_add, variance_add, mask_add, time = result.result()
-            psf = _get_first_psf_at_time(work_unit, obstime)
+            # Use the time carried by this result. Reading the submission loop's
+            # `obstime` here leaked its final value into every iteration, giving
+            # every image the last obstime's PSF.
+            psf = _get_first_psf_at_time(work_unit, time)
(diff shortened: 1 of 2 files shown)


maxwest-uw (approved):
nice catch!

PR #1160, opened 2026-09-10
Support injection across a "zeroed" background

wilsonbb:
Adds `zero_background=False` to retain only injected sources when enabled, including empty and failed-injection exposures. Injection runs on the original image before background subtraction to preserve gain inference.
- `zero_background`: defaults to `False`.
- `variance_scale`: defaults to `1.0`; other positive, finite values scale every returned variance plane. `1.0` *skips scaling entirely*.
Includes configuration logs and regression tests covering empty exposures, uniform scaling, and untouched variance pixels.
- `constant_variance=False`: When enabled, sets every returned variance plane to `1.0`. Rejects any `variance_scale` other than `1.0`.

We support piping this through parsl configurations in https://github.com/dirac-institute/kbmod-wf/pull/82


src/kbmod/injection.py
```diff
@@ -268,13 +287,18 @@ def inject_sources_into_ic(ic, catalog, butler, inject_config=None):
...
+                imdiff.image.array[:] = 0.0
             exposures.append(imdiff)
             injected_cats.append(
                 Table(names=catalog.colnames, dtype=[catalog[c].dtype for c in catalog.colnames])
             )
             references.append(ref)
             continue
 
+        # Preserve the real image for gain inference, even when removing its background.
+        original_image = np.array(imdiff.image.array, copy=True) if zero_background else None
```
jeremykubica, line 300: Should this line go before line 290? It looks like you are zeroing out the image and then saving the zeroed out image. Is that what you meant to do?

src/kbmod/injection.py
```diff
@@ -292,12 +318,26 @@ def inject_sources_into_ic(ic, catalog, butler, inject_config=None):
...
+            if zero_background:
+                imdiff.image.array[:] = 0.0
             exposures.append(imdiff)
             injected_cats.append(
                 Table(names=catalog.colnames, dtype=[catalog[c].dtype for c in catalog.colnames])
             )
         references.append(ref)
 
+    # Apply the same scaling to successful, empty-catalog, and no-render exposures.
+    if variance_scale != 1.0:
```
jeremykubica, line 330: When you zero the image, would you want to set the variable to some constant across the image?

src/kbmod/injection.py
jeremykubica: Although it looks like you test this case, so maybe it isn't a problem.

jeremykubica (commented):
A few questions

src/kbmod/injection.py
wilsonbb: Yeah the injection control flow is confusing here. There are two cases:

1. If `len(srccat) == 0` there was nothing to inject and line 290 will just zero out the image. We `continue` and don't proceed with the injection.
2. if `len(srccat) != 0` line 290 was not executed, and we save the original image which is later subtracted form the injected image. This is meant to avoid self-subtracting our injected PSF.

jeremykubica: Good point. I missed the continue.

src/kbmod/injection.py
wilsonbb: When scaling to a very small number in theory we preserve some depth information while keeping the injected signal a beacon for KBMOD to find, but I agree that simply setting to a constant probably is simpler and added as well for testing.

Now there's a `constant_variance` flag to just set the variance plane to `1.0` that is mutually exclusive for the scaling option.

jeremykubica (approved):

PR #1164, opened 2026-09-17
Store data on why the observations are marked invalid

### LINKED ISSUE #1163: obs_valid conflates off-patch geometry, masking and sigma-G clipping
### Problem

`obs_valid` is a single boolean per (row, epoch), but a `False` cell means one of three unrelated things:

1. the trajectory was never on the patch at that epoch,
2. it was on the patch but the pixel carried no data (masked, uncovered, or a hole in the stack),
3. it had data and sigma-G clipped it.

Anything that consumes `obs_valid` as a measure of "how much of this track did we actually recover" therefore mixes a geometric fact with an instrumental one with a statistical one. The third is the only one the name suggests, and it turns out to be the smallest by a wide margin.

### Measurement

Across ten 70 au reprojected LSSTCam patches (20'x20', 6000x6000, 70-100 day baselines), splitting the no-data cells by `pred_x` / `pred_y` against the image dimensions, which the results already carry:

| bucket | share of all cells |
|---|---|
| off-patch, the trajectory left the grid | 16.1% - 76.1% |
| masked, on the grid and still no data | 11.2% - 31.4% |
| kept | 9.5% - 50.9% |
| sigma-G clipped | 0.3% - 1.7% |

Sigma-G is under 2% of the `False` cells. Stated per epoch the trajectory was actually on the patch for, 36.6% to 69.5% still carry no data.

The method and its validation against kbmod's own output are in the comment on #1158.

### Why it bites

Metrics built on `obs_valid` divide by the whole stack regardless of whether the trajectory was ever on the patch, which makes them a function of patch geometry rather than of recovery. In our worst patch, 67.1% off-patch plus 22.9% masked left about 9.5% of epochs usable. That sits below the 0.12 floor of one of our downstream ramps, so the component read zero for every row in the patch and the score collapsed patch-wide. Nothing was wrong with the detections; the denominator counted epochs where the object could not have been imaged.

A 20'x20' patch over a multi-month baseline simply cannot hold every trajectory, so the off-patch share is expected and is not a defect. It just has to come out of any denominator that claims to measure recovery.

### Suggestion

Either of these would be enough on its own:

- Expose the reason alongside the boolean, so a caller can tell the three cases apart. The geometric split needs only `pred_x` / `pred_y` and the image dimensions, all of which the result already has; separating masked pixels from uncovered sky inside the remainder does need the WorkUnit mask planes.
- Where kbmod itself computes an epoch-count style quantity, document which denominator it uses, so downstream code does not have to guess whether off-patch epochs are included.

Happy to open a PR for the geometric split if that is a direction the project wants.

Related: #1158 (exposing a supported way to recompute `obs_valid`), #1162 (`obs_valid` and clustering).

jeremykubica:
Closes #1163 

Add another result column that tracks the reason why an observation is marked invalid and update the entire pipeline to populate this column. Observations can be marked invalid because they are out of bounds, are on a masked pixel, or fail sigma-g filtering.


wilsonbb (approved):
Thanks for this this!

PR #1165, opened 2026-09-17
Fix deprecation warning

jeremykubica:
Fix the `AstropyDeprecationWarning` "The isiterable function is deprecated and may be removed in a future version. Use numpy.iterable() instead."


wilsonbb (approved):
Thanks!

PR #1166, opened 2026-09-17
A series of optimization and checks

jeremykubica:
I asked copilot to look for possible errors or inefficiencies and it found a bunch. This fixes:
- unneeded "omp critical" sections that were making computation serial.
- missing "throw" statements for errors
- missing bounds checking
- formatting fixes
- making some things references


src/kbmod/search/psi_phi_array.cpp
```diff
@@ -177,6 +178,9 @@ PsiPhi PsiPhiArray::read_psi_phi(uint64_t time, int row, int col) const {
         (col >= meta_data.width)) {
         return result;
     }
+    if ((time < 0) || (time >= meta_data.num_times)) {
+        return result;
+    }
```
wilsonbb, line 182: Nice check to add!

src/kbmod/search/psi_phi_array.cpp
```diff
@@ -105,8 +104,10 @@ void PsiPhiArray::move_to_gpu() {
         logger->debug("Allocating times on GPU: " + gpu_time_array.stats_string());
         gpu_time_array.copy_vector_to_gpu(cpu_time_array);
         logger->debug(stat_gpu_memory_mb());
+
+        data_on_gpu = true;
```
wilsonbb: Trying to make sure I understand this part of the change

I guess previously an exception could occur, and `data_on_gpu` would be true when no GPU data was allocated

However we do copy twice to the GPU in this section. So it seems we could end up in the situation where
1. First copy succeeds
2. Second copy fails/triggers exception
3. `data_on_gpu` was never set true
4. We may not properly clean up allocated GPU memory (our old foe)

Does that seem right?

wilsonbb (commented):
Lots of good changes, just a question on one point

src/kbmod/search/psi_phi_array.cpp
jeremykubica: I think you are right here. Copilot pointed out that we might be setting `data_on_gpu` to True, but then failing to put the data onto the GPU. However, given how we handle the cleanup, we should set this to True as soon as any data might need to be cleaned up.

I reverted this part of the change and added a comment.

wilsonbb (approved):

PR #1168, opened 2026-09-18
Write results parquet to disk first

### LINKED ISSUE #1138: Write results parquet to disk first before working on extra columns
So that when the job is preempted, we don't start over.

jeremykubica:
Closes #1138

Write the main results file before the auxiliary files (primarily the image files) in case the job is pre-empted.

This contains several breaking API changes (which do not impact current KBMOD usage):
- `is_image_like()` method is removed. This was unused in practice since it was only called from `write_column()` and that always had the values explicitly set by the result of `_detect_image_columns()` instead.
- Rename the `colname` parameter to `column` in the `write_column()` method to reflect its changes role.


src/kbmod/results.py
```diff
@@ -818,47 +818,6 @@ def is_empty_value(self, colname):
...
-        if max_rows < 1:
-            return False
-
-        for idx in range(max_rows):
-            entry = self.table[colname][idx]
-            if isinstance(entry, np.ndarray) and entry.ndim >= 2:
-                return True
-        return False
-
     def filter_rows(self, rows, label=""):
```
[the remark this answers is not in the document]
jeremykubica: This is an intentional API change and the PR comment has been updated to reflect this.

src/kbmod/results.py
```diff
@@ -1127,89 +1086,90 @@ def write_table(
         # Write out the table.
         self.table.write(filename, overwrite=overwrite, **kwargs)
 
-    def write_column(self, colname, filename, overwrite=True, is_image=None):
+    def write_column(self, column, filename, overwrite=True, is_image=False):
```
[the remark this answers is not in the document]
jeremykubica: This is an intentional API change and the PR comment has been updated to reflect this.

src/kbmod/results.py
```diff
@@ -1368,52 +1333,47 @@ def write_results_to_files_destructive(
...
-    # Write the remaining data from the results to the main file.
+    # Write the remaining data from the results to the main file. We do this BEFORE writing
+    # out the auxiliary columns to guard against job pre-emption, ensuring the main table is
+    # completed first.
     logger.info(f"Saving results table to {filepath}")
     results.write_table(filepath, overwrite=overwrite, extra_meta=extra_meta)
+
+    # Write out the auxiliary columns to their own files after the main table has been saved.
+    # Note these columns have already been removed from the main table.
+    for col_name, col_data in all_separate_cols.items():
```
wilsonbb: A small comment is that this dictionary can be quite large in memory, so rather than iterating over the dictionary’s items, we can iterate over a list of the column names and pop off columns to deallocate them as they are written

wilsonbb (approved):
This looks good, thank you! A few line change suggested for memory purposes

src/kbmod/results.py
jeremykubica: I'm not sure the dictionary will be that large. It should be mapping strings (column name) to the column objects. Those objects should already be allocated in memory. We are just moving the reference from the table to the dictionary. Or am I wrong about that?

In terms of the alternate approach, I think we need to remove the columns from the table before this point. The write table call a few lines above needs the main table without the extra columns. So we need to have popped them before that. Or am I misinterpreting your suggestion.

wilsonbb: Yeah that is basically correct, this would be a non-major optimization since peak memory usage would be similar and we mostly care about that for OOMing. On line 1351 we do remove the auxiliary columns form the main table, so the dictionary should now be the main reference to pre-existing image columns, and my thought was that by `pop`ing from the `all_separate_cols` we could remove the final references to an individual auxiliary column after they are written. 
 
The main case that would be for me is if say we have a very large table (Colin has done experiments I believe with `num_results` set to 500,000) and several large image columns enabled, say 100x100 nightly coadds, would be 5 billion pixels per column which adds up. But in practice we are saving the workunits and then re-extracting coadds as needed to avoid the storage and write time. So it seemed nice to try free up memory a little earlier, but I don't think would be a required optimization

jeremykubica: Thanks for the clarification. I will leave it as-is for now and merge it in. But we can revisit this later if needed.

